Issue #1, opened 2023-01-13
Hubs v0.1 – Power BI on CM exports


flanakin:
## 🚀 Goal
Create an ARM template that configures a new Cost Management export to ADLS and generates a Power BI report that can work against that dataset. The template should also include a Data Factory pipeline to delete old files when data is exported each day.

## 📋 Tasks
```[tasklist]
### v0.0.1 - finops-hub template ([discuss](https://github.com/microsoft/cloud-hubs/discussions/44))
- [x] #5
- [x] Generate aka.ms links for all report pages
- [x] #21
- [x] Split out 0.1 into smaller releases
- [ ] https://github.com/microsoft/cloud-hubs/issues/24
- [ ] https://github.com/microsoft/cloud-hubs/issues/29
- [ ] https://github.com/microsoft/cloud-hubs/issues/37
- [ ] https://github.com/microsoft/cloud-hubs/issues/51
- [ ] https://github.com/microsoft/cloud-hubs/issues/23
- [ ] https://github.com/microsoft/cloud-hubs/issues/58
- [ ] https://github.com/microsoft/cloud-hubs/issues/65
- [ ] https://github.com/microsoft/cloud-hubs/issues/34
- [ ] https://github.com/microsoft/cloud-hubs/issues/59
- [ ] https://github.com/microsoft/cloud-hubs/issues/64
- [ ] https://github.com/microsoft/cloud-hubs/issues/124
- [ ] https://github.com/microsoft/cloud-hubs/issues/126
- [ ] https://github.com/microsoft/cloud-hubs/issues/125
- [ ] https://github.com/microsoft/cloud-hubs/issues/63
- [x] Test deployment
- [ ] https://github.com/microsoft/cloud-hubs/issues/144
- [ ] #26
- [x] Sync feature branches
- [x] Create new release
- [x] Release maintenance
- [x] Address all feedback from the release discussion
```

```[tasklist]
### v0.0.2 - Self-managed exports ([discuss](https://github.com/microsoft/cloud-hubs/discussions/45))
- [ ] https://github.com/microsoft/cloud-hubs/issues/168
- [ ] https://github.com/microsoft/finops-toolkit/issues/187
- [ ] https://github.com/microsoft/finops-toolkit/issues/186
- [x] Ingest data across hub instances
- [x] Add actual cost data
- [x] Historical data refresh (rerating)
- [ ] https://github.com/microsoft/cloud-hubs/issues/60
[...]

PR #5, opened 2023-01-28
Power BI on raw CM exports spec

flanakin:
## 🛠️ Description
Spec for Power BI on raw CM exports (#1).

Includes:
- How to deploy doc
- finops-hub and finops-hub-with-exports template specs
- hub and export module specs


BenShy (approved):

docs/deploy/README.md
```diff
@@ -0,0 +1,87 @@
...
+- [Deploy from Azure Quickstart Templates](#deploy-from-azure-quickstart-templates)
+- [Future considerations](#future-considerations)
+
+---
+
+## Using custom deployment in the Azure portal
+
+The Azure portal includes a **Custom deployment** option that supports all templates available in the Azure Quickstart Templates repository. To deploy a quickstart template:
+
+1. Open [Custom deployment](https://portal.azure.com/#create/Microsoft.Template)
```
tsilvers-ms, line 27: Not sure if it's important but the option in portal is "Deploy a custom template" than on the actual page as you point out it's "Custom deployment" might confuse some users the difference.

docs/reports/commitment-discounts.md
```diff
@@ -0,0 +1,158 @@
...
+
+The **Commitment discounts report** summarizes existing and potential savings from commitment-based discounts, like reservations and savings plans. This report enables you to:
+
+- Review Azure Hybrid Benefit usage.
+- Identify and resolve any under-utilized commitments (aka utilization).
+- Identify opportunity to save with more commitment-based discounts (aka coverage).
+- Determine which resources used commitment-based discounts (aka chargeback).
+- Summarize cost savings from commitment-based discounts.
+
+> 🚩 **Important**<br>FinOps hubs uses amortized costs. Amortization breaks reservation and savings plan purchases down and allocates costs to the resources that received the benefit. Due to this, amortized costs will not show purchase costs and will not match your invoice. Please use Cost Management to review invoice charges.
```
tsilvers-ms, line 14: Is there a link we can include to explain more about amortized cost, I'm concerned very new users will still not understand. Also can we link "Cost Management" to the actual portal landing page?

docs/reports/commitment-discounts.md
```diff
@@ -0,0 +1,158 @@
...
+
+The chart breaks down the cost of used (utilized) vs. unused charges. Unused charges are split out by commitment type (e.g., reservation, savings plan).
+
+The table shows resource usage against commitment-based discounts with columns for resource name, resource group, subscription, and commitment. Use the table for chargeback and savings calculations.
+
+This page filters usage down to only show charges related to commitment-based discounts, which means the total cost on the Commitments page won't match other pages, which aren't filtered by default.
+
+### 🛠️ Customization tips
+
+- Consider changing the columns in the table based on your chargeback needs.
```
tsilvers-ms, line 111: What are the thoughts around rework with any customizations. is it worth mentioning here?

docs/reports/cost-summary.md
```diff
@@ -0,0 +1,161 @@
...
+
+## See also
+
+- [Common terms](./terms.md)
+
+---
+
+## Future considerations
+
+- #11
```
tsilvers-ms, line 157: Should we add support for MG here?

docs/templates/README.md
```diff
@@ -0,0 +1,72 @@
...
+- [finops-hub](./finops-hub.md)
+- [finops-hub-with-exports](./finops-hub-with-exports.md)
+
+---
+
+## Creating templates
+
+Templates follow the [Azure Quickstart Templates guidelines](https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/README.md#contribution-guide):
+
+1. Every template should be in its own folder.
```
tsilvers-ms, line 18: Aren't we breaking this by having the two templates in the root templates folder?

docs/templates/finops-hub-with-exports.md
```diff
@@ -0,0 +1,46 @@
...
+
+- [Parameters](#parameters)
+- [Modules](#modules)
+- [Outputs](#outputs)
+
+---
+
+## Parameters
+
+- **subscription** (string) – ID of the subscription to deploy the hub instance to.
```
tsilvers-ms, line 24: I'm confused the description makes it sound like this template is the full export->adls->powerBI yet it lacks the parameters needed to create storage, tags, etc. like the other script.

docs/templates/modules/README.md
```diff
@@ -0,0 +1,16 @@
+# FinOps toolkit modules
+
+FinOps toolkit templates are comprised of [Bicep](https://learn.microsoft.com/azure/azure-resource-manager/bicep) modules. Bicep is a domain-specific language that uses declarative syntax to define and deploy Azure resources. For a guided learning experience, start with the [Fundamentals of Bicep](https://learn.microsoft.com/training/paths/fundamentals-bicep/).
```
tsilvers-ms, line 3: Since the custom deployment seems to be limited to json should we talk about Bicep?

docs/templates/modules/export.md
```diff
@@ -0,0 +1,46 @@
...
+- [Parameters](#parameters)
+- [Resources](#resources)
+- [Outputs](#outputs)
+
+---
+
+## Parameters
+
+- **exportName** (string) – Name of the export instance.
+- **amortize** (boolean) – Optional. Indicates whether the export should amortize reservation purchases. Default: `false`.
```
tsilvers-ms, line 21: The warning about amortized implies this would not be optional or false for RI. Please clarify.

docs/templates/modules/export.md
```diff
@@ -0,0 +1,46 @@
...
+- **frequency** (string) – Optional. Indicates how often data should be exported. Allowed: Daily, Weekly, Monthly, Annually. Default: `Daily`.
+- **startDate** (string) – Optional. Day to run the first export. Default: Current day.
+- **endDate** (string) – Optional. Day to run the last export. Default: 1 year from the current day.
+- **period** (string) – Optional. Time frame to export data for. Allowed: BillingMonthToDate, MonthToDate, TheLastBillingMonth, TheLastMonth, WeekToDate. Default: `MonthToDate`.
+- **storageAccountId** (string) – Resource ID for the storage account to export files to.
+- **storageAccountContainer** (string) – Name of the container to export files to.
+- **storageAccountPath** (string) – Path within the container to export files to.
+- **runNow** (bool) – Optional. Indicates whether to run the export immediately. Default: true.
```
tsilvers-ms, line 29: What about future looking, partition, compression, format etc.?

docs/templates/modules/export.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Required settings for FinOps hubs
+
+Note the following parameter values must be set for FinOps hubs to work as designed. Changing these will require additional changes to ensure reporting is accurate.
+
+- **amortize** must be `true`
+- **storageAccountContainer** must be `ms-cm-exports`
```
tsilvers-ms, line 36: Since these are requirements for FinOps should we exclude them from the template and hard code the values instead?

docs/templates/modules/hub.md
```diff
@@ -0,0 +1,60 @@
...
+}
+```
+
+## Outputs
+
+- **name** (string) – Name of the deployed hub instance.
+- **location** (string) – Azure resource location resources were deployed to.
+- **storageAccountId** (string) – Resource ID of the storage account created for the hub instance. This must be used when creating the Cost Management export.
+- **primaryBlobEndpoint** (string) – Primary blob endpoint reference for the storage account.
+- **pbix** (string) – URL for the Power BI file to use for this template.
```
tsilvers-ms, line 60: What about the actual pbix files, where do they come into play with the rest of the docs?

docs/deploy/README.md
flanakin: I can share the feedback with the portal team. I'm just documenting what they have right now. I plan to add a separate proposal for what we should do in CM where we can do more.

docs/reports/commitment-discounts.md
flanakin: That's a good call. Let me add something to the main reports readme.

docs/reports/cost-summary.md
flanakin: Let's see what customers and the community say. I'd rather avoid spending time there if we can, but if needed, I do think we have some cheap workarounds.

docs/templates/README.md
flanakin: This is just the docs. I need to move this over to dev docs under src since these notes are more internal than external. I'll do that in a future PR.

docs/templates/modules/README.md
flanakin: Bicep generates JSON. We should be good.

docs/templates/modules/export.md
flanakin: My plan was to use this module as a general-purpose way to automate export creation. It's not exclusively for hubs. `hub.bicep` forces amortization to be enabled for hubs.

docs/templates/modules/export.md
flanakin: We can add that when we plan to do the work. Trying to keep the spec hyper-focused on the first release.

docs/templates/modules/export.md
flanakin: Same as earlier: This module is intended to be a general-purpose way to create exports. If we don't want that, then I'd agree we should remove it. I see value in both approaches, since we don't currently have a published template for exports. I was thinking about using this for both 😄

docs/templates/modules/hub.md
flanakin: The reports folder covers what the PBIX files should include. We'll commit those into the src folder for download when we get there.

docs/deploy/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+<br>
+
+## Deploy from Microsoft Learn code samples
+
+Microsoft Learn hosts all templates available from the Azure Quickstart Templates repository. To deploy from Microsoft Learn:
+
+1. Open the [FinOps hub code sample](https://learn.microsoft.com/en-us/samples/azure/azure-quickstart-templates/finops-hub-with-exports).
+2. Select the **Deploy to Azure** button towards the top of the page.
+3. Complete [steps 5 and beyond](#edit-params) from the custom deployment steps above.
```
flanakin, line 59:
```suggestion
3. Complete [steps 4 and beyond](#edit-params) from the custom deployment steps above.
```

docs/deploy/README.md
```diff
@@ -0,0 +1,87 @@
...
+Templates are organized based on scope:
+
+- Tenant templates in [tenant-deployments](https://github.com/Azure/azure-quickstart-templates/tree/master/tenant-deployments)
+- Resource group templates in [quickstarts/microsoft.costmanagement](https://github.com/Azure/azure-quickstart-templates/tree/master/quickstarts/microsoft.costmanagement)
+
+We do not currently have any subscription or management group templates.
+
+To deploy a quickstart template:
+
+1. Open the [FinOps hub template folder](https://github.com/Azure/azure-quickstart-templates/tree/master/quickstarts/microsoft.costmanagement/finops-hub-with-exports).
```
flanakin, line 76:
```suggestion
1. Open the [FinOps hub with exports template](https://github.com/Azure/azure-quickstart-templates/tree/master/template-deployments/finops-hub-with-exports).
```

docs/deploy/README.md
```diff
@@ -0,0 +1,87 @@
...
+2. Select the **Deploy to Azure** button towards the top of the page.
+3. Complete [steps 5 and beyond](#edit-params) from the custom deployment steps above.
```
flanakin, line 78:
```suggestion
3. Complete [steps 4 and beyond](#edit-params) from the custom deployment steps above.
```

docs/reports/commitment-discounts.md
```diff
@@ -0,0 +1,158 @@
...
+> 🚩 **Important**<br>FinOps hubs uses [amortized costs](https://learn.microsoft.com/azure/cost-management-billing/reservations/reservation-amortization). Amortization breaks reservation and savings plan purchases down and allocates costs to the resources that received the benefit. Due to this, amortized costs will not show purchase costs and will not match your invoice. Please use [Cost Management](https://aka.ms/costmgmt) to review invoice charges.
+
+On this page:
+
+- [Common page layout](#common-page-layout)
+- [Get started](#get-started)
+- [Hybrid Benefit](#hybrid-benefit)
+- [Commitments](#commitments)
+- [Coverage](#coverage)
+- [Pricesheet](#pricesheet)
```
flanakin, line 23: Remove this since it won't be able to align to the FOCUS spec

docs/templates/finops-hub-with-exports.md
```diff
@@ -0,0 +1,48 @@
+---
```
aromano2, line 1: In the readme for templates, it says all templates should be placed within their own folder. Not sure if we care necessarily but should update one or the other.

aromano2: Never mind saw someone already commented on that

aromano2 (approved):

PR #47, opened 2023-02-07
Added datafactory to main template

aromano2:
## 🛠️ Description

Fixes #(issue)

- [ ] 💪 Unit tests

### DIFF
--- a/src/modules/Microsoft.Network/privateEndpoints/.bicep/nested_roleAssignments.bicep
+++ b/src/modules/Microsoft.Network/privateEndpoints/.bicep/nested_roleAssignments.bicep
@@ -1,0 +1,93 @@
+@sys.description('Required. The IDs of the principals to assign the role to.')
+param principalIds array
+
+@sys.description('Required. The name of the role to assign. If it cannot be found you can specify the role definition ID instead.')
+param roleDefinitionIdOrName string
+
+@sys.description('Required. The resource ID of the resource to apply the role assignment to.')
+param resourceId string
+
+@sys.description('Optional. The principal type of the assigned principal ID.')
+@allowed([
+  'ServicePrincipal'
+  'Group'
+  'User'
+  'ForeignGroup'
+  'Device'
+  ''
+])
+param principalType string = ''
+
+@sys.description('Optional. The description of the role assignment.')
+param description string = ''
+
+@sys.description('Optional. The conditions on the role assignment. This limits the resources it can be assigned to. e.g.: @Resource[Microsoft.Storage/storageAccounts/blobServices/containers:ContainerName] StringEqualsIgnoreCase "foo_storage_container".')
+param condition string = ''
+
+@sys.description('Optional. Version of the condition.')
+@allowed([
+  '2.0'
+])
+param conditionVersion string = '2.0'
+
+@sys.description('Optional. Id of the delegated managed identity resource.')
+param delegatedManagedIdentityResourceId string = ''
+
+var builtInRoleNames = {
+  'Avere Contributor': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '4f8fab4f-1852-4a58-a46a-8eaf358af14a')
+  'Avere Operator': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'c025889f-8102-4ebf-b32c-fc0c6f0c6bd9')
+  'Azure Kubernetes Service Policy Add-on Deployment': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '18ed5180-3e48-46fd-8541-4ea054d57064')
+  'Backup Contributor': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '5e467623-bb1f-42f4-a55d-6e525e11384b')
+  'Backup Operator': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '00c29273-979b-4161-815c-10b084fb9324')
+  Contributor: subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'b24988ac-6180-42a0-ab88-20f7382dd24c')
+  'Cosmos DB Operator': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '230815da-be43-4aae-9cb4-875f7bd000aa')
+  'Desktop Virtualization Virtual Machine Contributor': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'a959dbd1-f747-45e3-8ba6-dd80f235f97c')
+  'DevTest Labs User': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '76283e04-6283-4c54-8f91-bcf1374a3c64')
+  'DNS Resolver Contributor': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '0f2ebee7-ffd4-4fc0-b3b7-664099fdad5d')
+  'DNS Zone Contributor': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'befefa01-2a29-4197-83a8-272ff33ce314')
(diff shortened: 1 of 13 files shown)


src/modules/Microsoft.Network/privateEndpoints/.bicep/nested_roleAssignments.bicep
```diff
@@ -0,0 +1,93 @@
+@sys.description('Required. The IDs of the principals to assign the role to.')
```
flanakin, line 1: Can you remove the .bicep and .test folders? I don't think we need those.

src/modules/hub.bicep
```diff
@@ -7,7 +7,9 @@ param hubName string
 
 // Generate unique storage account name
 var storageAccountSuffix = 'store'
+var dataFactorySuffix = 'factory'
 var storageAccountName = '${substring(replace(toLower(hubName), '-', ''), 0, 24 - length(storageAccountSuffix))}${storageAccountSuffix}'
+var dataFactoryName = '${substring(replace(toLower(hubName), '-', ''), 0, 24 - length(dataFactorySuffix))}${dataFactorySuffix}'
```
flanakin, line 12: Does ADF have the same lowercase, no special characters limits as storage? If not, can we keep the original casing and not remove characters?

src/modules/hub.bicep
```diff
@@ -7,7 +7,9 @@ param hubName string
 
 // Generate unique storage account name
 var storageAccountSuffix = 'store'
+var dataFactorySuffix = 'factory'
```
flanakin, line 10: The spec calls out that this resource will be `'${hubName}-engine'`. Did you change that on purpose? We should either use that or update the docs if we think "factory" makes more sense.

flanakin: @tsilvers-ms @ankur-ms Do either of you have a strong opinion on the naming convention for the ADF instance? I don't see anything online outside of people picking test names. I used "engine" in the spec to signify the resource's purpose relative to the app, since this is the core component that runs everything. I don't have a problem with "factory" outside of it being more about what it is rather than what it does, but maybe I'm being overly pedantic 😜

src/modules/hub.bicep
```diff
@@ -58,6 +60,15 @@ module storageAccount 'Microsoft.Storage/storageAccounts/deploy.bicep' = {
   }
 }
 
+module dataFactory 'Microsoft.DataFactory/factories/deploy.bicep' = {
+  name: 'factory'
```
flanakin, line 64: nit: `dataFactory` in case there is some other "factory" that would overlap
```suggestion
  name: 'dataFactory'
```

src/modules/hub.bicep
```diff
@@ -70,3 +81,6 @@ output location string = location
 
 @description('Resource ID of the storage account created for the hub instance. This must be used when creating the Cost Management export.')
 output storageAccountId string = storageAccount.outputs.resourceId
+
+@description('The Resource ID of the Data factory.')
+output dataFactoryId string = dataFactory.outputs.resourceId
```
flanakin, line 86: Does the user need the data factory ID for anything? If not, I wouldn't pass it thru to outputs. It's easier to add it than take it away. If we need to keep it, we should update the docs as well.

src/templates/finops-hub/main.bicep
```diff
@@ -46,3 +46,6 @@ output location string = location
 
 @description('The resource ID of the deployed storage account.')
 output storageAccountId string = hub.outputs.storageAccountId
+
+@description('The Resource ID of the Data factory.')
+output dataFactoryId string = hub.outputs.dataFactoryId
```
flanakin, line 51: Same as in hub.bicep. Either remove or add to docs.

src/modules/Microsoft.Network/privateEndpoints/deploy.bicep
```diff
@@ -0,0 +1,131 @@
...
+    mode: 'Incremental'
+    template: {
+      '$schema': 'https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#'
+      contentVersion: '1.0.0.0'
+      resources: []
+    }
+  }
+}
+
+resource privateEndpoint 'Microsoft.Network/privateEndpoints@2022-05-01' = {
```
flanakin, line 65: How are we using this in the ADF instance? Is it really required?

flanakin (changes requested):

flanakin:
> ## 🔬 How has this been tested?
> * [ ]  🫰 PS -WhatIf / az validate
> * [ ]  👍 Manually deployed + verified
> * [ ]  💪 Unit tests

Did you deploy it? If so, check the second box.

src/modules/Microsoft.Network/privateEndpoints/deploy.bicep
aromano2: This would be used to connect to the integration runtime. It's not required and can be either public or private. I'm not sure if this is something we want to remove yet.

src/modules/Microsoft.Network/privateEndpoints/.bicep/nested_roleAssignments.bicep
aromano2: We do need the .bicep folder, it contains resources referenced in the main module

src/modules/hub.bicep
aromano2: removed

src/modules/hub.bicep
aromano2: fixed

src/templates/finops-hub/main.bicep
aromano2: removed

src/modules/hub.bicep
aromano2: I updated the naming convention and added min/max length validation.

src/modules/Microsoft.Network/privateEndpoints/deploy.bicep
flanakin: It's not needed: https://learn.microsoft.com/en-us/azure/data-factory/data-factory-private-link. We can safely remove it. If for some reason we need to add it later, we can bring it back.

src/modules/Microsoft.Network/privateEndpoints/readme.md
```diff
@@ -0,0 +1,487 @@
+# Private Endpoints `[Microsoft.Network/privateEndpoints]`
```
flanakin, line 1: We don't need this. We can remove it.

flanakin (changes requested):

src/modules/Microsoft.Network/privateEndpoints/deploy.bicep
aromano2: removed

src/modules/Microsoft.Network/privateEndpoints/readme.md
aromano2: removed

src/modules/hub.bicep
aromano2: removed

src/modules/hub.bicep
```diff
@@ -5,6 +5,11 @@
 @description('Optional. Name of the hub. Used to ensure unique resource names. Default: "finops-hub".')
 param hubName string
 
+@minLength(3)
+@maxLength(63)
+@description('Optional. Name of the data factory. Used to ensure unique resource names. Default: "finops-hub-engine".')
+param dataFactoryName string
```
flanakin, line 11: Do we need to take this as an input? I'd rather have fewer inputs and standardize the name across implementations. If they change it, then there's more risk that they might misconfigure it when they deploy the next release. Ultimately, when this is a managed solution, this will be a bit of a black box that they won't be able to change. Do you have any concerns with that approach?

flanakin (approved):

Issue #86, opened 2023-03-20
Scheduled action bicep module for RBAC scopes


flanakin:
## 📝 Scenario
As a **FinOps practitioner**, I need to **deploy scheduled actions as part of my templates** in order to **automate onboarding for new projects**

## 💎 Solution
Publish scheduled action modules to the [Bicep Registry](https://github.com/Azure/bicep-registry-modules) for each supported scope.

## ℹ️ Additional context
Bicep requires separate files per deployment scope. Since this would require excessive duplication of code, we'll introduce a simple templating language optimized for enabling multi-scope deployment modules that support native Bicep tooling. 

## 🙋‍♀️ Ask for the community

We could use your help:
1. Please vote this issue up (👍) to prioritize it.
2. Leave comments to help us solidify the vision.

flanakin:
Bicep modules are available from the Bicep Registry @ https://aka.ms/finops/toolkit/bicepregistry.

PR #87, opened 2023-03-20
Add bicep registry docs and scripts

flanakin:
## 🛠️ Description
This PR adds documentation and scripts to support modules that will be published to the [Bicep Registry](https://github.com/Azure/bicep-registry-modules). Splitting the docs and scripts to simplify the PR.

Supports #86

- [ ] 💪 Unit tests

### DIFF
--- a/src/architecture.md
+++ b/src/architecture.md
@@ -1,15 +1,16 @@
 # Architecture
 
-| Name                                             | Description                     |
-| ------------------------------------------------ | ------------------------------- |
-| [docs](../docs)                                  | Public-facing toolkit docs.     |
-| &nbsp;↳ [deploy](../docs/deploy)                 | How to deploy the toolkit.      |
-| &nbsp;↳ [reports](../docs/reports)               | About Power BI reports.         |
-| &nbsp;↳ [templates](../docs/templates)           | About ARM deployment templates. |
-| &nbsp; &nbsp; &nbsp;↳ [modules](../docs/modules) | About Bicep modules.            |
-| [src](../src)                                    | Source code and internal docs.  |
-| &nbsp;↳ [templates](../src/templates)            | ARM deployment templates.       |
-| &nbsp;↳ [modules](../src/modules)                | Bicep modules.                  |
+| Name                                               | Description                      |
+| -------------------------------------------------- | -------------------------------- |
+| [docs](../docs)                                    | Public-facing toolkit docs.      |
+| ├─ [deploy](../docs/deploy)                        | How to deploy the toolkit.       |
+| ├─ [reports](../docs/reports)                      | About Power BI reports.          |
+| └─ [templates](../docs/templates)                  | About ARM deployment templates.  |
+| &nbsp; &nbsp; &nbsp; └─ [modules](../docs/modules) | About Bicep modules.             |
+| [src](../src)                                      | Source code and internal docs.   |
+| ├─ [bicep-registry](../src/bicep-registry)         | Bicep registry module templates. |
+| ├─ [modules](../src/modules)                       | Bicep modules.                   |
+| └─ [templates](../src/templates)                   | ARM deployment templates.        |
 
 Files and folders should use kebab casing (e.g., `this-is-my-folder`). The only exception is for RP namespaces in module paths.
 
--- a/src/bicep-registry/README.md
+++ b/src/bicep-registry/README.md
@@ -136,7 +136,7 @@ Our templating language supports "scope directives" that enable you to:
 1. Leverage existing Bicep developer tooling for the primary scope.
 2. Target multiple scopes.
 3. Conditionally include a single line.
-4. Conditionally include a multiple lines.
+4. Conditionally include multiple lines.
 
 ### Scope directives
 
--- a/src/bicep-registry/README.md
+++ b/src/bicep-registry/README.md
@@ -1,0 +1,229 @@
+# Bicep registry modules
+
+This folder contains modules that are published to the official [Bicep Registry](https://github.com/Azure/bicep-registry-modules). Modules are maintained here using a simple templating language to facilitate generating modules for multiple scopes without duplicating code.
+
+- [Scheduled action](./scheduled-action)
+
+<br>
+
+On this page:
+
+- [Creating a new module](#creating-a-new-module)
+- [Building bicep registry modules](#building-bicep-registry-modules)
+- [Testing bicep registry modules](#testing-bicep-registry-modules)
+- [Publishing bicep registry modules](#publishing-bicep-registry-modules)
+- [Templating language](#templating-language)
+
+---
+
+## Creating a new module
+
+1. Create a folder for the module using kebab casing (e.g., `my-resource`). Module names should be:
+   - Singular (e.g., `my-resource` instead of `my-resources`).
+   - Named after the resource type (e.g., `virtual-machine` for `virtualMachines`).
+2. Run `brm generate` to create placeholder files for the module. [Learn more](https://github.com/Azure/bicep-registry-modules/blob/main/CONTRIBUTING.md#generating-module-files).
+3. Update the following files:
+
+   - `metadata.json` – Set the name, summary, and owner (should be `cost-management-contrib`).
+   - `version.json` – Set version to `1.0`.
+   - `README.md` – Add a title and description only. Do not set parameters, outputs, or examples.
+
+   > _Note: Do not update main.json._
+
+4. Implement the module in `main.bicep` using the templating language below.
+5. Implement tests in `test/main.test.bicep` also using the templating language.
+
+<br>
+
+## Building bicep registry modules
+
+There are 2 ways to build bicep registry modules. To build all toolkit modules and templates, run:
+
+```console
+cd $repo/src/scripts
+./Build-Toolkit
+```
+
+To build only a single module, run:
+
+```console
+cd $repo/src/scripts
+./Build-Bicep ..\bicep-registry\<module>
+```
+
+The `Build-Bicep` script supports the following parameters:
+
+| Parameter | Description                                                               |
+| --------- | ------------------------------------------------------------------------- |
+| `-Module` | Required. Specifies the module to build.                                  |
+| `-Scope`  | Optional. Indicates one scope to generate. Default: all scopes.           |
+| `-Debug`  | Optional. Writes primary file output to console. Does not generate files. |
+
+> ℹ️ _Note: Both build scripts must be run from the `src/scripts` folder._
+
+<br>
+
+## Testing bicep registry modules
+
+Before deploying a module, you first need to sign in to Azure:
+
+```console
+Connect-AzContext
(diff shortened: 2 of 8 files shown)


src/architecture.md
```diff
@@ -1,15 +1,16 @@
...
-| &nbsp;↳ [modules](../src/modules)                | Bicep modules.                  |
+| Name                                               | Description                      |
+| -------------------------------------------------- | -------------------------------- |
+| [docs](../docs)                                    | Public-facing toolkit docs.      |
+| ├─ [deploy](../docs/deploy)                        | How to deploy the toolkit.       |
+| ├─ [reports](../docs/reports)                      | About Power BI reports.          |
+| └─ [templates](../docs/templates)                  | About ARM deployment templates.  |
+| &nbsp; &nbsp; &nbsp; └─ [modules](../docs/modules) | About Bicep modules.             |
+| [src](../src)                                      | Source code and internal docs.   |
+| ├─ [bicep-registry](../src/bicep-registry)         | Bicep registry module templates. |
```
flanakin, line 11: This line is the only real change.

src/templates/finops-hub/test/main.test.bicep
flanakin, line 1: Added this file to move default-able parameters out of the deployment script.

src/bicep-registry/README.md
```diff
@@ -0,0 +1,229 @@
...
+1. Modules cannot target multiple scopes.
+2. Scope functions behave differently for each targeted scope.
+3. No mechanism for conditional logic based on scope.
+
+Our templating language supports "scope directives" that enable you to:
+
+1. Leverage existing Bicep developer tooling for the primary scope.
+2. Target multiple scopes.
+3. Conditionally include a single line.
+4. Conditionally include a multiple lines.
```
tsilvers-ms, line 139: remove the 'a' before multiple lines.

src/bicep-registry/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Scope directives
+
+Scope directives are single-line comments that reference one or more Bicep-supported scopes. Scopes are prefixed with `@` and are separated by spaces. Do not include additional characters after the scope directive.
```
tsilvers-ms, line 143: Should we list or link to Bicep-supported scopes, I mean ln 151 is likely the list but not explicitly called out.

tsilvers-ms, line 143: Will the choice of '@' conflict with other uses? vs say using '~' as the prefix.

src/bicep-registry/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+The following example references only the resource group scope:
+
+```bicep
+// @resourceGroup
```
tsilvers-ms, line 148: I'm assuming the comment // are required? It's a little confusing when seeing the references in a larger context with other @ directives without comments.

src/bicep-registry/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+Conditional lines are used to include (or exclude) a single line within a Bicep module. To use a conditional line, add a scope directive to the end of the desired line.
+
+The following example includes 2 conditional lines to represent allowed values that are only supported for specific scopes. `value1` is allowed for resource groups and subscriptions, and `value3` is allowed for tenant deployments:
+
+```bicep
+@allowed([
+  'value1' // @resourceGroup @subscription
+  'value2'
+  //'value3' // @tenant
```
tsilvers-ms, line 182: yeah, the // before the scope when the line is commented is confusing, is it possible to use a different character sequence? //'value3' || @tenant

src/bicep-registry/README.md
```diff
@@ -0,0 +1,229 @@
...
+Conditional blocks are used to include (or exclude) multiple lines within a Bicep module. To use a conditional block, add a scope directive immediately before the code block. Conditional blocks end when either another conditional block is specified or an empty line.
+
+The following example includes a single conditional block that is only includes `property1` and `property2` for subscriptions:
+
+```bicep
+resource myResource 'Microsoft.MyProvider/resource' = {
+  name: 'myResource'
+  properties: {
+    // @subscription
+    //// Comments must have 4 slashes
```
tsilvers-ms, line 199: Ah, but on ln186 the text says that ln182 is commented out which conflicts with this line saying Comments need 4 slashes. Consider defining comments as 4 slashes earlier and more explicitly.

src/scripts/Build-Bicep.ps1
```diff
@@ -0,0 +1,189 @@
...
+        # Init string builder for the scope output
+        $sb = [System.Text.StringBuilder]::new()
+  
+        # Write debug header
+        if ($Debug) {
+            Write-Host $moduleName
+            Write-Host "".PadLeft($moduleName.Length, "=")
+        }
+  
+        # Start on second line to get past the supported scope directive
```
tsilvers-ms, line 66: This seems a brittle assumption, we should eventually be more flexible so if there is a /n on first line and second line is the scope it still works.

src/bicep-registry/README.md
tsilvers-ms: What about an identifier scheme to make it easier to parse out? 
Ex. @scope:resourceGroup @scope:subscription or @scope(resourceGroup) @scope(subscription)
That way when identifying scopes vs other directives is clear?

src/scripts/Build-Bicep.ps1
```diff
@@ -0,0 +1,189 @@
...
+        $i = 1
+        $script:lastLineEmpty = $true
+        while ($i -lt ($lines.Count)) {
```
tsilvers-ms, line 69: nit: replace with a foreach on lines, unless I missed some use to $i?

src/scripts/Build-Bicep.ps1
```diff
@@ -0,0 +1,189 @@
...
+                [void]$sb.AppendLine($line)
+            }
+            # If end of module, close code block
+            elseif ($writingModule -and $line -eq '}') {
+                [void]$sb.AppendLine($line).AppendLine('```')
+                $writingModule = $false
+            }
+        }
+    
+        # Append examples to README file
```
tsilvers-ms, line 186: Will this make the readme grow and grow if run frequently?

src/scripts/Deploy-Toolkit.ps1
```diff
@@ -1,53 +1,150 @@
...
-$params = @{ hubName = $ResourceGroup };
-New-AzResourceGroupDeployment `
-    -TemplateFile "../templates/$Template/main.bicep" `
-    -TemplateParameterObject $params `
-    -ResourceGroupName $ResourceGroup `
-    -WhatIf:$WhatIf
+# Find bicep file (templates first)
+@("../templates", "../bicep-registry", "../../release") `
+| ForEach-Object { Get-Item (Join-Path $_ $Template (iff $Test test/main.test.bicep main.bicep)) -ErrorAction SilentlyContinue } `
+| ForEach-Object {
```
tsilvers-ms, line 68: Somewhere in this file is where we should be recording deployments via a logging API of some sort, can we add a placeholder?

src/bicep-registry/README.md
flanakin: Are you aware of any other uses of @ in comments? We could use # to align with pragma directives. I'm not picky. I was thinking "at subscription scope". I'm assuming it shouldn't matter unless the @ reference is paired with a scope.

src/bicep-registry/README.md
flanakin: Yes, the comments are required to make this a valid Bicep file. You have a good point about `@resourceGroup` in comparison with `@description`, `@allowed`, etc. If that's your main concern, I'd probably suggest switching to `#resourceGroup`. The regex would get more complicated if we add a prefix/function. I prefer `@resourceGroup`, but happy to change the character if we think that's important. I'd rather not add the prefix/function, but I only say that because it's simple today. If we need more capabilities, I could see that being a pattern. Not sure how far we'll need to go.

src/bicep-registry/README.md
flanakin: We could, but then that means we have 2 conventions for the same pattern, right? Isn't that more confusing? Fwiw, we don't need the `//`. We could simplify it a couple of ways:
```bicep
// @tenant: 'value3'
// 'value3' @tenant
```
Only reason I like it as-is would be consistency with line 180. Both 180 and 182 are the same, minus one being commented out to ensure valid Bicep code.

src/bicep-registry/README.md
flanakin: The problem on this line is I couldn't think of a way to differentiate a true comment (like this line) and a commented out conditional block. I agree `////` feels odd. I thought about other characters for comments, but didn't think that made sense with indented comments (see line 219). This felt more consistent and didn't require custom knowledge.
```bicep
// @subscription
//// Standard comment
// @resourceGroup 
//   // Indented comment

// @subscription
//! Non-standard character
// @resourceGroup 
//   ! Indented comment
```
The good news is whatever we choose is internal to us. We're not building a public API. If we were, I'd prefer to evaluate the line to determine if it's a comment or not.

src/scripts/Build-Bicep.ps1
flanakin: Yeah, I thought about that. I was trying to keep it simple :-P I suspect we will need to expand it in the future. A couple other options:
1. Auto-detect supported scopes from the entire file.
2. Look up supported scopes on the targetScope line only:
   ```bicep
   targetScope: 'subscription' // @resourceGroup, @tenant
   ```
Both of those might be better than requiring the first line to define scopes.

src/scripts/Build-Bicep.ps1
flanakin: I'm using `while` because conditional blocks require stepping thru the lines sequentially in a single loop. I couldn't do that from a `foreach` loop.

src/scripts/Build-Bicep.ps1
flanakin: No. We only append the examples to the generated file, so the original file is always example-less.

src/scripts/Deploy-Toolkit.ps1
flanakin: The Bicep Registry already has a mechanism for tracking deployments, so we shouldn't need one. That said, we do need telemetry outside of the Bicep Registry (like for the finops-hub template). I added an issue to track this: #89.

src/scripts/Build-Bicep.ps1
flanakin: Changed it to infer scopes and added guidance to add a scope directive with all scopes on the targetScope line as a fallback.

src/bicep-registry/README.md
flanakin: Resolving this for now. Let's discuss further and determine if we'd rather use a different character.

src/bicep-registry/README.md
flanakin: Resolving this for now. Let's discuss further and determine if we'd rather use a function to be more consistent with Bicep directives.

Issue #59, opened 2023-02-19
Pipeline to dedupe and convert to parquet


flanakin:
## 📝 Scenario
As a **FinOps practitioner**, I need to **dedupe CM exports** in order to **directly link Power BI to ADLS as a data source**

## 💎 Solution
Create an **ms-cm-exports_Transform** pipeline to copy data from the **ms-cm-exports** container to the **ingestion** container in parquet format and overwrite old exports.
Start from https://github.com/MSBrett/ccm_lite/blob/main/ARM/azuredeploy.json and copy relevant pieces of https://github.com/MSBrett/ccm_lite/blob/main/Deploy-EA.ps1, as appropriate.

## 🙋‍♀️ Ask for the community

Let us know if there's anything you'd like to see changed.

PR #91, opened 2023-03-28
Add ADF dataset for ms-cm-exports folder

MSBrett:
## 🛠️ Description
Add ADF dataset for ms-cm-exports folder.

Fixes #(issue)
#59 

- [ ] 💪 Unit tests

### DIFF
--- a/src/modules/datasets/ms_cm_exports.bicep
+++ b/src/modules/datasets/ms_cm_exports.bicep
@@ -1,0 +1,63 @@
+@description('Required. The name of the parent Azure Data Factory.')
+param dataFactoryName string
+
+@description('Required. The storage account where the data resides.')
+param linkedServiceName string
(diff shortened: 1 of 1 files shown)


src/modules/datasets/ms_cm_exports.bicep
```diff
@@ -0,0 +1,63 @@
+@description('Required. The name of the parent Azure Data Factory.')
```
tsilvers-ms, line 1: Should we have a description of what this file is doing in the file itself?

tsilvers-ms (approved):

src/modules/datasets/ms_cm_exports.bicep
MSBrett: Good plan.  Done.

PR #93, opened 2023-03-28
An Azure Data Factory dataset for parquet files in Azure Blob Storage

MSBrett:
## 🛠️ Description
An Azure Data Factory dataset for parquet files in Azure Blob Storage

Fixes #(issue)
#59 

- [ ] 💪 Unit tests

### DIFF
--- a/src/modules/datasets/ingestion_parquet.bicep
+++ b/src/modules/datasets/ingestion_parquet.bicep
@@ -1,0 +1,60 @@
+// Description: An Azure Data Factory dataset for parquet files in Azure Blob Storage
+
+@description('Required. The name of the parent Azure Data Factory.')
+param dataFactoryName string
+
+@description('Required. The storage account where the data resides.')
+param linkedServiceName string
+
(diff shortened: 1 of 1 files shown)


src/modules/datasets/ingestion_parquet.bicep
```diff
@@ -0,0 +1,60 @@
...
+    parameters: {
+      fileName: {
+        type: 'String'
+      }
+      folderName: {
+        type: 'String'
+      }
+    }
+    type: datasetType
+    typeProperties: {
```
flanakin, line 34: Given this is still 90% the same as the other datasets, I'd merge this into one dataset.bicep and change the properties based on the datasetType. You can do something like:
```bicep
var csvProps = { ... }
var parquetProps = { ... }
var commonProps = { ... }
...
typeProperties: union(commonProps, datasetType == 'Parquet' ? parquetProps : csvProps)
```

flanakin (changes requested):

PR #113, opened 2023-04-05
Rename ms-cm-exports to msexports

flanakin:
## 🛠️ Description
Renamed `ms-cm-exports` to `msexports` to remove the dashes.

- [ ] 💪 Unit tests


src/modules/hub.bicep
```diff
@@ -86,7 +82,7 @@ module dataFactory 'Microsoft.DataFactory/factories/deploy.bicep' = {
 }
 
 resource storageAccountLookup 'Microsoft.Storage/storageAccounts@2021-06-01' existing = {
-  name: storageAccount.name
+  name: storageAccountName
```
flanakin, line 85: This fixes a bug that was causing the deployment to fail. `storageAccount.name` references the deployment name, not the storage account name.

src/reports/CommitmentDiscounts.pbix
flanakin, line 1: Updated the container name in both PBIX files

MSBrett (approved):

PR #117, opened 2023-04-10
Hub Module

MSBrett:
## 🛠️ Description
FinOps Hub Module
This module deploys datasets and pipelines to ingest cost management data generated by the export feature in Azure Cost Management, deduplicate/compress/transform it and store the results in Azure Data Lake.  Both Parquet (Default) and compressed CSV output are supported.

- [ ] 💪 Unit tests


src/modules/hub.bicep
```diff
@@ -62,49 +65,60 @@ resource defaultTelemetry 'Microsoft.Resources/deployments@2022-09-01' = if (ena
 }
 
 // ADLSv2 storage account for staging and archive
-module storageAccount 'Microsoft.Storage/storageAccounts/deploy.bicep' = {
```
flanakin, line 65: If we're not using this module anymore, can you delete it? Same with the others.

src/modules/hub.bicep
```diff
@@ -62,49 +65,60 @@ resource defaultTelemetry 'Microsoft.Resources/deployments@2022-09-01' = if (ena
...
 }
 
-module dataFactory 'Microsoft.DataFactory/factories/deploy.bicep' = {
-  name: 'dataFactory'
-  params: {
-    name: dataFactoryName
-    location: location
-    tags: resourceTags
-  }
+resource blobService 'Microsoft.Storage/storageAccounts/blobServices@2021-06-01' = {
```
flanakin, line 84: FYI - You can also nest these within the parent resource if you want to make them a child. Not sure which is preferred. Just mentioning this in case you didn't know. When you nest them, you can leave out the `parent`.

src/modules/hub.bicep
```diff
@@ -131,6 +145,207 @@ resource uploadSettingsJson 'Microsoft.Resources/deploymentScripts@2020-10-01' =
...
+          type: 'LinkedServiceReference'
+        }
+        secretName: storageAccountName
+      }
+    }
+  }
+}
+
+module msexports 'datasets/adlsv2.bicep' = {
+  name: 'msexports'
```
flanakin, line 239: Should we call this out as a dataset? (Same for the others.)

src/modules/hub.bicep
```diff
@@ -131,6 +145,207 @@ resource uploadSettingsJson 'Microsoft.Resources/deploymentScripts@2020-10-01' =
...
+    dataFactoryName: dataFactoryName
+    linkedServiceName: storageAccountName
+    compressionCodec: 'none'
+    datasetType: 'DelimitedText'
+    containerName: exportContainer.name
+  }
+}
+
+module ingestion_csv 'datasets/adlsv2.bicep' = {
+  name: 'ingestion_csv'
```
flanakin, line 254: In the call you mentioned we'd only support one format. Should we only create one dataset or do you want to eventually support multiple?

src/modules/hub.bicep
```diff
@@ -131,6 +145,207 @@ resource uploadSettingsJson 'Microsoft.Resources/deploymentScripts@2020-10-01' =
...
+    transform_parquet
+  ]
+  params: {
+    dataFactoryName: dataFactoryName
+    exportContainerName: exportContainer.name
+  }
+}
+
+module trigger 'triggers/adlsv2.bicep' = {
+  name: 'trigger'
```
flanakin, line 336: I suspect we'll have a separate trigger for ingestion. Should we make this `trigger_msexports`?
```suggestion
  name: 'trigger_msexports'
```

src/modules/hub.bicep
```diff
@@ -131,6 +145,207 @@ resource uploadSettingsJson 'Microsoft.Resources/deploymentScripts@2020-10-01' =
...
+  dependsOn: [
+    extract_parquet
```
flanakin, line 338: Should this be both parquet and CSV?

src/modules/hub.bicep
```diff
@@ -131,6 +145,207 @@ resource uploadSettingsJson 'Microsoft.Resources/deploymentScripts@2020-10-01' =
...
+  params: {
+    dataFactoryName: dataFactoryName
+    exportContainerName: exportContainer.name
+  }
+}
+
+module extract_csv 'pipelines/msexports_extract.bicep' = {
+  name: 'msexports_extract_csv'
+  dependsOn: [
+    transform_parquet
```
flanakin, line 327: Is this supposed to be CSV?
```suggestion
    transform_csv
```

src/modules/hub.bicep
```diff
@@ -131,6 +145,207 @@ resource uploadSettingsJson 'Microsoft.Resources/deploymentScripts@2020-10-01' =
...
+    exportContainerName: exportContainer.name
+  }
+}
+
+module extract_parquet 'pipelines/msexports_extract.bicep' = {
+  name: 'msexports_extract_parquet'
+  dependsOn: [
+    transform_parquet
+  ]
+  params: {
```
flanakin, line 318: No parameter for the file type? (Same for CSV below.)

src/modules/hub.bicep
```diff
@@ -131,6 +145,207 @@ resource uploadSettingsJson 'Microsoft.Resources/deploymentScripts@2020-10-01' =
...
+module transform_parquet 'pipelines/msexports_transform.bicep' = {
+  name: 'msexports_transform_parquet'
+  dependsOn: [
+    msexports
+    ingestion_parquet
+  ]
+  params: {
+    dataFactoryName: dataFactoryName
+    sinkDataset:  replace('${ingestionContainer.name}_Parquet', '-', '_')
+    sourceDataset:  replace('${exportContainer.name}_DelimitedText', '-', '_')
```
flanakin, line 292: nit: Should we put the source before the sink? I know it doesn't matter. Just thinking about it in logical order. Maybe the same for export container before ingestion? (Same for transform_csv.)

src/modules/hub.bicep
```diff
@@ -131,6 +145,207 @@ resource uploadSettingsJson 'Microsoft.Resources/deploymentScripts@2020-10-01' =
...
+module ingestion_csv 'datasets/adlsv2.bicep' = {
+  name: 'ingestion_csv'
+  dependsOn: [
+    linkedService_storageAccount
+    linkedService_keyVault
+  ]
+  params: {
+    dataFactoryName: dataFactoryName
+    linkedServiceName: storageAccountName
+    compressionCodec: 'gzip'
```
flanakin, line 262: nit: Since both are gzip compression, should we just default that?

src/modules/hub.bicep
MSBrett: Removed Storage Account Module

src/modules/hub.bicep
MSBrett: I need to do it this way so I can create the containers independently and refer to them by symbollic name when configuring the datasets and trigger.

src/modules/hub.bicep
MSBrett: Good catch.

src/modules/hub.bicep
MSBrett: Can do.

src/modules/hub.bicep
MSBrett: Done

src/modules/hub.bicep
MSBrett: I want to support both parquet and .csv.gz as the output.
Parquet will be the only path that's enabled, but by modifying the trigger one can toggle from parquet to .csv.gz output.
Or remove the delete action from the end of the csv.gz code path and have that execute before the parquet conversion so we can output both.

src/modules/hub.bicep
MSBrett: Done

src/modules/hub.bicep
MSBrett: Added the outputformat param.
Good call.

src/modules/hub.bicep
MSBrett: Done.

src/modules/hub.bicep
MSBrett: Done

flanakin (approved):

PR #135, opened 2023-05-05
Update Commitment discounts report to use parquet

### LINKED ISSUE #63: Update Power BI data sources to use parquet

flanakin:
## 🛠️ Description
Update Commitment discounts report to use parquet.

Fixes #63 

- [ ] 💪 Unit tests


flanakin:
@MSBrett I changed the storage account parameter to use the account name instead of the full URL. I'm thinking that would be simpler and easier to capture than the full URL. If you're okay with that, can you ack by signing off on the PR? If not, let me know why you'd prefer the full URL so I can document that and I'll get this updated.

src/templates/finops-hub/reports/CommitmentDiscounts.pbix
flanakin, line 1: @MSBrett I changed the storage account parameter to use the account name instead of the full URL. I'm thinking that would be simpler and easier to capture than the full URL. If you're okay with that, can you ack by signing off on the PR? If not, let me know why you'd prefer the full URL so I can document that and I'll get this updated.

src/templates/finops-hub/reports/CostSummary.pbix
flanakin, line 1: @MSBrett I changed the storage account parameter to use the account name instead of the full URL. I'm thinking that would be simpler and easier to capture than the full URL. If you're okay with that, can you ack by signing off on the PR? If not, let me know why you'd prefer the full URL so I can document that and I'll get this updated.

PR #137, opened 2023-05-09
Reorganize all bicep modules

flanakin:
## 🛠️ Description
Groups bicep code for each resource type together into fewer files. Saves 3m in deployment time! 🙌

- [ ] 💪 Unit tests


src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -0,0 +1,715 @@
...
+param storageAccountName string
+
+@description('Required. The name of the container where Cost Management data is exported.')
+param exportContainerName string
+
+@description('Required. The name of the container where normalized data is ingested.')
+param ingestionContainerName string
+
+@description('Optional. Indicates whether ingested data should be converted to Parquet. Default: true.')
+param convertToParquet bool = true
```
flanakin, line 21: I simplified the `datasetType` logic to just support opting in to converting to parquet. If we want to support more formats, we can always add them later, but the long-term goal is to get into a database, so I'm not too concerned.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -0,0 +1,715 @@
...
+resource pipeline_transformExport 'Microsoft.DataFactory/factories/pipelines@2018-06-01' = {
+  name: '${safeExportContainerName}_transform'
+  parent: dataFactory
+  dependsOn: [
+    dataset_msexports
+    dataset_ingestion
+  ]
+  properties: {
+    activities: [
+      // (start) -> Wait -> Scope -> Metric -> Date -> File -> Folder -> Delete Target -> Convert CSV -> Delete CSV -> (end)
```
flanakin, line 345: FYI - I added this to help document the code. It's not sent to ADF. The other comments above each step are just to identify which section is which when collapsed.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -0,0 +1,715 @@
...
+              'Completed'
+            ]
+          }
+        ]
+        userProperties: []
+        typeProperties: {
+          variableName: 'metric'
+          value: {
+            // TODO: Parse metric out of the export path with self-managed exports -- value: '@first(split(split(pipeline().parameters.folderName, \'/\')[sub(length(split(pipeline().parameters.folderName, \'/\')), 4)], \'-\'))'
+            value: 'amortizedcost'
```
flanakin, line 394: FYI - Hard-coding for v0.0.1. Left the original code commented out so we can bring it back in v0.0.2 with self-managed exports.

_Side note: I just saw the PR to allow dashes and underscores in export names, so that should be fixed in the next week or two._

MSBrett: Love it.  Tested it.  Looks great.

MSBrett (approved):
Looks good to me.
Ran my test script with a couple of different folder names for the export and it all worked great.  Parquet output is also working now.

Issue #60, opened 2023-02-19
[Hubs] Managed exports – Create exports on behalf of customers


flanakin:
## 📝 Scenario
As a **hub admin**, I need to **automate CM export creation** in order to **minimize administrative overhead**

## 💎 Solution
Create a `msexports_Setup` pipeline in ADF to manage Cost Management data exports automatically. Use `config\settings.json` to store the list of exports that will be managed.

## 📗 Documentation updates

1. Remove steps 2 and 3 (create and run exports) in [Create a new hub](https://github.com/microsoft/cloud-hubs/blob/dev/docs/finops-hub/README.md#-create-a-new-hub).
2. Add the following paragraph before the info message under the "Access the cost data from custom tools" bullet in [Get started with hubs](https://github.com/microsoft/cloud-hubs/blob/dev/docs/finops-hub/README.md#-get-started-with-hubs).
   > Data is ingested into the `msexports` container and transformed when moved into the `ingestion` container. Do not use the `msexports` container for anything other than Cost Management exports. If manipulating data, please do that in the `ingestion` container after the transform pipeline completes. Don't remove or rename columns, as that can break Power BI reports.
3. Remove the comment about needing to setup Cost Management exports on the [FinOps hub template](https://github.com/microsoft/cloud-hubs/blob/dev/docs/finops-hub/template.md) description.

## 🙋‍♀️ Ask for the community

We could use your help:
1. Please vote this issue up (👍) to prioritize it.
3. Leave comments to help us solidify the vision.

flanakin:
This shipped in FinOps hubs 0.4

PR #161, opened 2023-05-19
Config Dataset

MSBrett:
## 🛠️ Description
Dataset for the settings.json file in the config container for #60 - Self Managed Exports.

Fixes #(issue)

- [ ] 💪 Unit tests

### DIFF
--- a/src/templates/finops-hub/modules/dataFactory.bicep
+++ b/src/templates/finops-hub/modules/dataFactory.bicep
@@ -17,9 +17,8 @@ param exportContainerName string
 @description('Required. The name of the container where normalized data is ingested.')
 param ingestionContainerName string
 
-@description('Optional. The name of the container where normalized data is ingested.')
-param configContainerName string = 'config'
-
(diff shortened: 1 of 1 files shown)


src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -17,6 +17,10 @@ param exportContainerName string
 @description('Required. The name of the container where normalized data is ingested.')
 param ingestionContainerName string
 
+@description('Optional. The name of the container where normalized data is ingested.')
+param configContainerName string = 'config'
```
flanakin, line 21: nit: Do you want to out a default? Since this is internal, it's fine for us to make it required to protect us from future changes if we rename something.

flanakin (approved):

src/templates/finops-hub/modules/dataFactory.bicep
MSBrett: Good point.

PR #162, opened 2023-05-20
Managed Exports

MSBrett:
## 🛠️ Description
Adds the required ADF pipelines and triggers for #60 to create or update exports in Azure Cost Management.
Triggered whenever an update is made to settings.json.

Expects the following schema for exportScopes:
~~~
...
    "exportScopes": [
        {
             "scope": "00000000-0000-0000-0000-000000000000"
        }
    ]
...
~~~
Fixes #(issue)

- [ ] 💪 Unit tests


src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -314,6 +320,458 @@ resource trigger_exportContainer 'Microsoft.DataFactory/factories/triggers@2018-
...
+    }
+  }
+}
+
+//------------------------------------------------------------------------------
+// Pipelines
+//------------------------------------------------------------------------------
+
+//------------------------------------------------------------------------------
+// Azure Cost Management add export pipeline
```
flanakin, line 361:
```suggestion
// Microsoft Cost Management add export pipeline
```

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -314,6 +320,458 @@ resource trigger_exportContainer 'Microsoft.DataFactory/factories/triggers@2018-
...
+// Triggered when settings.json is updated.
+// Creates an export in Azure Cost Management.
```
flanakin, line 363:
```suggestion
// Creates an export in Cost Management.
```

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -314,6 +320,458 @@ resource trigger_exportContainer 'Microsoft.DataFactory/factories/triggers@2018-
...
+                    }
+                  }
+                  {
+                    name: 'Set Storage Account URI Suffix'
+                    type: 'SetVariable'
+                    dependsOn: []
+                    userProperties: []
+                    typeProperties: {
+                      variableName: 'storageAccountUriSuffix'
+                      value: 'dfs.core.usgovcloudapi.net'
```
flanakin, line 481: Is this right?

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -314,6 +320,458 @@ resource trigger_exportContainer 'Microsoft.DataFactory/factories/triggers@2018-
...
+                  dependencyConditions: [
+                    'Succeeded'
+                  ]
+                }
+              ]
+              userProperties: []
+              typeProperties: {
+                variableName: 'exportName'
+                value: {
+                  value: '@if(greater(length(replace(tolower(concat(pipeline().parameters.StorageAccountName,\'oamortized\', last(split(item().scope, \'/\')))), \'-\', \'\')), 64), substring(replace(tolower(concat(pipeline().parameters.StorageAccountName,\'oamortized\', last(split(item().scope, \'/\')))), \'-\', \'\'), 0, 63), replace(tolower(concat(pipeline().parameters.StorageAccountName,\'oamortized\', last(split(item().scope, \'/\')))), \'-\', \'\'))'
```
flanakin, line 653: Can we move the metric to the end? Maybe `FinOpsHub-${hubName}-${metric}`? We can remove the prefix if it's too long. My goal is just to group them together in the list. We probably don't care about the scope or storage account names in this.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -314,6 +320,458 @@ resource trigger_exportContainer 'Microsoft.DataFactory/factories/triggers@2018-
...
+                  }
+                }
+                body: {
+                  value: '{\n  "properties": {\n    "schedule": {\n      "status": "Active",\n      "recurrence": "Monthly",\n      "recurrencePeriod": {\n        "from": "@{utcNow()}",\n        "to": "2099-12-31T00:00:00Z"\n      }\n    },\n    "partitionData": "True",\n    "format": "Csv",\n    "deliveryInfo": {\n      "destination": {\n        "resourceId": "@{pipeline().parameters.StorageAccountId}",\n        "container": "msexports",\n        "rootFolderPath": "@{item().scope}"\n      }\n    },\n    "definition": {\n      "type": "amortizedcost",\n      "timeframe": "TheLastBillingMonth",\n      "dataSet": {\n        "granularity": "Daily"\n      }\n    }\n  }\n}'
+                  type: 'Expression'
+                }
+              }
+            }
+            {
+              name: 'Set Closed Month Export Name'
```
flanakin, line 696: nit: Do we really need separate exports for open, closed, and historical months? Would it make sense to only have one and just change the dates as we need them. Fwiw, we could also use that to schedule the export to happen when we want from a scheduled ADF pipeline instead of waiting for exports to run. Just a thought. Not a big deal either way.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -314,6 +320,458 @@ resource trigger_exportContainer 'Microsoft.DataFactory/factories/triggers@2018-
...
+                  dependencyConditions: [
+                    'Succeeded'
+                  ]
+                }
+              ]
+              userProperties: []
+              typeProperties: {
+                variableName: 'exportName'
+                value: {
+                  value: '@if(greater(length(replace(tolower(concat(pipeline().parameters.StorageAccountName,\'camortized\', last(split(item().scope, \'/\')))), \'-\', \'\')), 64), substring(replace(tolower(concat(pipeline().parameters.StorageAccountName,\'camortized\', last(split(item().scope, \'/\')))), \'-\', \'\'), 0, 63), replace(tolower(concat(pipeline().parameters.StorageAccountName,\'camortized\', last(split(item().scope, \'/\')))), \'-\', \'\'))'
```
flanakin, line 710: nit: If we do keep separate exports, "c" isn't very clear what it is. Should we explicitly say "closed"? Again, just a nit.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -314,6 +320,458 @@ resource trigger_exportContainer 'Microsoft.DataFactory/factories/triggers@2018-
...
+// Pipelines
+//------------------------------------------------------------------------------
+
+//------------------------------------------------------------------------------
+// Azure Cost Management add export pipeline
+// Triggered when settings.json is updated.
+// Creates an export in Azure Cost Management.
+//------------------------------------------------------------------------------
+resource pipeline_setup 'Microsoft.DataFactory/factories/pipelines@2018-06-01' = {
+  name: '${safeExportContainerName}_setup'
```
flanakin, line 366: Thinking ahead to when we support other clouds, should this be `config_setup` and not tied to `msexports` explicitly?

flanakin (approved):

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -11,6 +11,9 @@ param keyVaultName string
 @description('Required. The name of the Azure storage account instance.')
 param storageAccountName string
 
+@description('Required. The resource Id of the Azure storage account instance.')
+param storageAccountId string
```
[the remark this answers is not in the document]
MSBrett: Exports need the storage account in the format of /subscriptions/<SubscriptionID>/resourceGroups/<RGName>/providers/Microsoft.Storage/storageAccounts/<StorageAccount>
I don't have a way of determining either the subscription ID or resource group name at runtime.

src/templates/finops-hub/modules/dataFactory.bicep
MSBrett: I was thinking we could create othercloud_setup pipeline and subscribe it to the same trigger.
The msexports_setup pipeline will ignore export scopes which don't have the "cloud" attribute set to AzureCloud or AzureUSGovernment.

src/templates/finops-hub/modules/dataFactory.bicep
MSBrett: Doing the closed month export is a recommendation from the PM for the export feature to catch any adjustments that get made once the books close.
I'm planning on disabling the schedule for all the cost management exports and having ADF trigger them per it's schedule rather than waiting for CM to run.

src/templates/finops-hub/modules/dataFactory.bicep
MSBrett: Good point.  Unneeded now.

src/templates/finops-hub/modules/dataFactory.bicep
MSBrett: Resolved

src/templates/finops-hub/modules/dataFactory.bicep
MSBrett: Resolved

src/templates/finops-hub/modules/dataFactory.bicep
MSBrett: Resolved

src/templates/finops-hub/modules/dataFactory.bicep
MSBrett: I do need to differentiate between daily (open) and monthly (closed) exports, so how about:
${hubName}-daily-${metric} 
${hubName}-monthly-${metric}

MSBrett:
- Fixed naming of exports.
- Use managed identity of the data factory for auth rather than storage account key.
- Remove KeyVault module.
- Simplified the pipeline a LOT.
- Removed the "Cloud" and "TenantID" parameters entirely - exportScope schema = 
    "exportScopes": [
        {
             "scope": "00000000-0000-0000-0000-000000000000"
        }
    ]

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -102,6 +108,21 @@ resource identityRoleAssignments 'Microsoft.Authorization/roleAssignments@2022-0
   }
 }]
 
+// Create managed identity to manage exports in cost management
```
flanakin, line 111: nit: Do we need a CM Contributor role in the list of roles? (Basing this on the comment. May also want to change the `storageRbacRoles` variable name as well.)

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -72,6 +68,12 @@ var autoStartRbacRoles = [
   '673868aa-7521-48a0-acc6-0f60742d39f5' // Data Factory contributor - https://learn.microsoft.com/azure/role-based-access-control/built-in-roles#data-factory-contributor
 ]
 
+var storageRbacRoles = [
```
flanakin, line 71: Do you want to merge the scripts into one pre-deployment script? Not sure if everything can run before... 🤔 (Probably better in a separate PR.)

flanakin (approved):

src/templates/finops-hub/modules/dataFactory.bicep
MSBrett: To create an export the identity needs Storage Account Contributor and Reader permissions.
For ADF to read/write data to/from the storage account it needs Storage Blob Data Contributor permissions.
The identity needs to be granted EA/MCA/CM Contributor permissions against the export scope before adding it to settings.json.

src/templates/finops-hub/main.bicep
```diff
@@ -29,6 +29,13 @@ param exportRetentionInDays int = 0
 @description('Optional. Number of months of cost data to retain in the ingestion container. Default: 13.')
 param ingestionRetentionInMonths int = 13
 
+@description('Optional. Remote storage account for ingestion dataset.')
+param hubStorageAccountUri string = ''
```
flanakin, line 33: nit: Can we be more explicit about this being the remote storage account? Maybe `remoteHubStorageUri`? Same for the key.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -168,15 +178,49 @@ resource stopHubTriggers 'Microsoft.Resources/deploymentScripts@2020-10-01' = {
...
 
+resource linkedService_keyVault 'Microsoft.DataFactory/factories/linkedservices@2018-06-01' = {
+  name: keyVaultName
+  parent: dataFactory
+  properties: {
+    annotations: []
+    parameters: {}
+    type: 'AzureKeyVault'
+    typeProperties: {
+      baseUrl: 'https://${keyVaultName}${environment().suffixes.keyvaultDns}/'
```
flanakin: The template analyzer used by AQT doesn't like when we generate URLs. Can we get this as a property from the KeyVault instance? If not, maybe leave a comment that we can't get it from properties and we have to generate it.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -168,15 +178,49 @@ resource stopHubTriggers 'Microsoft.Resources/deploymentScripts@2020-10-01' = {
...
+    parameters: {}
+    type: 'AzureBlobFS'
+    typeProperties: {
+      url: reference('Microsoft.Storage/storageAccounts/${storageAccount.name}', '2021-08-01').primaryEndpoints.dfs
+    }
+  }
+}
+
+resource linkedService_hubStorageAccount 'Microsoft.DataFactory/factories/linkedservices@2018-06-01' = if (!empty(hubStorageAccountUri)) {
+  name: 'hubStorageAccount'
```
flanakin, line 208: nit: Same comment for remote hubs: `remoteHubStorage`?

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -244,105 +288,966 @@ resource dataset_ingestion 'Microsoft.DataFactory/factories/datasets@2018-06-01'
...
+              typeProperties: {
+                variableName: 'thisMonth'
+                value: {
+                  value: '@variables(\'nextMonth\')'
+                  type: 'Expression'
+                }
+              }
+            }
+            {
+              name: 'update export end date'
```
flanakin, line 626:
```suggestion
              name: 'Update export end date'
```

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -76,17 +95,26 @@ resource dataFactory 'Microsoft.DataFactory/factories@2018-06-01' existing = {
...
+
 //------------------------------------------------------------------------------
-// Stop all triggers before deploying
+// Identities and RBAC
 //------------------------------------------------------------------------------
 
 // Create managed identity to start/stop triggers
 resource identity 'Microsoft.ManagedIdentity/userAssignedIdentities@2023-01-31' = {
-  name: '${dataFactoryName}_${exportContainerName}_extract_triggerManager'
+  name: '${dataFactory.name}_${exportContainerName}_extract_triggerManager'
```
flanakin, line 114: Should this be generic for all triggers? Or are you wanting to leave it as-is given it's already been created?
```suggestion
  name: '${dataFactory.name}_triggerManager'
```

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -179,147 +235,1364 @@ resource linkedService_storageAccount 'Microsoft.DataFactory/factories/linkedser
...
+        }
+      }
+    ]
+    type: 'ScheduleTrigger'
+    typeProperties: {
+      recurrence: {
+        frequency: 'Hour'
+        interval: 24
+        startTime: '2023-01-01T01:01:00'
+        timeZone: 'Pacific Standard Time'
```
flanakin, line 412: This is going to cause concerns from my team because it means everyone using hubs will pull data at the same time, which will look like a DDoS attempt. Do we have a way to naturally distribute it? Could we set the timezone to be based on the resource location? We may need PowerShell for that.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -179,147 +235,1364 @@ resource linkedService_storageAccount 'Microsoft.DataFactory/factories/linkedser
...
+  }
+}
+
+//------------------------------------------------------------------------------
+// Pipelines
+//------------------------------------------------------------------------------
+
+//------------------------------------------------------------------------------
+// Microsoft Cost Management backfill exports
+// Triggered by ???.
```
flanakin, line 460: Should we replace `???`?

src/templates/finops-hub/main.bicep
```diff
@@ -29,6 +29,13 @@ param exportRetentionInDays int = 0
 @description('Optional. Number of months of cost data to retain in the ingestion container. Default: 13.')
 param ingestionRetentionInMonths int = 13
 
+@description('Optional. Remote storage account for ingestion dataset.')
```
flanakin, line 32: Just to confirm, this is a remote hub for you to **PUSH** data to, right?
```suggestion
@description('Optional. Storage account to push data to for ingestion into a remote hub.')
```

src/templates/finops-hub/main.bicep
```diff
@@ -29,6 +29,13 @@ param exportRetentionInDays int = 0
...
+param remoteHubStorageUri string = ''
+
+@description('Optional. Storage account key for remote storage account.')
```
flanakin, line 35:
```suggestion
@description('Optional. Storage account key to use when pushing data to a remote hub.')
```

flanakin (dismissed):

src/templates/finops-hub/modules/dataFactory.bicep
MSBrett: Really good point.
I've added a module to determine the timezone based on the location.  It defaults to UTC for undefined regions.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -358,24 +1638,16 @@ resource pipeline_extractExport 'Microsoft.DataFactory/factories/pipelines@2018-
...
 //------------------------------------------------------------------------------
-
 resource pipeline_transformExport 'Microsoft.DataFactory/factories/pipelines@2018-06-01' = {
   name: '${safeExportContainerName}_transform'
   parent: dataFactory
-  dependsOn: [
-    dataset_msexports
-    dataset_ingestion
-  ]
   properties: {
```
flanakin, line 1649: Given your workflow is to work in DF Studio and copy the properties into bicep, would it be more efficient to store each of these in separate JSON files and pull them in via `loadJsonContent()`? This can be a separate PR. Just calling it out in case that would streamline the dev flow.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -179,147 +242,1364 @@ resource linkedService_storageAccount 'Microsoft.DataFactory/factories/linkedser
...
+      finOpsHub: {
+        type: 'String'
+        defaultValue: hubName
+      }
+      resourceManagementUri: {
+        type: 'String'
+        defaultValue: environment().resourceManager
       }
-    }
-    type: any(convertToParquet ? 'Parquet' : 'DelimitedText')
```
flanakin, line 225: If we're forcing parquet, we should get rid of the `convertToParquet` parameter.

src/templates/finops-hub/modules/hub.bicep
```diff
@@ -121,13 +128,16 @@ resource dataFactory 'Microsoft.DataFactory/factories@2018-06-01' = {
 module dataFactoryResources 'dataFactory.bicep' = {
   name: 'dataFactoryResources'
   params: {
-    dataFactoryName: dataFactoryName
+    dataFactoryName: dataFactory.name
     convertToParquet: convertToParquet
```
flanakin, line 132: If we're forcing parquet, shouldn't we remove this option? This can be a separate PR.

src/templates/finops-hub/modules/keyVault.bicep
```diff
@@ -14,8 +14,9 @@ param location string = resourceGroup().location
 @description('Optional. Array of access policies object.')
 param accessPolicies array = []
 
-@description('Required. Name of the storage account to store access keys for.')
-param storageAccountName string
+@description('Optional. Name of the storage account to store access keys for.')
```
flanakin, line 17: Is this outdated?
```suggestion
@description('Optional. Storage account key to store as a secret.')
```
(Feel free to use a better description)

flanakin (approved):

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -179,147 +242,1364 @@ resource linkedService_storageAccount 'Microsoft.DataFactory/factories/linkedser
...
+    typeProperties: datasetPropsCommon
+    linkedServiceName: {
+      parameters: {}
+      referenceName: linkedService_storageAccount.name
+      type: 'LinkedServiceReference'
+    }
+  }
+}
 
 resource dataset_msexports 'Microsoft.DataFactory/factories/datasets@2018-06-01' = {
```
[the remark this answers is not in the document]
MSBrett: Yes.  ADF now uses it's managed identity to access the storage account.  We still use a key for remote ingestion containers.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -401,12 +1672,11 @@ resource pipeline_transformExport 'Microsoft.DataFactory/factories/pipelines@201
         typeProperties: {
           variableName: 'scope'
           value: {
-            value: '@replace(split(pipeline().parameters.folderName,split(pipeline().parameters.folderName, \'/\')[sub(length(split(pipeline().parameters.folderName, \'/\')), 4)])[0],\'${exportContainerName}\',\'${ingestionContainerName}\')'
+            value: '@replace(split(pipeline().parameters.folderName,split(pipeline().parameters.folderName, \'/\')[sub(length(split(pipeline().parameters.folderName, \'/\')), 4)])[0],\'msexports\',\'ingestion\')'
```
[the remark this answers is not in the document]
MSBrett: Yes.  I realized I'd hard coded those names into the transform.  Now they will update correctly if we change container names.

src/templates/finops-hub/modules/keyVault.bicep
flanakin: In that case, we should probably name it more clearly. I was about to ask that. Guess I should've followed thru with my initial instincts 😜

PR #171, opened 2023-05-31
Add function Save-FinOpsHubTemplate

aromano2:
## 🛠️ Description

Fixes #(issue)

- [ ] 💪 Unit tests

### DIFF
--- a/src/powershell/FinOpsToolkit.psm1
+++ b/src/powershell/FinOpsToolkit.psm1
@@ -1,28 +1,145 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+# Import the localized Data
+$script:localizedData = Import-LocalizedData -FileName 'FinOpsToolkit.strings.psd1' -BaseDirectory $PSScriptRoot
+
+#region Private functions
+<#
+    .SYNOPSIS
+        Creates a directory if it does not already exist.
+
+    .PARAMETER Path
+        Path to create directory.
+#>
+function New-Directory
+{
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSUseShouldProcessForStateChangingFunctions', '')]
+    [CmdletBinding()]
+    param
+    (
+        [Parameter(Mandatory = $true)]
+        [string]
+        $Path
+    )
+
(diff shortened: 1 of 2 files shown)


src/powershell/FinOpsToolkit.psm1
```diff
@@ -1,28 +1,144 @@
...
+        [string]
+        $Destination = $env:temp,
+
+        [Parameter()]
+        [switch]
+        $Force
+    )
+
+    New-Directory -Path $Destination
+    $releases = Get-FinOpsToolkitVersion -Latest:$($Version -eq 'Latest') -Preview:$Preview
```
flanakin, line 78: Is `$Version -eq 'Latest'` case-sensitive? The value you set it to is lowercase above.

src/powershell/FinOpsToolkit.psm1
```diff
@@ -1,28 +1,144 @@
...
+
+    if ($Version -eq 'Latest')
```
flanakin, line 80: Shouldn't you factor `$Preview` into this logic?

flanakin (approved):

src/powershell/FinOpsToolkit.psm1
```diff
@@ -1,28 +1,144 @@
...
+        Save-FinOpsHubTemplate
+
+        Downloads the latest version of FinOps hub template to current users' temp folder.
+
+    .EXAMPLE
+        Save-FinOpsHubTemplate -Version '1.0.0' -Destination 'C:\myHub' -Force
+
+        Downloads version 1.0.0 of FinOpsHub template to c:\myHub directory. It will overwrite an existing 1.0.0.zip file if it exists.
+#>
+function Save-FinOpsHubTemplate
```
flanakin, line 55: Should we add tests for this? I know we don't have the infrastructure for that yet, but even if it's a placeholder, that might still be nice until we get the infrastructure in place. Thoughts?

src/powershell/FinOpsToolkit.psm1
aromano2: it is, I pass it in to Get-FinOpsToolkitVersion

PR #181, opened 2023-06-11
Add ADX cluster/db resources

flanakin:
## 🛠️ Description
Creates a Data Explorer cluster and database.

Fixes #(issue)

- [ ] 💪 Unit tests

### DIFF
--- a/src/templates/finops-hub/modules/dataExplorer.bicep
+++ b/src/templates/finops-hub/modules/dataExplorer.bicep
@@ -1,0 +1,134 @@
+//==============================================================================
+// Parameters
+//==============================================================================
+
+@description('Required. Name of the hub. Used to ensure unique resource names.')
+param hubName string
+
+@description('Optional. Location for all resources.')
+param location string = resourceGroup().location
+
+@description('Optional. Resource tags.')
+param tags object = {}
+
+@description('Required. The name of the Azure Key Vault instance.')
+param keyVaultName string
+
+//------------------------------------------------------------------------------
+// Variables
+//------------------------------------------------------------------------------
+
+// Generate globally unique cluster name: 4-22 chars; lowercase letters and numbers
+var safeHubName = replace(replace(toLower(hubName), '-', ''), '_', '')
+var clusterName = replace('${take(safeHubName, 22)}', '--', '-')
+
+//==============================================================================
+// Resources
+//==============================================================================
+
+resource cluster 'Microsoft.Kusto/clusters@2022-12-29' = {
+  name: clusterName
+  location: location
+  tags: tags
+  sku: {
+    capacity: 1
+    name: 'string'
+    tier: 'string'
+  }
+  identity: {
+    type: 'string'
+    userAssignedIdentities: {}
+  }
+  properties: {
+    acceptedAudiences: [
+      {
+        value: 'string'
+      }
+    ]
+    allowedFqdnList: [
+      'string'
+    ]
+    allowedIpRangeList: [
+      'string'
+    ]
+    enableAutoStop: true
+    enableDiskEncryption: false
+    enableDoubleEncryption: false
+    enablePurge: false
+    enableStreamingIngest: false
+    engineType: 'V3'
+    keyVaultProperties: {
+      keyName: keyVaultName
+      keyVaultUri: 'string'
+      keyVersion: 'string'
+      userIdentity: 'string'
+    }
+    languageExtensions: {
+      value: [
+        {
+          languageExtensionImageName: 'string'
+          languageExtensionName: 'string'
+        }
+      ]
+      value: [
+        {
+          languageExtensionImageName: 'string'
+          languageExtensionName: 'string'
+        }
+      ]
+    }
+    optimizedAutoscale: {
+      isEnabled: bool
+      maximum: int
+      minimum: int
+      version: int
+    }
+    publicIPType: 'string'
+    publicNetworkAccess: 'string'
+    restrictOutboundNetworkAccess: 'string'
+    trustedExternalTenants: [
+      {
+        value: 'string'
+      }
+    ]
+    virtualClusterGraduationProperties: 'string'
+    virtualNetworkConfiguration: {
+      dataManagementPublicIpId: 'string'
+      enginePublicIpId: 'string'
+      subnetId: 'string'
+    }
+  }
+  zones: [
+    'string'
+  ]
+}
+
+resource database 'Microsoft.Kusto/clusters/databases@2022-12-29' = {
+  name: 'Hub'
+  location: location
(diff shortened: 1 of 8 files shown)


src/templates/finops-hub/modules/dataExplorer.bicep
```diff
@@ -0,0 +1,134 @@
...
+resource cluster 'Microsoft.Kusto/clusters@2022-12-29' = {
+  name: clusterName
+  location: location
+  tags: tags
+  sku: {
+    capacity: 1
+    name: 'string'
+    tier: 'string'
+  }
+  identity: {
```
flanakin, line 38: Do we need this? If so, what do we need to set?

src/templates/finops-hub/modules/dataExplorer.bicep
```diff
@@ -0,0 +1,134 @@
...
+    type: 'string'
+    userAssignedIdentities: {}
+  }
+  properties: {
```
flanakin, line 42: Need help with all these properties...

src/templates/finops-hub/modules/dataExplorer.bicep
```diff
@@ -0,0 +1,134 @@
...
+      subnetId: 'string'
+    }
+  }
+  zones: [
+    'string'
+  ]
+}
+
+resource database 'Microsoft.Kusto/clusters/databases@2022-12-29' = {
+  name: 'Hub'
```
flanakin, line 107: Open to suggestions about the db name. I haven't put any thought into it yet...

src/templates/finops-hub/modules/dataExplorer.bicep
```diff
@@ -0,0 +1,134 @@
...
+  location: location
+  parent: cluster
+  kind: 'ReadWrite'
+  properties: {
```
flanakin, line 111: How are these used?

src/templates/finops-hub/modules/dataExplorer.bicep
```diff
@@ -0,0 +1,134 @@
...
+    hotCachePeriod: 'string'
+    softDeletePeriod: 'string'
+  }
+}
+
+//==============================================================================
+// Outputs
```
flanakin, line 118: Anything else critical to be used after someone deploys the template?

src/templates/finops-hub/modules/dataExplorer.bicep
slavikn84: You only need these if you want to use Managed Identities (MIs). On behalf of who will the ingestions into Kusto run? If you plan to use managed identities for this, then you'll need to put the MI details here.

src/templates/finops-hub/modules/dataExplorer.bicep
```diff
@@ -0,0 +1,134 @@
...
+
+//==============================================================================
+// Resources
+//==============================================================================
+
+resource cluster 'Microsoft.Kusto/clusters@2022-12-29' = {
+  name: clusterName
+  location: location
+  tags: tags
+  sku: {
```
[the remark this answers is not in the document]
slavikn84: Bassel - please provide your input.

src/templates/finops-hub/modules/dataExplorer.bicep
slavikn84: Bassel - please provide your input.

src/templates/finops-hub/modules/dataExplorer.bicep
slavikn84: How about CostManagement?

src/templates/finops-hub/modules/dataExplorer.bicep
slavikn84: Michael - I'd set both to "30.00:00:00" (30 days).
Bassel - please correct me if I'm wrong about the format.

src/templates/finops-hub/modules/dataExplorer.bicep
slavikn84: Bassel - please provide your input.

src/templates/finops-hub/modules/dataExplorer.bicep
flanakin: You tell me 🙂 What's ideal? I'm assuming we probably do want a managed identity to run the ingestion. I don't know enough to make that call, tho. Thoughts? If we do it, can you share what this should be set to?

src/templates/finops-hub/modules/dataExplorer.bicep
flanakin: Long-term, the database will host more than just cost information. That's why I was starting with something more generic. Not sure if there are specific recommendations around when to split data into separate databases.

src/templates/finops-hub/modules/dataExplorer.bicep
flanakin: Bassel, please resolve if you ack. Otherwise, let me know what you think we should set these to and/or if we should make it explicitly configurable by the customer.

src/templates/finops-hub/modules/dataExplorer.bicep
slavikn84: Let's talk about it.

src/templates/finops-hub/modules/dataExplorer.bicep
slavikn84: Splitting tables across different databases will not affect performance. Having separate databases allows you to have better contol of permissions (since these are set per database, not per table), if you think it's valuable to you.

src/templates/finops-hub/modules/dataExplorer.bicep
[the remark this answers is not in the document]
basaba: AFAIR tier is not required, only name

src/templates/finops-hub/modules/dataExplorer.bicep
basaba: There are many props, which of them isn't clear, you can leave the default values

src/templates/finops-hub/modules/dataExplorer.bicep
basaba: you can use hotCachePeriodInDays and softDeletePeriodInDays, see https://learn.microsoft.com/en-us/azure/data-explorer/create-cluster-database?tabs=arm

src/templates/finops-hub/modules/dataExplorer.bicep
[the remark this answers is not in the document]
flanakin: @slavikn84 can you find the allowed values and recommend a good one for the default (probably free tier)?

PR #228, opened 2023-08-01
Build script for FinOpsToolkit module

aromano2:
## 🛠️ Description

Fixes #(issue)

- [ ] 💪 Unit tests


src/scripts/Build-FinOpsPsModule.ps1
```diff
@@ -0,0 +1,78 @@
+param
+(
+    [Parameter(Mandatory = $true)]
+    [string]
+    [ValidateScript({$_ -match '^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$'})]
+    $Version,
+
+    [Parameter()]
+    [ValidateSet('alpha', 'beta')]
```
flanakin, line 9: We usually use "preview" instead of "beta". 
```suggestion
    [ValidateSet('alpha', 'preview')]
```

src/scripts/Build-FinOpsPsModule.ps1
```diff
@@ -0,0 +1,78 @@
...
+    [string]
+    $PrereleaseTag
+)
+
+$rootPath = (Get-Item -Path $PSScriptRoot).Parent.Parent.FullName
+$moduleName = 'FinOpsToolkit'
+$moduleFullName = "$moduleName.psm1"
+$modulePath = Join-Path -Path $rootPath -ChildPath "src\powershell\$moduleFullName"
```
flanakin, line 17: Use forward slashes instead of backward slashes so the script works on Mac. Same below.

flanakin (changes requested):

src/scripts/Build-FinOpsPsModule.ps1
aromano2: Join-Path converts directory separators for you depending on the environment you're using, even if the child path has backward slashes in string. Do we want to standardize this going forward seeing as how a lot of other scripts use \ instead of /?

src/powershell/FinOpsToolkit.psm1
```diff
@@ -2,9 +2,10 @@
 # Licensed under the MIT License.
 
 # Import the localized Data
-$script:localizedData = Import-LocalizedData -FileName 'FinOpsToolkit.strings.psd1' -BaseDirectory $PSScriptRoot
+$script:localizedData = Import-LocalizedData -FileName 'FinOpsToolkit.strings.psd1' -BaseDirectory (Join-Path -Path $PSScriptRoot -ChildPath 'en-US')
```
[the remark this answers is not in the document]
aromano2: to be honest, I haven't supported localized strings other than en-Us before. If we think it would be necessary then absolutely we should start tracking

flanakin (approved):

src/scripts/Build-FinOpsPsModule.ps1
aromano2: PowerShell repo doesn't seem consistent either 
image
I'll just use forward slashes

PR #232, opened 2023-08-11
Script for building powershell module documentation

aromano2:
## 🛠️ Description

Fixes #(issue)

- [ ] 💪 Unit tests


src/powershell/FinOpsToolkit.psm1
```diff
@@ -117,23 +117,23 @@ function Save-FinOpsHubTemplate
 
 <#
     .SYNOPSIS
-        Retrieves available version numbers of the FinOps toolkit.
```
flanakin, line 120: What's the standard here? I've seen spaces added and now spaces removed. What's the official guidance?

src/scripts/Build-ModuleDoc.ps1
```diff
@@ -0,0 +1,210 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+<#
+    .SYNOPSIS
+    Builds a README file in markdown for a specified module. Includes sytnax, description, required modules, parameters and examples.
```
flanakin, line 6:
```suggestion
    Builds a README file in markdown for a specified module. Includes sytnax, description, required modules, parameters, and examples.
```

src/scripts/Build-ModuleDoc.ps1
flanakin, line 1: Should this just be `Build-PowerShell.ps1`? Then anything else we want to build when building PowerShell can be done together. Thoughts? I mainly want to avoid "Module" since that's generic and applies to other things, like Bicep, but I'm also keen to keep it open-ended for other future build tasks.

flanakin, line 1: Should this just be `Build-PowerShell.ps1`? Then anything else we want to build when building PowerShell can be done together. Thoughts? I mainly want to avoid "Module" since that's generic and applies to other things, like Bicep, but I'm also keen to keep it open-ended for other future build tasks.

src/scripts/Build-ModuleDoc.ps1
```diff
@@ -0,0 +1,210 @@
...
+    Path to a .psm1 module file.
+
+    .PARAMETER Name
+    Name of a PowerShell module. (Must be discoverable by Get-Module -ListAvailable)
+
+    .PARAMETER OutputPath
+    Path to folder for resulting .md file. Defaults to $pwd.
+
+    .PARAMETER OutputFileName
+    Name for the resulting .md file. Defaults to README.md
```
flanakin, line 18: Can we hard-code all of these? I don't mind targeting a single module, but we only plan to have one, so we probably don't need that today. The output path should be hard-coded so we don't have to specify it every time we call the script, like we have for other build scripts.

src/scripts/Build-ModuleDoc.ps1
```diff
@@ -0,0 +1,210 @@
...
+        [Parameter(Mandatory = $true)]
+        [string]
+        $Function
+    )
+
+    $help = Get-Help -Name $Function -Full
+    $excludeCommonParameters = [System.Management.Automation.PSCmdlet]::CommonParameters + [System.Management.Automation.PSCmdlet]::OptionalCommonParameters
+    $examples = @()
+    $parameters = @()
+    foreach ($parameter in $help.parameters.parameter.Where({$excludeCommonParameters -notcontains $_.Name}))
```
flanakin, line 87: nit: Shouldn't these be Pascal-cased? Same for the properties on lines 90-4, 101-3, etc.
```suggestion
    foreach ($parameter in $help.parameters.parameter.Where({$excludeCommonParameters -NotContains $_.Name}))
```

src/scripts/Build-ModuleDoc.ps1
```diff
@@ -0,0 +1,210 @@
...
+# Check for manifest, if it exists add Description and required modules
+$manifestPath = Get-ChildItem -Path (Split-Path -Path $modulePath -Parent) -Filter '*.psd1'
+if ($manifestPath)
+{
+    $manifest = Import-PowerShellDataFile -Path $manifestPath.FullName
+    [void]$sb.AppendLine($manifest.Description).Appendline()
+
+    if ($manifest.RequiredModules)
+    {
+        [void]$sb.AppendLine('## Required Modules').AppendLine()
```
flanakin, line 155:
```suggestion
        [void]$sb.AppendLine('## Required modules').AppendLine()
```

src/scripts/Build-ModuleDoc.ps1
```diff
@@ -0,0 +1,210 @@
...
+        {
+            [void]$sb.AppendLine("$($requiredModule.ModuleName) | $($requiredModule.ModuleVersion)")
+        }
+
+        [void]$sb.AppendLine()
+    }        
+}
+
+# Parse functions
+[void]$sb.AppendLine('## Functions').AppendLine()
```
flanakin, line 169: nit: Should we use the term "commands" to generalize and avoid the difference between a cmdlet and a function, which doesn't matter to end users?
```suggestion
[void]$sb.AppendLine('## Commands').AppendLine()
```

src/scripts/Build-ModuleDoc.ps1
```diff
@@ -0,0 +1,210 @@
...
+    if (-not [String]::IsNullOrEmpty($functionHelp.syntax))
+    {
+        [void]$sb.AppendLine('#### Syntax').AppendLine()
+        [void]$sb.AppendLine("``$($functionHelp.Syntax)``").AppendLine()
+    }
+
+    if ($functionHelp.Parameters.Count -gt 0)
+    {
+        [void]$sb.AppendLine('#### Parameters').AppendLine()
+        [void]$sb.AppendLine('Name | Type | Description | Required? | Default Value')
```
flanakin, line 185:
```suggestion
        [void]$sb.AppendLine('Name | Type | Description | Required? | Default value')
```

flanakin, line 185: Should required be before the description? I feel like that's the typical order. Fwiw, all of our parameter docs put that first in the description.

src/scripts/Build-ModuleDoc.ps1
```diff
@@ -0,0 +1,210 @@
...
+        [void]$sb.AppendLine('|:---:|:---:|---|:---:|:---:|')
```
flanakin, line 186: Doesn't this center everything? I don't think we want any of them centered, right?
```suggestion
        [void]$sb.AppendLine('|---|---|---|---|---|')
```

src/scripts/Build-ModuleDoc.ps1
flanakin, line 1: Can you run this and generate the content in the same PR so we can see what it looks like?

flanakin (changes requested):

src/powershell/FinOpsToolkit.psm1
aromano2: No space is the correct way. Get-Help doesn't parse the keywords correctly if there are spaces. Which was news to me

src/scripts/Build-ModuleDoc.ps1
aromano2: Yes it centers, I'm adding the example output to PR

src/scripts/Build-ModuleDoc.ps1
aromano2: @flanakin added Example.README,md

src/scripts/Build-ModuleDoc.ps1
aromano2: The inputs will be provided by a build task. I don't mind hard-coding I'm just used to making everything generic.

src/scripts/Build-ModuleDoc.ps1
aromano2: I centered everything except the description. Want me to keep default alignment?

src/scripts/Build-ModuleDoc.ps1
flanakin: Is the build task going to also submit a PR to update the docs? All these docs live in the repo in code.

src/scripts/Build-ModuleDoc.ps1
flanakin: Yeah, left-aligned is probably better. The Required column could go either way, I suppose, but the others are variable length text that would probably be best left-aligned.

src/scripts/Build-ModuleDoc.ps1
flanakin: Instead of publishing an example, can we just publish the real docs in the `docs/powershell` folder? That's what I meant. Sorry, I should've been more explicit. 

Although, if we do that, then we won't be able to release until the module is published 🤔 Should we move this PR to the `features/powershell` branch so it's not blocking a release? (Either that or plan to release the module next week. It's okay if we add more later.)

src/scripts/Build-ModuleDoc.ps1
aromano2: We can do that yes. I haven’t gotten that far yet though. Initially I was thinking it be a manual process or vs code task. But if you’re ok with an auto pr I think that’d be better.

src/scripts/Build-ModuleDoc.ps1
aromano2: I re-based to features/powershell and added doc

src/scripts/Build-ModuleDoc.ps1
aromano2: After looking into hard-coding, I think we should wait until the pipeline is built since things will likely change in terms of pathing. Also the module would need to be "built" first before this can run.

flanakin (changes requested):

PR #244, opened 2023-08-15
Implement InvokeBuild and PSDepend

aromano2:
## 🛠️ Description

Fixes #(issue)

- [ ] 💪 Unit tests


.build/BuildHelper/Build-FinOpsModule.ps1
```diff
@@ -0,0 +1,82 @@
...
+        [string]
+        $PrereleaseTag
+    )
+
+    $rootPath = (Get-Item -Path $PSScriptRoot).Parent.FullName
+    $moduleName = 'FinOpsToolkit'
+    $moduleFullName = "$moduleName.psm1"
+    $modulePath = Join-Path -Path $rootPath -ChildPath "src/powershell/$moduleFullName"
+    $stringsPath = Join-Path -Path $rootPath -ChildPath 'src/powershell/en-US'
+    $releasePath = Join-Path -Path $rootPath -ChildPath "release/FinOpsToolkit/$Version"
```
flanakin, line 22: Should we label this as PowerShell? Is there ever a scenario where we'd build an old version that isn't the current one?

.build/BuildHelper/Build-FinOpsModule.ps1
```diff
@@ -0,0 +1,82 @@
...
+    ./New-Directory $releasePath
+    Pop-Location
+
+    $manifestProperties = @{
+        ModuleVersion     = $Version
+        Path              = $manifestPath
+        Guid              = '00f120b5-2007-6120-0000-b03e1254e770'
+        Author            = 'Microsoft Corporation'
+        CompanyName       = 'Microsoft Corporation'
+        Description       = 'FinOps toolkit - Cmdlets to deploy and manage resusable FinOps solutions in Azure. This module is compatible with PowerShell and Windows PowerShell. For more information about the FinOps toolkit, please visit https://aka.ms/finops/toolkit'
```
flanakin, line 49: nit: "Commands" instead of "cmdlets"?
```suggestion
        Description       = 'FinOps toolkit - Commands to deploy and manage resusable FinOps solutions in the Microsoft Cloud. This module is compatible with PowerShell and Windows PowerShell. For more information about the FinOps toolkit, please visit https://aka.ms/finops/toolkit'
```

.build/BuildHelper/Build-FinOpsModule.ps1
```diff
@@ -0,0 +1,82 @@
...
+                ModuleVersion = '2.11.1'
+            },
+            @{
+                ModuleName    = 'Az.Resources'
+                ModuleVersion = '6.5.1'
+            }
+        )
+        IconUri           = 'https://raw.githubusercontent.com/microsoft/cloud-hubs/4747859b1c800f49d5b0c3cd7894a40fe8641c3a/src/images/FinOpsHubs.svg'
+        LicenseUri        = 'https://github.com/microsoft/cloud-hubs/blob/dev/LICENSE'
+        ProjectUri        = 'https://github.com/microsoft/cloud-hubs'
```
flanakin, line 69:
```suggestion
        IconUri           = 'https://raw.githubusercontent.com/microsoft/cloud-hubs/4747859b1c800f49d5b0c3cd7894a40fe8641c3a/src/images/FinOpsHubs.svg'
        LicenseUri        = 'https://github.com/microsoft/finops-toolkit/blob/dev/LICENSE'
        ProjectUri        = 'https://github.com/microsoft/finops-toolkit'
```

.build/BuildHelper/New-Directory.ps1
flanakin, line 1: Should we reference src/New-Directory.ps1 instead of duplicating it here?

flanakin (changes requested):

.build/BuildHelper/Build-FinOpsModule.ps1
aromano2: Version will be computed at build time with [GitVersion](https://gitversion.net/url). At least that's what I was planning on doing.

.build/BuildHelper/New-Directory.ps1
aromano2: I copied instead of moving because other scripts use it. My goal was to make an internal module so that we don't have to worry about pathing for calling scripts. I was going to move other build-related scripts into this module but figured smaller pr was better.

.build/BuildHelper.psm1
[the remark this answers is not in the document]
aromano2: Yes I will work on that

.build/BuildHelper/Build-FinOpsModule.ps1
flanakin: My point is that we'll only ever build one version and everything in the release folder is that same version, so it probably isn't needed to add the version as a subfolder. Just a nit, I suppose.

.build/BuildHelper/New-Directory.ps1
flanakin: Wouldn't an internal module be more complicated because we'd have to continually re-install it? That sounds like more trouble than switching directories to run a script. That said, we can work around that through other means, like NPM scripts. I wonder if that'd be simpler.

.build/BuildHelper/Build-FinOpsModule.ps1
aromano2: Oh I see what you mean. Modules have versioned subfolders. This way you can have multiple versions of any one module installed together

.build/BuildHelper/New-Directory.ps1
aromano2: You don’t have to install it. You can manually import it, or add it to your powershell environment path so it loads automatically. If you prefer to keep them as scripts then I can change it. To me its safer to centralize functionality in a module, than have someone down the line move one file that breaks 10 different things in a release.

aromano2: Also want to point out that this module would only be for build automation purposes. Most of the code will only ever run on build agents.

.build/BuildHelper.psm1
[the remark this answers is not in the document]
aromano2: @flanakin added documentation

PR #260, opened 2023-08-23
Get-FinOpsHub

aromano2:
## 🛠️ Description

Fixes #(issue)

- [ ] 💪 Unit tests


src/powershell/Public/Get-FinOpsHub.ps1
```diff
@@ -0,0 +1,95 @@
...
+
+    if ($resourceMatches.Count -gt 0)
+    {
+        $output = @()
+        $groups = $resourceMatches | Group-Object -Property 'HubId'
+        foreach ($group in $groups)
+        {
+            $groupProperties = [ordered]@{
+                Name = $group.Group.Name | Select-Object -Unique
+                HubId = $group.Group.HubId | Select-Object -Unique
```
flanakin, line 86: Should this be `Id` like other resources? Given it's not a real resource, I'm not 100% sure one way or another. For that reason, `HubId` may be better? Thinking out loud 🤔 I'm fine with whichever everyone feels makes sense.

src/powershell/Tests/Unit/FinOpsToolkit.Tests.ps1
flanakin, line 1: nit: This is probably better for a separate PR, but should we split the hub tests out into a separate file?

src/templates/finops-hub/modules/hub.bicep
```diff
@@ -62,6 +62,7 @@ var finOpsToolkitVersion = '0.0.1'
 
 resource defaultTelemetry 'Microsoft.Resources/deployments@2022-09-01' = if (enableDefaultTelemetry) {
   name: 'pid-${telemetryId}-${uniqueString(deployment().name, location)}'
+  tags: resourceTags
```
flanakin, line 65: No tags on this one. This is not really a release.

src/templates/finops-hub/modules/keyVault.bicep
```diff
@@ -86,6 +86,7 @@ resource storageRef 'Microsoft.Storage/storageAccounts@2022-09-01' existing = {
 resource keyVault_secrets 'Microsoft.KeyVault/vaults/secrets@2022-11-01' = {
   name: storageRef.name
   parent: keyVault
+  tags: tags
```
flanakin, line 89: Do secrets need tags? We should be good with only the vault instance, right?

flanakin (approved):

src/powershell/Public/Get-FinOpsHub.ps1
```diff
@@ -0,0 +1,95 @@
...
+        $ResourceGroupName = '*'
+    }
+    
+    $context = Get-AzContext
+    if (-not $context)
+    {
+        throw $script:localizedData.ContextNotFound
+    }
+
+    $tagTemplate = '/subscriptions/{0}/resourceGroups/{1}/providers/Microsoft.Cloud/hubs/{2}'
```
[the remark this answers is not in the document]
aromano2: -like is case insensitive

src/templates/finops-hub/modules/keyVault.bicep
aromano2: I just put them on anything that supported tags. I agree we probably don't need them on secrets

src/powershell/Tests/Unit/FinOpsToolkit.Tests.ps1
aromano2: I split this one out for now and will follow up with splitting the rest out

src/templates/finops-hub/modules/hub.bicep
aromano2: removed

src/powershell/Public/Get-FinOpsHub.ps1
aromano2: I'm going to leave as is for now until we figure out the path forward

PR #264, opened 2023-08-24
FinOps - Register Resource Providers

saadmsft:
## 🛠️ Description

Registers Azure resource providers required for FinOps hub.
Optional. Shows what would happen if the command runs without actually running it.

EXAMPLE
Register-FinOpsHubProviders -WhatIf

Shows what would happen if the command runs without actually running it.

NOTES
This command registers the following Azure resource providers required for FinOps hub:
- Microsoft.EventGrid
- Microsoft.CostManagementExports

Fixes #(issue)
#248

- [Yes] 🫰 PS -WhatIf / az validate
- [Yes] 👍 Manually deployed + verified
- [Yes] 💪 Unit tests [Currently added only 2 tests]


src/powershell/Public/RegisterFinOpsHubProviders.ps1
```diff
@@ -0,0 +1,55 @@
...
+
+.NOTES
+    This command registers the following Azure resource providers required for FinOps hub:
+    - Microsoft.EventGrid
+    - Microsoft.CostManagementExports
+#>
+function Register-FinOpsHubProviders {
+    [CmdletBinding()]
+    param (
+        [Switch]$WhatIf
```
aromano2, line 23: PowerShell has a built-in mechanism to provide this functionality. It would be a CmdletBinding attribute: [CmdletBinding(SupportsShouldProcess)]. https://learn.microsoft.com/en-us/powershell/scripting/learn/deep-dives/everything-about-shouldprocess?view=powershell-7.3#supportsshouldprocess

src/powershell/Public/RegisterFinOpsHubProviders.ps1
```diff
@@ -0,0 +1,55 @@
...
+    )
+
+    # Define the resource providers to register
+    $providers = "Microsoft.EventGrid", "Microsoft.CostManagementExports"
+
+    try {
```
aromano2, line 29: Any exception generated in this try block would get lost without a catch block. Also need to make sure any of these commands actually produce a terminating error vs a non-terminating error.

src/powershell/Public/RegisterFinOpsHubProviders.ps1
```diff
@@ -0,0 +1,55 @@
...
+                    Write-Output "WhatIf: Registering resource provider $provider"
+                }
+                else {
+                    Write-Verbose "Registering resource provider $provider"
+                    Register-AzResourceProvider -ProviderNamespace $provider   
+                }
+            }
+            # If the provider is already registered, log a message saying so
+            else {
+                Write-Verbose "Resource provider $provider is already registered"
```
aromano2, line 47: we should move any logging strings to the localized data in the en-US folder

src/powershell/Public/RegisterFinOpsHubProviders.ps1
```diff
@@ -0,0 +1,55 @@
...
+            }
+        }
+    }
+    finally {
+        # Clean up any resources that were used
+        $registered = $null
```
aromano2, line 53: Don't need this

src/powershell/Public/RegisterFinOpsHubProviders.ps1
saadmsft: As of now if we try out these commands, except would error would be produced only when the namespace is wrong which is not the case in our code but happy to catch the exception. Forgot to add the piece of code. 

do you think if namespaces are provided already in code rather than a param from the user we should remove try-catch-finally

src/powershell/en-US/FinOpsToolkit.strings.psd1
```diff
@@ -9,4 +9,6 @@ ConvertFrom-StringData -StringData @'
     FoundLatestRelease = Found latest release '{0}'.
     NewDirectory = Creating directory '{0}'.
     TemplateNotFound = Could not find template 'main.bicep' at path '{0}'.
+    RegisterProvider =  Registering resource provider {0}.
+    ResourceProviderExists =  Resource provider {0} is already registered. 
 '@
```
saadmsft, line 14: @aromano2  Please see if there is any issue with it.

src/powershell/Public/RegisterFinOpsHubProviders.ps1
saadmsft: please have a look now

src/powershell/Public/RegisterFinOpsHubProviders.ps1
saadmsft: Changed it.

src/powershell/Public/RegisterFinOpsHubProviders.ps1
saadmsft: Tried to do that please check if it would suffice. Added string to psd1 file

src/powershell/Public/RegisterFinOpsHubProviders.ps1
saadmsft: Removed

src/powershell/Public/RegisterFinOpsHubProviders.ps1
```diff
@@ -0,0 +1,52 @@
...
+                }
+            }
+            # If the provider is already registered, logging a message saying so
+            else {
+                Write-Verbose -Message $($LocalizedData.ResourceProviderExists -f $provider)
+            }
+        }
+    }
+    catch {
+        Write-Error $_.Exception.Message
```
flanakin, line 50: Should we rethrow the full exception?

flanakin (changes requested):

src/powershell/Public/RegisterFinOpsHubProviders.ps1
```diff
@@ -0,0 +1,52 @@
...
+
+.PARAMETER WhatIf
+    Optional. Shows what would happen if the command runs without actually running it.
+
+.EXAMPLE
+    Register-FinOpsHubProviders -WhatIf
+
+    Shows what would happen if the command runs without actually running it.
+
+.NOTES
```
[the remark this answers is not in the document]
saadmsft: Added Description

src/powershell/Public/RegisterFinOpsHubProviders.ps1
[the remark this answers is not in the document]
saadmsft: Renamed

src/powershell/Public/RegisterFinOpsHubProviders.ps1
saadmsft: changed it to, [Microsoft.Azure.PowerShell.Cmdlets.ResourceManager.Runtime.AzurePowerShellException]

PR #288, opened 2023-09-01
Add telemetry to workbook template

flanakin:
## 🛠️ Description
Add telemetry to workbook template so we can measure custom deployments.

- [ ] 💪 Unit tests


sebassem (approved):
LGTM

PR #703, opened 2024-05-14
Added release logic to Optimization Engine

helderpinto:
## 🛠️ Description

Changed build, package and publish scripts to include the required assets for the Optimization Engine release and deployment process. The `docs/deploy` folder includes files that are just copies coming from the `src/optimization-engine` folder.

Also included minor bug fixes in workbooks and runbooks.

> - [ ] 🤏 Lint tests
> - [ ] 💪 Unit tests
> - [ ] 🙌 Integration tests

> - [X] ❎ Not needed (small/internal change)

> - [X] ❎ Not needed (small/internal change)


src/scripts/Build-Toolkit.ps1
```diff
@@ -118,3 +118,28 @@ Get-ChildItem "$PSScriptRoot/../templates/$Template*" -Directory -ErrorAction Si
 
     Write-Host ''
 }
+
+# Package optimization engine
```
flanakin, line 122: nit: Should we move this into a separate Build-OptimizationEngine script? 🤔 I'm fine keeping it here. Just thinking out loud. Fwiw, the templates should probably be moved to a Build-Template script as well. Alternatively, we can keep it here for now and define a more general approach to solve this consistently. I'm fine with anything. Just wanted to share that to see what your thoughts are.

src/scripts/Package-Toolkit.ps1
```diff
@@ -88,10 +88,34 @@ $templates = Get-ChildItem $relDir -Directory `
         }
     }
 
-    # Copy azuredeploy.json to docs/deploy folder
     Write-Verbose "Updating $($path.Name) deployment file in docs..."
-    Copy-Item "$path/azuredeploy.json" "$deployDir/$($path.Name)-$version.json"
-    Copy-Item "$path/azuredeploy.json" "$deployDir/$($path.Name)-latest.json"
+
+    $packageManifestPath = "$path/package-manifest.json"
+    if (Test-Path $packageManifestPath)
```
flanakin, line 94: nit: Would it be better to define a fallback manifest instead of just falling back to the previous logic? That way, we always work with a manifest, we just define a default one. I'm okay to not do that or maybe just add a TODO comment. Just a thought that crossed my mind.

src/scripts/Publish-Toolkit.ps1
```diff
@@ -95,7 +95,7 @@ function Find-Repo($config, [string]$templateName)
 $ver = & "$PSScriptRoot/Get-Version.ps1"
 
 # Loop thru templates
-Get-ChildItem "$relDir/$Template*" -Directory `
+Get-ChildItem "$relDir/$Template*" -Directory -Exclude "optimization-engine" `
```
flanakin, line 98: Should we add a `# TODO` comment above this line to remove the exclusion after we determine how we're going to publish AOE to other repos? I don't think we want this to be the end state.

flanakin (approved):

docs/deploy/optimization-engine/azuredeploy-nested.bicep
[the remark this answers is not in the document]
helderpinto: Yes, I agree with you, especially to be able to deploy all the runbooks of some past version. I am going to commit a new version of the Package-Toolkit script & package-manifest.json to accommodate this.

docs/deploy/finops-hub-0.2.1-rc.2.json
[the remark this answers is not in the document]
helderpinto: Next commit will revert these files.

src/scripts/Build-Toolkit.ps1
helderpinto: Well, if we move templates and AOE to a different script, why don't we move workbooks and Bicep registry modules as well? Maybe each toolkit component type should have its own build script following some naming pattern and then we had the Build-Toolkit script calling all of them, provided the component directory (templates, workbooks, aoe, etc.) had a build script... But I'd keep it as is for now. I am adding a TODO to review the logic. If you prefer, I can open an issue for this.

src/scripts/Package-Toolkit.ps1
helderpinto: Yes, I agree. I was about to include this change, but then I thought: where do we define the fallback manifest? Inline or as a file placed somewhere in the src structure? As I didn't want to make this decision now, I am leaving a TODO comment for now.

PR #716, opened 2024-05-29
added changes to register event grid provider

### LINKED ISSUE #696: [Hubs] Register the EventGrid RP as part of the template
## 🐛 Problem
Data is not ingested if someone forgets to register the EventGrid RP.

## 👣 Repro steps
1. On a new subscription, confirm the EventGrid RP is not registered.
2. Deploy the FinOps hubs template.
3. Go to the ADF resource > Launch Studio > Manage > Triggers.
4. Verify the msexports_FileAdded trigger is Started (it should be stopped)
5. Start the trigger, the select the **Publish all** button to commit that change (this should fail)
6. Open the notifications pane and select **View details** to see why it failed

The error should explain that the EventGrid RP hasn't been registered.

## 🤔 Expected
Automate the EventGrid RP registration as part of template deployment.

## ℹ️ Additional context
While ARM does a good job of registering RPs for resources that are included in a template, ADF creates the EventGrid resource behind the scenes when the trigger is started. ADF doesn't register the RP, so it needs to be done explicitly.

We need to investigate if there is a way to declare some RP is needed as part of the template so ARM will auto-register. If not, we can create a script that registers the desired RPs. We can leverage the Initialize-FinOpsHubDeployment command if deployment scripts support installing modules from the PowerShell gallery. Alternatively, we just do the direct RP registration ourselves. The nice part about using our command is that it will cover any new pre-deployment steps that come up over time. 

## 🙋‍♀️ Ask for the community

We could use your help:

didayal-msft:
## 🛠️ Description

Fixes # 
#Fixes #696
## 📷 Screenshots
image

> - [ ] 🤏 Lint tests
> - [ ] 💪 Unit tests
> - [ ] 🙌 Integration tests

> - [ ] ❎ Not needed (small/internal change)

> - [ ] ❎ Not needed (small/internal change)

### DIFF
--- a/src/templates/finops-hub/modules/hub.bicep
+++ b/src/templates/finops-hub/modules/hub.bicep
@@ -40,16 +40,20 @@ param enableDefaultTelemetry bool = true
 // Add cm-resource-parent to group resources in Cost Management
 var finOpsToolkitVersion = loadTextContent('ftkver.txt')
 var resourceTags = union(tags, {
-    'cm-resource-parent': '${resourceGroup().id}/providers/Microsoft.Cloud/hubs/${hubName}'
-    'ftk-version': finOpsToolkitVersion
-    'ftk-tool': 'FinOps hubs'
-  })
+  'cm-resource-parent': '${resourceGroup().id}/providers/Microsoft.Cloud/hubs/${hubName}'
+  'ftk-version': finOpsToolkitVersion
+  'ftk-tool': 'FinOps hubs'
+})
 
 // Generate globally unique Data Factory name: 3-63 chars; letters, numbers, non-repeating dashes
 var uniqueSuffix = uniqueString(hubName, resourceGroup().id)
 var dataFactoryPrefix = '${replace(hubName, '_', '-')}-engine'
 var dataFactorySuffix = '-${uniqueSuffix}'
-var dataFactoryName = replace('${take(dataFactoryPrefix, 63 - length(dataFactorySuffix))}${dataFactorySuffix}', '--', '-')
+var dataFactoryName = replace(
+  '${take(dataFactoryPrefix, 63 - length(dataFactorySuffix))}${dataFactorySuffix}',
+  '--',
+  '-'
+)
 
 // The last segment of the telemetryId is used to identify this module
 var telemetryId = '00f120b5-2007-6120-0000-40b000000000'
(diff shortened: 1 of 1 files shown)


src/templates/finops-hub/modules/hub.bicep
```diff
@@ -83,6 +87,70 @@ resource defaultTelemetry 'Microsoft.Resources/deployments@2022-09-01' = if (ena
...
+
+resource RegisterEventGridProvider 'Microsoft.EventGrid/namespaces@2023-12-15-preview' = {
+  name: '${uniqueSuffix}-registerRP'
+  location: location
+  sku: {
+    capacity: 1
+    name: 'Standard'
+  }
+  properties: {
+    publicNetworkAccess: 'Enabled'
```
flanakin, line 102: Given some orgs may have a policy to disable public network access, should we set this to disabled or whatever the most restrictive value is?

MSBrett: Could we put the register provider PS commands into the script we use to start the triggers after deployment rather than deploying a dummy event grid and deleting it afterwards? We need to register the provider before starting the trigger in ADF.  It's not a dependency for deploying the trigger, just starting it.

src/templates/finops-hub/modules/hub.bicep
```diff
@@ -83,6 +87,70 @@ resource defaultTelemetry 'Microsoft.Resources/deployments@2022-09-01' = if (ena
...
+  kind: 'AzureCLI'
+  identity: {
+    type: 'UserAssigned'
+    userAssignedIdentities: {
+      '${cleanupidentity.id}': {}
+    }
+  }
+  properties: {
+    azCliVersion: '2.0.80'
+    scriptContent: 'az resource delete --id ${RegisterEventGridProvider.id} --verbose'
```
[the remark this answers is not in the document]
MSBrett: We should put the provider registration into the start triggers script and remove this.  It's not necessary till that point.

MSBrett (changes requested):
Could we move the bits to register the provider into the start-triggers.ps1 file?
The provider doesn't need to be registered until we're ready to start the triggers and it will save us having to deploy and delete a bunch of resources to accomplish the same task.

src/templates/finops-hub/modules/hub.bicep
flanakin: The goal here is to avoid the registration step because registration requires subscription-level access. If we deploy an EventGrid resource, then ARM will do the registration automatically.

MSBrett (approved):

src/templates/finops-hub/modules/hub.bicep
didayal-msft: done

flanakin (approved):

PR #805, opened 2024-07-11
feat: Create materialized view for FocusClean table

Springstone:
This commit adds a new materialized view called FocusClean on the table Focus. The view is created in the adxTableSchema.kql file and adxViews.kql file. The purpose of this view is to summarize the data in the Focus table based on various columns such as BillingAccountId, ChargePeriodStart, CommitmentDiscountId, Region, ResourceId, SkuPriceId, SubAccountId, x_AccountOwnerId, x_CostCenter, x_InvoiceSectionId, x_SkuDetails, x_SkuMeterId, and x_SkuOfferId. The view is created with the option to backfill existing data.

### DIFF
--- a/src/templates/finops-hub/modules/adxTableSchema.kql
+++ b/src/templates/finops-hub/modules/adxTableSchema.kql
@@ -95,7 +95,7 @@
   x_SkuTier:string
   )
 
-.create materialized-view FocusClean on table Focus
+.create-or-alter materialized-view FocusClean on table Focus
 {
     Focus
     | extend IngestionTime = ingestion_time()
--- a/src/templates/finops-hub/modules/adxTableSchema.kql
+++ b/src/templates/finops-hub/modules/adxTableSchema.kql
@@ -94,3 +94,11 @@
   x_SkuTerm:string,
   x_SkuTier:string
   )
+
+.create materialized-view FocusClean on table Focus
+{
+    Focus
+    | extend IngestionTime = ingestion_time()
+    | summarize arg_max(IngestionTime , *) by BillingAccountId, ChargePeriodStart, CommitmentDiscountId, Region, ResourceId, SkuPriceId, SubAccountId, x_AccountOwnerId, x_CostCenter, x_InvoiceSectionId, x_SkuDetails, x_SkuMeterId,
+x_SkuOfferId
+}
--- a/src/templates/finops-hub/modules/adxViews.kql
+++ b/src/templates/finops-hub/modules/adxViews.kql
@@ -1,0 +1,9 @@
+// Use this on an existing cluster to create a materialized view for the Focus table based on all the existing data.
+
+.create materialized-view with (backfill=true) FocusClean on table Focus
+{
+    Focus
+    | extend IngestionTime = ingestion_time()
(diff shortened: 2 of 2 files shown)


src/templates/finops-hub/modules/adxViews.kql
flanakin, line 1: Do we need this in addition to the other one? Seems like we should only need one since they do the same thing, right?

src/templates/finops-hub/modules/adxTableSchema.kql
```diff
@@ -94,3 +94,11 @@
   x_SkuTerm:string,
   x_SkuTier:string
   )
+
+.create-or-alter materialized-view FocusClean on table Focus
+{
+    Focus
+    | extend IngestionTime = ingestion_time()
+    | summarize arg_max(IngestionTime , *) by BillingAccountId, ChargePeriodStart, CommitmentDiscountId, Region, ResourceId, SkuPriceId, SubAccountId, x_AccountOwnerId, x_CostCenter, x_InvoiceSectionId, x_SkuDetails, x_SkuMeterId,
```
flanakin, line 102: Do we need `by BillingAccountId, ...`? We know those don't uniquely identify rows, so why are they used? Do we need every column or can we remove by "by" segment?

flanakin (approved):

src/templates/finops-hub/modules/adxTableSchema.kql
Springstone: No this is required, as that collection of columns uniquely identifies the meter/SKU/resource for the day it is charged (ChargePeriodStart). Arg_Max takes the latest entry (so on July 10, it will take the July 9 entry) based on those columns... since there is no way to uniquely identify a row. This is working incredibly well at the moment, based on 10 days of data so far.

src/templates/finops-hub/modules/adxViews.kql
Springstone: This one is optional, we don't deploy this. This is for people who already have a bunch of exports processed and want to ingest them to ADX.

src/templates/finops-hub/modules/adxTableSchema.kql
flanakin: If that's the case, this won't work since there's no combination of columns that uniquely identify a row. Wouldn't this result in some records getting left out if they have the same values for each of those columns?

FYI - There are actually more columns needed to identify uniqueness (e.g., Tags), but even if you had all of them, that still won't be unique since there are multiple rows with these same values per day.

src/templates/finops-hub/modules/adxViews.kql
flanakin: If this doesn't get deployed, it's probably better off in documentation. Otherwise, it'll sit in source and never get referenced or used and likely get outdated.

src/templates/finops-hub/modules/adxTableSchema.kql
flanakin: I'll sign off on this, but we still need to revisit it as I don't think this is correct.

flanakin (approved):
Approving so we can get it into the feature branch with the note that we still need to review this more thoroughly.

PR #982, opened 2024-09-14
Add react-ts Vite template

flanakin:
## 🛠️ Description
Adds the Vite react-ts template into `src/web`.

> - [ ] 🤏 Lint tests
> - [ ] 💪 Unit tests
> - [ ] 🙌 Integration tests


src/web/.gitignore
flanakin, line 1: @Ivanmtta Should we merge this with the root .gitignore file? Most of these are surprisingly not covered in that file.

src/web/package.json
flanakin, line 1: @Ivanmtta This adds a second package.json nested under the one at the root. For simplicity, I'm thinking we should probably only have one for now. What do you think? I can move these dependencies up or we can leave the nested packages if you feel that's best. 

An alternative might be to have a structure like this:
<root>
- tools
  - package.json
- web
  - package.json

Do you have a preference?

src/web/src/App.tsx
flanakin, line 1: @Ivanmtta I don't love having `src/web/src`. Would it make sense to move files up one level? That's the main reason why I had another comment about moving web to the root and renaming the current `src` to `tools`. Or should I just accept the two src folders?

Ivanmtta (approved):

Ivanmtta:
Looks good! :)

PR #1071, opened 2024-10-20
added normalize.css package as reset css

akiskips:
## 🛠️ Description
Added normalize.css package from https://github.com/necolas/normalize.css/
Added import at main.tsx

Fixes # 

> - [ ] 🤏 Lint tests
> - [ ] 💪 Unit tests
> - [ ] 🙌 Integration tests

### DIFF
--- a/src/web/package.json
+++ b/src/web/package.json
@@ -12,7 +12,6 @@
   },
   "dependencies": {
     "eslint-plugin-react": "^7.37.1",
-    "normalize.css": "^8.0.1",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
     "react-router-dom": "^6.27.0",
--- a/src/web/package.json
+++ b/src/web/package.json
@@ -12,6 +12,7 @@
   },
   "dependencies": {
     "eslint-plugin-react": "^7.37.1",
+    "normalize.css": "^8.0.1",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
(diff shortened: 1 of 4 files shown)


akiskips:
@Ivanmtta Not sure if it would be better to add an @import to index.css instead

flanakin (approved):

src/web/package.json
```diff
@@ -12,6 +12,7 @@
   },
   "dependencies": {
     "eslint-plugin-react": "^7.37.1",
+    "normalize.css": "^8.0.1",
```
Ivanmtta, line 15: Not sure if this is necessary. Cold we just add this to the main css file?
```
* {
  padding: 0;
  margin: 0;
  box-sizing: border-box;
}
```

akiskips: Sure we can do that . I removed the package and added the reset rule in the start of index.css
My idea was to use the normalize.css that seems to add a variaty of reset rules for different browsers. But it adds to complexity and probably we dont need it atm:

https://github.com/necolas/normalize.css/blob/master/normalize.css

flanakin: Ivan and I discussed this. We agreed to revert back to normalize.css out of the box.

@akiskips You can resolve this comment when you've done that. Thanks!

akiskips: Reverted to the normalize.css package.

flanakin (approved):

PR #1090, opened 2024-11-04
General best practices documentation update

nteyan:
## 🛠️ Description

This documentation update introduces new queries to summarize the counts of various resources. The aim is to provide users with an overview of their resource inventory, helping them understand the distribution and quantity of resources in their environment.

> - [ ] 🤏 Lint tests
> - [ ] 💪 Unit tests
> - [ ] 🙌 Integration tests

> - [X] 🤏 The change is less than 20 lines of code.

> - [X] ❎ Log not needed (small/internal change)

> - [X] ✅ Public docs in `docs` (required for `dev`)

### DIFF
--- a/docs/_docs/best-practices/general/general.md
+++ b/docs/_docs/best-practices/general/general.md
@@ -20,6 +20,7 @@ Discover essential FinOps best practices to optimize cost efficiency and governa
 <details open markdown="1">
    <summary class="fs-2 text-uppercase">On this page</summary>
 
+- [Resource inventory summary](#resource-inventory-summary)
 - [Carbon Optimization](#carbon-optimization)
 - [🙋‍♀️ Looking for more?](#️-looking-for-more)
 - [🧰 Related tools](#-related-tools)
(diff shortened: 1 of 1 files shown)


docs/_docs/best-practices/general/general.md
```diff
@@ -28,6 +29,157 @@ Discover essential FinOps best practices to optimize cost efficiency and governa
 
 ---
 
+## Resource inventory summary
```
flanakin, line 32: This is not a service. Maybe use "Resource Manager" instead?

flanakin, line 32: Move this after ACO so it's sorted alphabetically.

docs/_docs/best-practices/general/general.md
```diff
@@ -28,6 +29,157 @@ Discover essential FinOps best practices to optimize cost efficiency and governa
...
+
+### Query: Count of all resources
```
flanakin, line 34: Is this really necessary? It feels too simple to document. What scenario does this actually address? Fwiw, I think we have a larger effort to document scenarios across the entire BPL, but it's fine if we don't do that right now.

docs/_docs/best-practices/general/general.md
```diff
@@ -28,6 +29,157 @@ Discover essential FinOps best practices to optimize cost efficiency and governa
...
+<h4>Query</h4>
+
+```kql
+Resources
+| summarize count()
+```
+
+<br>
+
+### Query: Count of all resources per subscription
```
flanakin, line 51: This still feels too simplistic, right? If we think this and the previous one are needed, should we merge both with the next one, which is closer to what I think people are looking for with a resource inventory?

docs/_docs/best-practices/general/general.md
```diff
@@ -28,6 +29,157 @@ Discover essential FinOps best practices to optimize cost efficiency and governa
...
+<h4>Category</h4>
+
+Inventory
+
+<h4>Query</h4>
+
+```kql
+resources
+| summarize Count=count(id) by subscriptionId
+| order by Count desc
```
flanakin, line 64: Query styling
```suggestion
Resources
| summarize Count = count(id) by subscriptionId
| order by Count desc
```

flanakin, line 64: nit: I'm on the fence as to whether we should put sorting in the queries. We don't want unnecessary load on the server. I get it may be useful in some scenarios, but I'd probably leave that to the user to decide on their own unless sorting is critical for the scenario.

docs/_docs/best-practices/general/general.md
```diff
@@ -28,6 +29,157 @@ Discover essential FinOps best practices to optimize cost efficiency and governa
...
+<h4>Category</h4>
+
+Inventory
+
+<h4>Query</h4>
+
+```kql
+Resources 
+| extend type = case(
+type contains 'microsoft.netapp/netappaccounts', 'NetApp Accounts',
```
flanakin, line 82: This is unmaintainable and not technically accurate, if I'm reading it correctly. We should discuss this one.

docs/_docs/best-practices/general/general.md
```diff
@@ -28,6 +29,157 @@ Discover essential FinOps best practices to optimize cost efficiency and governa
...
+
+<h4>Category</h4>
+
+Inventory
+
+<h4>Query</h4>
+
+```kql
+resources
+| summarize count() by location
```
flanakin, line 178: Query styling...
```suggestion
Resources
| summarize Count = count() by location
```

flanakin (changes requested):

PR #1130, opened 2024-11-16
Adding Advisor and Resource Graph recommendations exports to Hubs

helderpinto:
## 🛠️ Description

This is the first PR in a series of PRs that will add Advisor and Resource Graph recommendations exports to Hubs. This PR includes the initial Advisor/ARG recommendations export and the schema for recommendations. It is compatible with Data Explorer ingestion. For this PR, we also fixed Western Europe timezones.

The upcoming PRs will address documentation, reporting, and granting permissions to the ADF identity at a higher scope, for the recommendations exports to succeed. Currently, Reader permissions have to be granted manually at the subscription or management group level.

> - [ ] 🤏 Lint tests
> - [ ] 💪 Unit tests
> - [ ] 🙌 Integration tests

> - [X] ➡️ Will add log in a future PR (feature branch PRs only)

> - [X] ➡️ Will add docs in a future PR (feature branch PRs only)


src/templates/finops-hub/modules/scripts/Recommendations.json
```diff
@@ -0,0 +1,74 @@
...
+        },
+        {
+            "version": "1.0",
+            "source": "ResourceGraph",
+            "provider": "Microsoft",
+            "type": "VMsWithoutAHBRecommendations",
```
flanakin, line 63: Customers will have more VMs without AHB than they will with AHB. This will be mostly noise. How are we conveying that this not a high confidence recommendation? Fwiw, I'm not sure how we define the impact and whether these recommendations meet that bar. But I have similar concerns to other properties.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -841,6 +886,37 @@ resource trigger_DailySchedule 'Microsoft.DataFactory/factories/triggers@2018-06
   }
 }
 
+resource trigger_RecommendationsDailySchedule 'Microsoft.DataFactory/factories/triggers@2018-06-01' = {
```
flanakin, line 889: Any reason we need to create separate schedule? We already have one for exports. Can we just use one schedule so everything runs at the same time?

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -4863,6 +4939,401 @@ resource pipeline_RerunETL 'Microsoft.DataFactory/factories/pipelines@2018-06-01
   }
 }
 
+//------------------------------------------------------------------------------
+// recommendations export pipeline
+// Triggered by dailyRecommendations trigger
+//------------------------------------------------------------------------------
+@description('Extracts Azure Advisor and custom recommendations from the Resource Graph API.')
+resource pipeline_ExecuteRecommendations 'Microsoft.DataFactory/factories/pipelines@2018-06-01' = {
+  name: '${recommendationsDataSet}_Execute'
```
flanakin, line 4948: Our pattern is to name these based on the source/destination. From our perspective, the source is the query files, not exactly ARG, so I'm thinking that might be the best source 🤔
```suggestion
  name: 'queries_ETL_ingestion'
```

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -4863,6 +4939,401 @@ resource pipeline_RerunETL 'Microsoft.DataFactory/factories/pipelines@2018-06-01
...
+        ]  
+        policy: {
+          secureOutput: false
+          secureInput: false
+        }
+        userProperties: []
+        typeProperties: {
+          variableName: 'blobBasePath'
+          value: {
+            value: '@concat(\'${recommendationsDataSet}/\', variables(\'blobExportTimestamp\'), \'${recommendationsScope}/\', variables(\'instanceId\'), \'${ingestionIdFileNameSeparator}\')'
```
flanakin, line 5007: Thinking ahead, we'll have non-recommendation queries as well. We should pull this from the config and not hard-code it.

src/templates/finops-hub/modules/scripts/Recommendations.json
```diff
@@ -0,0 +1,74 @@
+{
+    "queries": [
```
flanakin, line 2: Thinking ahead for when we support pulling data from Log Analytics and other systems, should we add a root-level "source" or something to indicate that we should run this against ARG? We also need to indicate what table this will go into, to save room for non-recommendation queries.
```suggestion
    "source": "AzureResourceGraph",
    "dataType": "Recommendations",
    "queries": [
```

src/templates/finops-hub/modules/scripts/Recommendations.json
flanakin, line 1: Since this only has ARG recommendations, should we call this `Recommendations.ResourceGraph.json`? That will keep all recommendation files together.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -4863,6 +4939,401 @@ resource pipeline_RerunETL 'Microsoft.DataFactory/factories/pipelines@2018-06-01
...
+        type: 'SetVariable'
+        dependsOn: []
+        policy: {
+          secureOutput: false
+          secureInput: false
+        }
+        userProperties: []
+        typeProperties: {
+          variableName: 'schemaFile'
+          value: 'recommendations_1.0.json'
```
flanakin, line 5023: This should be dynamic based on the target data type.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -4863,6 +4939,401 @@ resource pipeline_RerunETL 'Microsoft.DataFactory/factories/pipelines@2018-06-01
...
+        dependsOn: [
+          {
+            activity: 'Load Schema Mappings'
+            dependencyConditions: ['Failed']
+          }
+        ]
+        userProperties: []
+        typeProperties: {
+          message: {
+            value: '@concat(\'Unable to load the \', variables(\'schemaFile\'), \' recommendations schema file. Please confirm the schema and version are supported for FinOps hubs ingestion. Unsupported files will remain in the ingestion container.\')'
```
flanakin, line 5097:
```suggestion
            value: '@concat(\'Unable to load the \', variables(\'schemaFile\'), \' schema file. Please confirm the schema and version are supported for FinOps hubs ingestion. Unsupported files will remain in the ingestion container.\')'
```

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -4863,6 +4939,401 @@ resource pipeline_RerunETL 'Microsoft.DataFactory/factories/pipelines@2018-06-01
...
+            }
+            formatSettings: {
+              type: 'JsonReadSettings'
+            }
+          }
+          dataset: {
+            referenceName: dataset_config.name
+            type: 'DatasetReference'
+            parameters: {
+              fileName: 'Recommendations.json'
```
flanakin, line 5132: Make this dynamic to pull all JSON files in the queries folder.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -4863,6 +4939,401 @@ resource pipeline_RerunETL 'Microsoft.DataFactory/factories/pipelines@2018-06-01
...
+              inputs: [
+                {
+                  referenceName: 'resourcegraph'
+                  type: 'DatasetReference'
+                  parameters: {}
+                }
+              ]
+              outputs: [
+                {
+                  referenceName: 'ingestion'
```
flanakin, line 5227:
```suggestion
                  referenceName: dataset_ingestion.name
```

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -4863,6 +4939,401 @@ resource pipeline_RerunETL 'Microsoft.DataFactory/factories/pipelines@2018-06-01
...
+                }
+                enableStaging: false
+                translator: {
+                  value: '@activity(\'Load Schema Mappings\').output.firstRow.translator'
+                  type: 'Expression'
+                }
+              }
+              inputs: [
+                {
+                  referenceName: 'resourcegraph'
```
flanakin, line 5220: This should be dynamic based on the data source to support other targets. It's fine if we don't support them now, but we should at least be prepared so we don't have a bunch of cleanup work later.

Also, use the `dataset_resourcegraph.name` variable reference instead of hard-coding.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -4863,6 +4939,401 @@ resource pipeline_RerunETL 'Microsoft.DataFactory/factories/pipelines@2018-06-01
...
+              name: 'Fail Pipeline'
+              type: 'Fail'
+              dependsOn: []
+              userProperties: []
+              typeProperties: {
+                message: {
+                  value: 'Pipeline failed'
+                  type: 'Expression'
+                }
+                errorCode: 'ARGQueriesFailed'
```
flanakin, line 5306: We should make the prefix dynamic and pull from the JSON file.

src/templates/finops-hub/modules/dataFactory.bicep
flanakin, line 1: This PR is missing docs, which are required to merge to dev. Can you change the target branch to `features/recs`? I just created it for this. Also create an issue to track all the work that's needed before we merge to dev. That way you add things to the issue instead of fixing everything in one massive PR. That'll make it so we can make small PRs which are easier to review and ensure we don't miss anything as we go by confirming the task list in the issue.

flanakin (changes requested):

src/templates/finops-hub/modules/azuretimezones.bicep
```diff
@@ -12,26 +12,26 @@ param timezoneobject object = {
   eastasia: 'China Standard Time'
   eastus: 'Eastern Standard Time'
   eastus2: 'Eastern Standard Time'
-  francecentral: 'Central European Time'
-  germanynorth: 'Central European Time'
-  germanywestcentral: 'Central European Time'
+  francecentral: 'W. Europe Standard Time'
+  germanynorth: 'W. Europe Standard Time'
+  germanywestcentral: 'W. Europe Standard Time'
```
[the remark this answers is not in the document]
helderpinto: The previous time zone codes were incorrect and were leading to incorrect ADF deployments. This was also addressed in [#1182](https://github.com/microsoft/finops-toolkit/pull/1182).

src/templates/finops-hub/modules/scripts/IngestionSetup.kql
```diff
@@ -1175,7 +1175,23 @@ CommitmentDiscountUsage_transform_v1_0()
     x_SourceName:                   string,    // Hubs add-on
     x_SourceProvider:               string,    // Hubs add-on
     x_SourceType:                   string,    // Hubs add-on
-    x_SourceVersion:                string     // Hubs add-on
+    x_SourceVersion:                string,    // Hubs add-on
+    ResourceId:                     string,    // Advisor
```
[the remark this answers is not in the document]
helderpinto: OK. I'll also do a similar reordering for the `Recommendations_final_v1_0` table schema.

src/templates/finops-hub/modules/scripts/IngestionSetup.kql
```diff
@@ -1175,7 +1175,23 @@ CommitmentDiscountUsage_transform_v1_0()
...
+    ResourceName:                   string,    // Advisor
```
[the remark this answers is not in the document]
helderpinto: You are right. This is inspired by Advisor's schema, but the properties naming is specific to Hubs.

src/templates/finops-hub/modules/scripts/IngestionSetup.kql
```diff
@@ -1226,6 +1242,20 @@ Recommendations_transform_v1_0()
         'SkuSize', coalesce(SKU, SkuName),
         'SkuTerm', isoMonths(Term)
       ),
+      x_SourceType endswith 'Recommendations', bag_pack(
```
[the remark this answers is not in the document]
helderpinto: Changed to `endswith '-HubsRecommendations'` to be more specific to recommendations generated by Hubs.

src/templates/finops-hub/modules/scripts/IngestionSetup.kql
```diff
@@ -1226,6 +1242,20 @@ Recommendations_transform_v1_0()
...
+        'ResourceId', ResourceId,
+        'ResourceName', ResourceName,
+        'ResourceGroupName', x_ResourceGroupName,
+        'ResourceType', x_ResourceType,
```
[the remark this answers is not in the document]
helderpinto: Why not? I just promoted them to columns and adjusted the transform query accordingly.

src/templates/finops-hub/modules/dataFactory.bicep
helderpinto: Hubs recommendations queries are declared in a single `Recommendations.json` file. This ADF pipeline is dedicated to process Hubs recommendations only. The queries folder might have other JSON files for other purposes, for example, for declaring KQL queries that are going to feed the governance report with resource inventories. Other queries types will have their own ADF pipeline.

src/templates/finops-hub/modules/dataFactory.bicep
helderpinto: Changed to refer to  `dataset_resourcegraph.name` instead.

Regarding dynamic data sets based on the data source, I agree this will be needed sooner or later. However, this ADF activity was specifically designed for ARG queries and making it generic at this moment would make it much more complex. I prefer to keep it simple at this moment and be prepared to make it more generic when the need arises.

src/templates/finops-hub/modules/dataFactory.bicep
helderpinto: Can you be more specific? This activity is very generic, being there to raise a pipeline error only when at least one of the queries raised an error that is not attributable to a result with empty rows. This is needed to ensure the pipeline finishes successfully in case no cost recommendations were found.

src/templates/finops-hub/modules/dataFactory.bicep
helderpinto: I just changed the target branch to `features/recs`. I will create an issue to track all the work that is still needed.

helderpinto: Issue created: #1224

src/templates/finops-hub/modules/dataFactory.bicep
flanakin: My point is we should keep this more open and not force everything to have its own pipeline, which requires more code and more maintenance and likely more bugs. We don't need to build everything now, but at a minimum we need to support non-recommendation ARG queries. Nicolas is working on that and waiting for this, so let's just make it open upfront.

helderpinto: I agree with your point, but making it open is more than looking for all JSON query files. We will have multiple output schemas (recommendations, inventory, etc.) and we need to make the bridge between the query file and the respective output schema. The query file will have to refer to the schema name to be used. This will require me to refactor the pipeline to make it generic and need more time to deliver a complete solution.

helderpinto: @flanakin, I just committed the changes to make the pipeline query type agnostic. Due to the ADF limitations regarding activity nesting, the solution became more complex than expected and forced me to explode all ARG queries into multiple JSON files. Let me know if it can be improved further before merging.

src/templates/finops-hub/modules/dataFactory.bicep
helderpinto: With the pipeline refactoring, I ended up implementing what you suggested. It hurt but the result looks good :-)

src/templates/finops-hub/modules/dataFactory.bicep
helderpinto: I realized now there are 29 more conversations to address :-D Hold on! :-D

src/templates/finops-hub/modules/scripts/IngestionSetup.kql
```diff
@@ -1226,6 +1242,20 @@ Recommendations_transform_v1_0()
...
+        'RecommendationCategory', x_RecommendationCategory,
+        'RecommendationId', x_RecommendationId,
+        'RecommendationSolution', x_RecommendationSolution,
+        'RecommendationTypeId', x_RecommendationTypeId,
+        'RecommendationImpact', x_RecommendationImpact,
+        'RecommendationControl', x_RecommendationControl,
+        'RecommendationMaturityLevel', x_RecommendationMaturityLevel,
```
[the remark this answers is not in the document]
helderpinto: Yes, when we started implementing this, I was also suggesting these columns to become standard, but this was before seeing the reservations recommendations coming from ACM. I am also unsure about how to populate them for non-Advisor/custom recommendations.

src/templates/finops-hub/modules/scripts/IngestionSetup.kql
```diff
@@ -1248,7 +1278,10 @@ Recommendations_transform_v1_0()
       x_SourceName,
       x_SourceProvider,
       x_SourceType,
-      x_SourceVersion
+      x_SourceVersion,
+      SubAccountName,
+      x_RecommendationDescription,
+      x_RecommendationDate
```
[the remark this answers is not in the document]
helderpinto: Sorted

src/templates/finops-hub/modules/scripts/IngestionSetup.kql
```diff
@@ -1267,7 +1300,10 @@ Recommendations_transform_v1_0()
     x_SourceName:                string,
     x_SourceProvider:            string,
     x_SourceType:                string,
-    x_SourceVersion:             string
+    x_SourceVersion:             string,
+    SubAccountName:              string,
+    x_RecommendationDescription: string,
+    x_RecommendationDate:        datetime
```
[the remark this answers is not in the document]
helderpinto: Sorted

src/templates/finops-hub/modules/scripts/IngestionSetup.kql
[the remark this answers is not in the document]
helderpinto: I understand and agree with your point, but I am not sure about how to update the version. We have to discuss this.

src/templates/finops-hub/schemas/recommendations_1.0.json
```diff
@@ -0,0 +1,145 @@
+{
+    "additionalColumns": [],
+    "translator": {
+        "type": "TabularTranslator",
+        "mappings": [
+            {
+                "source": {
+                    "path": "[['x_RecommendationId']"
```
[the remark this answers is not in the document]
helderpinto: Sorted

src/templates/finops-hub/modules/storage.bicep
```diff
@@ -404,6 +408,10 @@ resource uploadSettings 'Microsoft.Resources/deploymentScripts@2023-08-01' = {
         name: 'schemaFiles'
         value: string(schemaFiles)
       }
+      {
+        name: 'queriesFiles'
+        value: string(queriesFiles)
```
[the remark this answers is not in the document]
helderpinto: Fixed

src/templates/finops-hub/modules/storage.bicep
```diff
@@ -77,6 +77,10 @@ var schemaFiles = {
   'reservationrecommendations_2023-05-01_mca': loadTextContent('../schemas/reservationrecommendations_2023-05-01_mca.json')
   'reservationtransactions_2023-05-01_ea': loadTextContent('../schemas/reservationtransactions_2023-05-01_ea.json')
   'reservationtransactions_2023-05-01_mca': loadTextContent('../schemas/reservationtransactions_2023-05-01_mca.json')
+  'recommendations_1.0': loadTextContent('../schemas/recommendations_1.0.json')
+}
+var queriesFiles = {
```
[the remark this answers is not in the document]
helderpinto: Fixed

src/templates/finops-hub/modules/storage.bicep
```diff
@@ -77,6 +77,10 @@ var schemaFiles = {
   'reservationrecommendations_2023-05-01_mca': loadTextContent('../schemas/reservationrecommendations_2023-05-01_mca.json')
   'reservationtransactions_2023-05-01_ea': loadTextContent('../schemas/reservationtransactions_2023-05-01_ea.json')
   'reservationtransactions_2023-05-01_mca': loadTextContent('../schemas/reservationtransactions_2023-05-01_mca.json')
+  'recommendations_1.0': loadTextContent('../schemas/recommendations_1.0.json')
```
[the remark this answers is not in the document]
helderpinto: Fixed

src/templates/finops-hub/modules/scripts/Recommendations.json
```diff
@@ -0,0 +1,74 @@
...
...    "query": "advisorresources | where type == 'microsoft.advisor/recommendations' | where properties.category == 'Cost' | project x_RecommendationId=id, x_ResourceGroupName=tolower(resourceGroup), SubAccountId=subscriptionId, x_RecommendationCategory=tostring(properties.category), x_RecommendationImpact=tostring(properties.impact), x_RecommendationTypeId= tostring(properties.recommendationTypeId), x_RecommendationControl=tostring(properties.extendedProperties.recommendationControl), x_RecommendationMaturityLevel=tostring(properties.extendedProperties.maturityLevel), x_RecommendationDescription=tostring(properties.shortDescription.problem), x_RecommendationSolution=tostring(properties.shortDescription.solution), ResourceId=tolower(properties.resourceMetadata.resourceId), x_ResourceType=tolower(properties.impactedField), ResourceName=tolower(properties.impactedValue), x_RecommendationDetails=tostring(properties.extendedProperties), x_RecommendationDate=tostring(properties.lastUpdated) | join kind=leftouter ( resourcecontainers | where type == 'microsoft.resources/subscriptions' | project SubAccountName=name, SubAccountId=subscriptionId ) on SubAccountId | project-away SubAccountId1"
```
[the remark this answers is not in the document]
helderpinto: We've already discussed this before. The JSON format does not play well with multiple line strings. If you know some trick for dealing with this limitation, let me know.

src/templates/finops-hub/modules/scripts/Recommendations.json
```diff
@@ -0,0 +1,74 @@
+{
+    "queries": [
+        {
+            "version": "1.0",
+            "source": "Advisor",
+            "provider": "Microsoft",
+            "type": "AdvisorCostRecommendations",
```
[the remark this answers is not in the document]
helderpinto: Properties sorted. Advisor does not have a type for recommendations. Only a recommendation type ID (GUID).

src/templates/finops-hub/modules/scripts/Recommendations.json
```diff
@@ -0,0 +1,74 @@
...
+        },
+        {
+            "version": "1.0",
+            "source": "ResourceGraph",
+            "provider": "Microsoft",
+            "type": "UnattachedDisksRecommendations",
```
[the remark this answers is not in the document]
helderpinto: The recommendation type format was changed and maybe your suggestion does not apply anymore.

src/templates/finops-hub/modules/scripts/Recommendations.json
```diff
@@ -0,0 +1,74 @@
...
+        },
+        {
+            "version": "1.0",
+            "source": "ResourceGraph",
+            "provider": "Microsoft",
+            "type": "NonSpotAKSClustersRecommmendations",
```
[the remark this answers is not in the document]
helderpinto: The recommendation type format was changed and maybe your suggestion does not apply anymore.

src/templates/finops-hub/modules/scripts/Recommendations.json
```diff
@@ -0,0 +1,74 @@
+{
+    "queries": [
+        {
```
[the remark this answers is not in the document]
helderpinto: I believe all these queries are already part of the best practices library.

src/templates/finops-hub/modules/scripts/Recommendations.json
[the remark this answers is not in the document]
helderpinto: I do not agree with your suggestion, because not leveraging AHB does not necessarily mean the VM is unlicensed. PAYG Windows (not AHB) is licensed Windows.

src/templates/finops-hub/modules/dataFactory.bicep
helderpinto: Solved

src/templates/finops-hub/modules/dataFactory.bicep
helderpinto: Solved

src/templates/finops-hub/modules/scripts/Recommendations.json
helderpinto: Solved

src/templates/finops-hub/modules/dataFactory.bicep
helderpinto: I meanwhile understood what you meant. However, given the nature of each source type (ARG, ADX, etc.), error handling will probably be specific to the source type. We can discuss this further offline.

src/templates/finops-hub/modules/scripts/Recommendations.json
helderpinto: This recommendation is coming from the same query that feeds the AHB section of the Cost Optimization Workbook. Your point is valid and we should discuss this with an overarching perspective of the recommendations we generate across the toolkit.

src/templates/finops-hub/modules/dataFactory.bicep
helderpinto: I don't see a problem in using the same schedule. The only problem is customers who just deployed hubs will have to wait until the next day to see recommendations. I decided for a specific schedule so that we set the trigger start time to be 1h after the deployment time.

src/templates/finops-hub/modules/dataFactory.bicep
helderpinto: Addressed with a different solution, due to the refactoring required by the queries processing generalization.

src/templates/finops-hub/modules/dataFactory.bicep
helderpinto: Solved

src/templates/finops-hub/modules/scripts/Recommendations.json
helderpinto: This comment does not apply anymore, as we now have a query file per query with the input dataset specified in the file itself.

src/templates/finops-hub/modules/dataFactory.bicep
helderpinto: Solved

helderpinto:
@flanakin, the ADF pipeline was refactored to become query type-agnostic. It is now prepared to deal with other query types in the future. There are still some unresolved conversations. Feel free to ping me offline or let's set up a meeting to discuss this further.

src/templates/finops-hub/modules/dataFactory.bicep
helderpinto: I ended up realizing that query error handling can be generalized, if we assume that empty query results will not be considered as an error. I just committed the code with this assumption.

src/templates/finops-hub/modules/scripts/IngestionSetup.kql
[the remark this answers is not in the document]
helderpinto: I just updated the Recommendations table version to 1.0r2

src/templates/finops-hub/modules/dataFactory.bicep
helderpinto: Changed to reuse single daily schedule, setting the trigger start time to be 1h after the deployment time.

src/templates/finops-hub/modules/scripts/HubsRecommendations-AdvisorCost.json
```diff
@@ -0,0 +1,11 @@
...
+    "schemaFile": "recommendations_1.0.json",
```
flanakin, line 6: nit: Isn't this implied from the dataset and version numbers?

src/templates/finops-hub/modules/scripts/HubsRecommendations-AdvisorCost.json
```diff
@@ -0,0 +1,11 @@
...
+    "scope": "azurerbac",
```
flanakin, line 7: What is this for?

src/templates/finops-hub/modules/scripts/HubsRecommendations-AdvisorCost.json
```diff
@@ -0,0 +1,11 @@
...
+    "source": "Advisor",
```
flanakin, line 8: Should we use "Azure Advisor" to distinguish from AWS Trusted Advisor, assuming we eventually support that. (Or customers decide to add it themselves.)

src/templates/finops-hub/modules/scripts/HubsRecommendations-EmptySQLElasticPools.json
```diff
@@ -0,0 +1,11 @@
...
+    "schemaFile": "recommendations_1.0.json",
+    "scope": "azurerbac",
+    "source": "ResourceGraph",
```
flanakin, line 8: We should make these display strings. That's the norm in FOCUS.
```suggestion
    "source": "Resource Graph",
```

src/templates/finops-hub/modules/scripts/IngestionSetup.kql
```diff
@@ -1439,6 +1471,74 @@ CommitmentDiscountUsage_transform_v1_0()
 .create-or-alter function
 with (docstring='All recommendations transformed to FOCUS 1.0.', folder='Recommendations')
 Recommendations_transform_v1_0()
+{
```
flanakin, line 1474: This looks like you're changing v1_0 and not v1_0r2. Can you confirm that your changes are in the right place? Ideally, we should leave v1_0 untouched.

src/templates/finops-hub/modules/scripts/IngestionSetup.kql
```diff
@@ -1496,19 +1596,35 @@ Recommendations_transform_v1_0()
...
     )
     //
     // Sort columns and apply final transforms
     | project
       ProviderName,
-      SubAccountId = iff(isnotempty(SubscriptionId), strcat('/subscriptions/', SubscriptionId), ''),
-      x_IngestionTime = ingestion_time(),
+      ResourceId,
+      ResourceName,
+      ResourceType      = coalesce(ResourceType, x_ResourceType),
```
flanakin, line 1617: Look up x_ResourceType via the `resource_type()` function. TODO is fine.

src/templates/finops-hub/modules/scripts/IngestionSetup.kql
```diff
@@ -1519,18 +1635,41 @@ Recommendations_transform_v1_0()
 
 // Recommendations_final_v1_0 table
 .create-merge table Recommendations_final_v1_0 (
```
flanakin, line 1637: This table shouldn't change

src/templates/finops-hub/modules/storage.bicep
```diff
@@ -72,12 +72,25 @@ var schemaFiles = {
...
   'pricesheet_2023-05-01_mca': loadTextContent('../schemas/pricesheet_2023-05-01_mca.json')
+  'recommendations_1.0': loadTextContent('../schemas/recommendations_1.0.json')
   'reservationdetails_2023-03-01': loadTextContent('../schemas/reservationdetails_2023-03-01.json')
   'reservationrecommendations_2023-05-01_ea': loadTextContent('../schemas/reservationrecommendations_2023-05-01_ea.json')
   'reservationrecommendations_2023-05-01_mca': loadTextContent('../schemas/reservationrecommendations_2023-05-01_mca.json')
   'reservationtransactions_2023-05-01_ea': loadTextContent('../schemas/reservationtransactions_2023-05-01_ea.json')
   'reservationtransactions_2023-05-01_mca': loadTextContent('../schemas/reservationtransactions_2023-05-01_mca.json')
 }
+var queryFiles = {
+  'HubsRecommendations-AdvisorCost': loadTextContent('./scripts/HubsRecommendations-AdvisorCost.json')
```
flanakin, line 83: Shouldn't these be in the queries folder?
```suggestion
  'HubsRecommendations-AdvisorCost': loadTextContent('./queries/HubsRecommendations-AdvisorCost.json')
```

src/templates/finops-hub/schemas/recommendations_1.0.json
```diff
@@ -0,0 +1,145 @@
+{
+    "additionalColumns": [],
+    "translator": {
+        "type": "TabularTranslator",
+        "mappings": [
+            {
+                "source": {
+                    "path": "[['ResourceId']"
```
flanakin, line 8: Is this path correct? It doesn't look right with the extra `[`. Same for the others.

src/templates/finops-hub/schemas/recommendations_1.0.json
```diff
@@ -0,0 +1,145 @@
...
+                "source": {
+                    "path": "[['x_RecommendationId']"
+                },
+                "sink": {
+                    "path": "x_RecommendationId"
+                }
+            },
+            {
+                "source": {
+                    "path": "[['x_RecommendationImpact']"
```
flanakin, line 88: Is the impact actually useful for customers? Can we stand by the impact assessment for any of these recommendations? I've never trusted it, but I also haven't had a conversation with customers about it.

src/templates/finops-hub/schemas/recommendations_1.0.json
```diff
@@ -0,0 +1,145 @@
...
+                },
+                "sink": {
+                    "path": "x_RecommendationImpact"
+                }
+            },
+            {
+                "source": {
+                    "path": "[['x_RecommendationMaturityLevel']"
```
flanakin, line 96: What values does this have? Who assesses maturity? Is it standardized?

flanakin (changes requested):

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -4785,7 +5503,7 @@ resource startTriggers 'Microsoft.Resources/deploymentScripts@2020-10-01' = {
       }
       {
         name: 'Pipelines'
-        value: join([ pipeline_InitializeHub.name ], '|')
+        value: join([pipeline_InitializeHub.name], '|')
```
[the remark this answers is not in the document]
helderpinto: I can't tell you. I wasn't the author.

flanakin (approved):

src/templates/finops-hub/modules/scripts/HubsRecommendations-AdvisorCost.json
```diff
@@ -0,0 +1,11 @@
+{
+    "inputDataset": "resourcegraph",
```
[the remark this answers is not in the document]
helderpinto: Actually, queries are just a bridge between input and output datasets. We are mapping the `inputDataset` value with a corresponding dataset in ADF. In order to keep the recommendations pipelines' ADF code understandable, I prefer to keep the dataset term.

src/templates/finops-hub/modules/scripts/HubsRecommendations-AdvisorCost.json
```diff
@@ -0,0 +1,11 @@
...
+    "outputDataset": "Recommendations",
```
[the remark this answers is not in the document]
helderpinto: See previous comment. If we are keeping `inputDataset` then it makes sense to let this one be `outputDataset`.

src/templates/finops-hub/modules/scripts/HubsRecommendations-AdvisorCost.json
helderpinto: I am not getting it. This schema file is used for the ADF mappings between the ARG input dataset and the blob storage output dataset. It has a version number like all the other ADF schema files (FOCUS and other ACM exports).

src/templates/finops-hub/modules/scripts/HubsRecommendations-AdvisorCost.json
helderpinto: The scope of recommendations queries is dependent on the scope of the Reader permissions that are granted to the ADF managed identity. We cannot anticipate which scope customers will choose. Therefore, we need some neutral string that will be used in the blob storage path used by recommendations exports files. I am open to suggestions for a better naming.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -4737,6 +4953,503 @@ resource pipeline_ExecuteIngestionETL 'Microsoft.DataFactory/factories/pipelines
...
+            formatSettings: {
+              type: 'JsonReadSettings'
+            }
+          }
+          dataset: {
+            referenceName: dataset_config.name
+            type: 'DatasetReference'
+            parameters: {
+              fileName: 'settings.json'
+              folderPath: '${configContainerName}/queries'
```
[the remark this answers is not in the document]
helderpinto: This folder will contain definitions meant to specify data export from one dataset (e.g., ARG) to another (e.g., Recommendations folder in the `ingestion` container). Therefore, I wouldn't call it "datasets", because datasets are inputs and outputs of the definitions. In the context of this PR, "queries" seemed to be the logical naming. Looking forward and more generically speaking, maybe something like "transformations" would be better. I am open to better naming suggestions.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -4737,6 +4953,503 @@ resource pipeline_ExecuteIngestionETL 'Microsoft.DataFactory/factories/pipelines
...
+        dependsOn: []
+        policy: {
+          secureOutput: false
+          secureInput: false
+        }
+        userProperties: []
+        typeProperties: {
+          variableName: 'blobExportTimestamp'
+          value: {
+            value: '@concat(utcNow(\'yyyy\'),\'/\',utcNow(\'MM\'),\'/\',utcNow(\'dd\'),\'/\')'
```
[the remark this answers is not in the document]
helderpinto: This implementation allows for multiple instances of the same query execution on the same day, which will result in different blobs being generated for each query run. Although the ADF trigger schedule runs once per day for each query, we are assuming more/less frequent runs are possible. It will be up to the data visualizations to handle multiple recommendations of the same type in a specific time period.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -4737,6 +4953,503 @@ resource pipeline_ExecuteIngestionETL 'Microsoft.DataFactory/factories/pipelines
...
+            {
+              type: 'Fail'
+              name: 'Unsupported input dataset'
+              userProperties: []
+              typeProperties: {
+                message: {
+                  value: '@concat(\'Unable to execute the specified query because the input data set is not supported. Dataset: \', pipeline().parameters.inputDataset)'
+                  type: 'Expression'
+                }
+                errorCode: 'UnsupportedInputDataset'
```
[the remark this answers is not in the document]
helderpinto: Done

src/templates/finops-hub/modules/scripts/HubsRecommendations-AdvisorCost.json
helderpinto: Agree. Done.

src/templates/finops-hub/modules/scripts/HubsRecommendations-EmptySQLElasticPools.json
helderpinto: Agree. Done.

src/templates/finops-hub/modules/storage.bicep
helderpinto: We don't have a `queries` folder. Can I create it at the same level of the `scripts` folder?

src/templates/finops-hub/schemas/recommendations_1.0.json
helderpinto: It has to be defined this way. I don't remember why ADF requires it, but it works :-)

src/templates/finops-hub/schemas/recommendations_1.0.json
helderpinto: Yeah, impact level is always subjective. Different customers perceive different impacts for the same recommendation type. This field is there just because of Azure Advisor. I think we should not omit data from the source that might relevant for some customers.

src/templates/finops-hub/schemas/recommendations_1.0.json
helderpinto: This is something that comes from Azure Advisor. It is used to denote recommendations that are still in Preview. I was thinking of using it also for our recommendations, to distinguish between GA and Preview ones...

arthurclares (approved):
Once this is merged, I can add the remaining recommendations.

src/templates/finops-hub/modules/scripts/HubsRecommendations-AdvisorCost.json
```diff
@@ -0,0 +1,11 @@
...
...   "query": "advisorresources | where type == 'microsoft.advisor/recommendations' | where properties.category == 'Cost' | project x_RecommendationId=id, x_ResourceGroupName=tolower(resourceGroup), SubAccountId=subscriptionId, x_RecommendationCategory=tostring(properties.category), x_RecommendationImpact=tostring(properties.impact), x_RecommendationTypeId= tostring(properties.recommendationTypeId), x_RecommendationControl=tostring(properties.extendedProperties.recommendationControl), x_RecommendationMaturityLevel=tostring(properties.extendedProperties.maturityLevel), x_RecommendationDescription=tostring(properties.shortDescription.problem), x_RecommendationSolution=tostring(properties.shortDescription.solution), ResourceId=tolower(properties.resourceMetadata.resourceId), x_ResourceType=tolower(properties.impactedField), ResourceName=tolower(properties.impactedValue), x_RecommendationDetails=tostring(properties.extendedProperties), x_RecommendationDate=tostring(properties.lastUpdated) | join kind=leftouter ( resourcecontainers | where type == 'microsoft.resources/subscriptions' | project SubAccountName=name, SubAccountId=subscriptionId ) on SubAccountId | project-away SubAccountId1",
```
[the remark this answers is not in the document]
helderpinto: Thanks for the suggestion, @arthurclares. I'll review the query and adjust it to the recommendations schema.

src/templates/finops-hub/schemas/recommendations_1.0.json
flanakin: That doesn't sound like maturity level. That feels more like a release stage. Or maybe we just have an IsPreview boolean? Given this isn't commonly used, what if we just put it in x_RecommendationDetails? That would allow us to see if it's even worth being a dedicated column. I wouldn't want a dedicated column unless we believe this belongs on every recommendation.

src/templates/finops-hub/schemas/recommendations_1.0.json
flanakin: Given we can't stand by this and I'm not sure we'd apply it to every recommendation, I would put this into x_RecommendationDetails. That will keep it available but not treat it as something required for all recommendations.

flanakin (approved):
Please add all pending feedback to the todo list and resolve the comments and you can go ahead and merge this. I have some fundamental concerns with a few things, but we can cover those later.

Going forward, let's shoot for smaller PRs so these are easier to get in 🙂

src/templates/finops-hub/modules/dataFactory.bicep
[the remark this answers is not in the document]
helderpinto: To be addressed in issue #1407

src/templates/finops-hub/modules/scripts/HubsRecommendations-AdvisorCost.json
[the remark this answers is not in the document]
helderpinto: Yes, but it is represented as a dataset in ADF.

src/templates/finops-hub/modules/scripts/HubsRecommendations-AdvisorCost.json
helderpinto: Understood. Adding a TODO to address this.

PR #1242, opened 2025-01-09
Fix deployment errors in OpenDataFunctions.kql

MSBrett:
## 🛠️ Description
This PR fixes deployment caused by

- shortening the length of the OpenDataFunctions.kql script script to less than 131K chars
- casting the type of data returned in tmp_resourcedetails to string.

> - [X] 🤏 Lint tests
> - [ ] 💪 Unit tests
> - [ ] 🙌 Integration tests

> - [X] 🤏 The change is less than 20 lines of code.

> - [X] ❎ Log not needed (small/internal change)

> - [X] ❎ Docs not needed (small/internal change)

### DIFF
--- a/src/templates/finops-hub/modules/scripts/OpenDataFunctions.kql
+++ b/src/templates/finops-hub/modules/scripts/OpenDataFunctions.kql
@@ -6,31 +6,9 @@
 with (docstring = 'Return details about the specified ID.', folder = 'OpenData')
 resource_type(id: string) {
     dynamic({
-        "arizeai.observabilityeval/organizations": { "SingularDisplayName": "Azure Native ArizeAi Cloud Service" }
-        ,"astronomer.astro/organizations": { "SingularDisplayName": "Astro Organization" }
-        ,"citrix.services/xenappessentials": { "SingularDisplayName": "Citrix Virtual Apps Essentials" }
-        ,"citrix.services/xendesktopessentials": { "SingularDisplayName": "Citrix Virtual Desktops Essentials" }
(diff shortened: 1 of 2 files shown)


src/templates/finops-hub/modules/scripts/OpenDataFunctions.kql
```diff
@@ -6,31 +6,9 @@
 with (docstring = 'Return details about the specified ID.', folder = 'OpenData')
```
RolandKrummenacher, line 6: My VS code is telling me that it is still 132138 characters. I can't deploy it: Error BCP184: File 'C:\Projects\finops-toolkit\src\templates\finops-hub\modules\scripts\OpenDataFunctions.kql' exceeded maximum size of 131072 characters.

src/templates/finops-hub/modules/scripts/OpenDataFunctions.kql
```diff
@@ -6,31 +6,9 @@
...
 resource_type(id: string) {
     dynamic({
-        "arizeai.observabilityeval/organizations": { "SingularDisplayName": "Azure Native ArizeAi Cloud Service" }
```
flanakin, line 9: How are you deciding what to keep and what to remove?

flanakin (changes requested):
I don't love removing values. While I feel there has to be a better solution, I submitted a PR to split the file into multiple files since we know it will grow over time. I don't love it, but it should get us past the current issue. I validated the deployment so it should resolve the current issues. Thanks for digging into this.

https://github.com/microsoft/finops-toolkit/pull/1269

Issue #1344, opened 2025-02-25
Issue with Governance and WorkloadOptimization KQL and Storage PowerBI Reports


DMarshall-1:
## 🐛 Problem
We have just setup FinOps V8 and downloaded the KQL and Storage example reports from [Releases](https://github.com/microsoft/finops-toolkit/releases/tag/v0.8). In both cases after unzipping and entering the Parameters the Governance and WorkloadOptimization reports fail to populate data with an error saying "The Import Storage URL matches no exports. Did you miss a module reference?"

Examples from the Governance Report below

![Image](https://github.com/user-attachments/assets/abe36ae5-b7d0-4167-ab21-d8ce8c30c7a7)

![Image](https://github.com/user-attachments/assets/e583ccdf-2034-4ddc-bcfc-8e26d573252a)

![Image](https://github.com/user-attachments/assets/c2fc479c-e105-42c2-90ab-84511a7fc535)

![Image](https://github.com/user-attachments/assets/edef9a21-0f79-4a39-a80d-4b1e5b8dad1e)

![Image](https://github.com/user-attachments/assets/924e7a82-2846-4a1c-9852-2d9570a55fef)

The other reports in the zip file, RateOptimization, CostSummary and DataIngestion all work fine. 

## 👣 Repro steps
Download the latest PBIT reports (https://github.com/microsoft/finops-toolkit/releases/download/v0.8/PowerBI-kql.zip or https://github.com/microsoft/finops-toolkit/releases/download/v0.8/PowerBI-storage.zip) 
from releases at time of writing they were updated last week. 
Unzip the files to your local PC
Open the Governance Report or WorkloadOptimization Report 
Enter your Cluster info and click OK
Error shows during initial refresh setup. 

## 🤔 Expected
Report should load without errors. 

## 📷 Screenshots
See Above

flanakin:
I can acknowledge this is a bug on us. There's an easy fix from our side. We can get this fixed in 0.9. For now, go to Transform data, right-click **ftk_DemoFilter**, select **Advanced Editor**, replace the contents with `() => ""`, then select **Done** and **Close & Apply**. This should fix the issue.

Sorry for the inconvenience!

PR #1374, opened 2025-03-08
Fix Total savings and AHB reports

### LINKED ISSUE #730: Azure Hybrid Use Benefit calculations seem off
### Discussed in https://github.com/microsoft/finops-toolkit/discussions/709

<div type='discussions-op-text'>

<sup>Originally posted by **lundejd** May 20, 2024</sup>
I finally got around to reviewing the Hybrid Benefit tab and am immediately confused by the numbers. For example, we have a Standard_D2ds_v5 server that has 2 vCPUs. On the bottom that same server is listed but the column vCPUs says 16. I have another of the same Standard_D2ds_v5 that has vCPUs of 2, which is correct, and AHB vCPUs of 8 which is correct.

This is only the beginning. Let's say I wanted to license a 16 vCPU server in Azure with AHUB, I would only need to use 16 core licenses to do it. The chart shows 64 and a usage quantity of 24. 

I would think AHUB should look like this:

vCPUs.  AHB vCPUs
1                   8
2                  8
8                   8
10                 16
16                 16
20                24

and so on. minimum 8, multiples of 8. vCPU systems that don't divide into 8 end up with some "wasted" AHUB cores.

AHB vCPUs that are enabled should equal UsageQuantity. If I have an 8 vCPU machine enabled for AHUB it should use 8 vCPUs. UsageQuantity seems to be 24 for most.

Here is a screen shot with the same size server, the first has accurate info but the following do not.

Screenshot 2024-05-20 at 8 42 30 AM</div>

flanakin:
## 🛠️ Description
- Fixed core count double-counting on the Hybrid Benefit page.
- Fixed savings to include negotiated discounts on the Total savings page.
- Fixed Storage URL reference on KQL reports

Fixes #730, #1344 

> - [ ] 🤏 Lint tests
> - [ ] 💪 Unit tests
> - [ ] 🙌 Integration tests


flanakin:
@allcontributors please add @japyjaap for bug

Issue #1400, opened 2025-03-14
[Hubs] FinOps hub apps


flanakin:
## 📝 Scenario
As an **engineer**, I need to **extend FinOps hubs safely** in order to **add new or change existing features to meet specific requirements at scale in a way that can be deployed across many hub instances**

## 💎 Solution
Build an extensibility model in FinOps hubs that allow customers and partners to extend FinOps hubs without risk of upgrades breaking them.

## 🙋‍♀️ Ask for the community

We could use your help:
1. Please vote this issue up (👍) to prioritize it.
2. Leave comments to help us solidify the vision.

PR #1494, opened 2025-04-09
Create a hub-pipeline bicep module as a placeholder for ADF pipelines

flanakin:
## 🛠️ Description
Creates a placeholder for the hub-pipeline module that will eventually be used for creating and triggering ADF pipelines.

This is a notional PR right now. I'm not convinced this is the right approach because it would lose typing. However, if we move to maintaining pipeline code in JSON files instead of bicep, this would be acceptable. Sharing it to get others' thoughts.

Related to #1400

> - [ ] 🤏 Lint tests
> - [ ] 💪 Unit tests
> - [ ] 🙌 Integration tests

### DIFF
--- a/src/templates/finops-hub/modules/dataFactory.bicep
+++ b/src/templates/finops-hub/modules/dataFactory.bicep
@@ -981,476 +981,480 @@ resource trigger_MonthlySchedule 'Microsoft.DataFactory/factories/triggers@2018-
 // config_InitializeHub pipeline
 //------------------------------------------------------------------------------
 @description('Initializes the hub instance based on the configuration settings.')
-resource pipeline_InitializeHub 'Microsoft.DataFactory/factories/pipelines@2018-06-01' = if (deployDataExplorer) {
-  name: '${safeConfigContainerName}_InitializeHub'
-  parent: dataFactory
-  properties: {
-    activities: [
-      { // Get Config
-        name: 'Get Config'
-        type: 'Lookup'
-        dependsOn: []
-        policy: {
-          timeout: '0.00:05:00'
-          retry: 2
-          retryIntervalInSeconds: 30
-          secureOutput: false
-          secureInput: false
-        }
-        userProperties: []
-        typeProperties: {
-          source: {
-            type: 'JsonSource'
-            storeSettings: {
-              type: 'AzureBlobFSReadSettings'
-              recursive: true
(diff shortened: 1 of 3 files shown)


src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -981,476 +981,480 @@ resource trigger_MonthlySchedule 'Microsoft.DataFactory/factories/triggers@2018-
...
-              type: 'AzureBlobFSReadSettings'
-              recursive: true
-              enablePartitionDiscovery: false
+module pipeline_InitializeHub 'hub-pipeline.bicep' = if (deployDataExplorer) {
+  name: 'pipeline_InitializeHub'
+  params: {
+    dataFactoryName: dataFactory.name
+    pipelineName: '${safeConfigContainerName}_InitializeHub'
+    pipelineProperties: {
+      activities: [
```
flanakin, line 990: This is mostly spacing changes. To hide spacing changes, use the gear menu above the files 👆 and select "Hide whitespace", then apply and reload.

src/templates/finops-hub/modules/dataFactory.bicep
```diff
@@ -981,476 +981,480 @@ resource trigger_MonthlySchedule 'Microsoft.DataFactory/factories/triggers@2018-
...
-            storeSettings: {
-              type: 'AzureBlobFSReadSettings'
-              recursive: true
-              enablePartitionDiscovery: false
+module pipeline_InitializeHub 'hub-pipeline.bicep' = if (deployDataExplorer) {
+  name: 'pipeline_InitializeHub'
+  params: {
+    dataFactoryName: dataFactory.name
+    pipelineName: '${safeConfigContainerName}_InitializeHub'
+    pipelineProperties: {
```
flanakin, line 989: My main hesitation with this approach is that it removes typing for the pipeline properties. If any app owner wants to hand-craft pipelines, then this would be a painful loss.

The alternative approach is to require people create pipelines using normal bicep and then we have a bicep module that only subscribes pipelines to events. I'm curious what others think 🤔

RolandKrummenacher: We use both, strongly typed and JSON-based pipelines. I could live without the typing. But I also like the idea to bring my own pipeline and have it wired to a trigger / event.

PR #1743, opened 2025-07-03
Fix circular reference in vnet module

### LINKED ISSUE #1735: Deployments fail when private networking is enabled.
## 🐛 Problem

When trying to deploy this error is encountered:

<code id='' style='white-space:pre-wrap'><div>The response for resource had empty or invalid content.</div></code></br> (Code: ResourceDeploymentFailure, Target: /subscriptions/.../resourceGroups/ftk011test01/providers/Microsoft.Storage/storageAccounts/finscriptz72k36vl2wc5k)

## 👣 Repro steps
TODO: Add repro steps below:
Deploy using web link

## 🤔 Expected
works

## 📷 Screenshots
<code id='' style='white-space:pre-wrap'><div>The response for resource had empty or invalid content.</div></code></br> (Code: ResourceDeploymentFailure, Target: /subscriptions/cab7feeb-759d-478c-ade6-9326de0651ff/resourceGroups/ftk011test01/providers/Microsoft.Storage/storageAccounts/finscriptz72k36vl2wc5k)

## ℹ️ Additional context

## 🙋‍♀️ Ask for the community

We could use your help:

MSBrett:
## 🛠️ Description
Fix circular reference in vnet module which was causing private network enabled deployments to fail.

Fixes #1735

> - [x] 🤏 Lint tests
> - [ ] 💪 Unit tests
> - [ ] 🙌 Integration tests

### DIFF
--- a/src/templates/finops-hub/modules/infrastructure.bicep
+++ b/src/templates/finops-hub/modules/infrastructure.bicep
@@ -16,8 +16,6 @@ param hub HubProperties
 // Variables
 //==============================================================================
 
-var safeHubName = replace(replace(toLower(hub.name), '-', ''), '_', '')
-// cSpell:ignore vnet
 var nsgName = '${hub.routing.networkName}-nsg'
 
 // Workaround https://github.com/Azure/bicep/issues/1853
@@ -171,7 +169,7 @@ resource nsg 'Microsoft.Network/networkSecurityGroups@2023-11-01' = if (hub.opti
 }
 
 resource vNet 'Microsoft.Network/virtualNetworks@2023-11-01' = if (hub.options.privateRouting) {
-  name: vNetName
+  name: hub.routing.networkName
   location: hub.location
   tags: getHubTags(hub, 'Microsoft.Storage/virtualNetworks')
   properties: {
@@ -201,6 +199,9 @@ resource vNet 'Microsoft.Network/virtualNetworks@2023-11-01' = if (hub.options.p
 // Required for the Azure portal and Storage Explorer
 resource blobPrivateDnsZone 'Microsoft.Network/privateDnsZones@2024-06-01' = if (hub.options.privateRouting) {
   name: string(hub.routing.dnsZones.blob.name)
+  dependsOn: [
+    vNet
+  ]
   location: 'global'
   tags: getHubTags(hub, 'Microsoft.Storage/privateDnsZones')
   properties: {}
@@ -212,7 +213,7 @@ resource blobPrivateDnsZone 'Microsoft.Network/privateDnsZones@2024-06-01' = if 
     properties: {
       registrationEnabled: false
       virtualNetwork: {
-        id: vNet.id
+        id: hub.routing.networkId
       }
     }
   }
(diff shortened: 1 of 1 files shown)


src/templates/finops-hub/modules/infrastructure.bicep
```diff
@@ -213,7 +213,7 @@ resource blobPrivateDnsZone 'Microsoft.Network/privateDnsZones@2024-06-01' = if
     properties: {
       registrationEnabled: false
       virtualNetwork: {
-        id: hub.routing.networkId
+        id: vNet.id
```
flanakin, line 216: FYI - I was on the fence whether to use an on-demand reference or the ID that we have to maintain anyway. I ran into issues using the reference in some cases so I tended to use the original ID to work around bicep dependency limitations. I'm not overly opinionated, however. Just wanted to share the reasoning behind it being an ID rather than a reference.

src/templates/finops-hub/modules/infrastructure.bicep
```diff
@@ -303,7 +303,7 @@ resource scriptStorageAccount 'Microsoft.Storage/storageAccounts@2022-09-01' = i
       defaultAction: 'Deny'
       virtualNetworkRules: [
         {
-          id: hub.routing.subnets.scripts
+          id: vNet::scriptSubnet.id
```
flanakin, line 306: I wouldn't recommend using this approach. While an on-demand reference sounds nice, nobody else can use that unless we document the names and treat them as an API that never changes. If we use the variable in `hub.routing.subnets`, then we can tell everyone to use that forever without worrying about the name, even if we change it. The reason for us to use that same variable is because if we don't, then there might be a latent bug if the variable gets out of sync. When we rely on the same variable customers will use for extensibility, then we're much less likely to ship a bug.
```suggestion
          id: hub.routing.subnets.scripts
```

src/templates/finops-hub/modules/infrastructure.bicep
```diff
@@ -19,7 +19,7 @@ param hub HubProperties
 var safeHubName = replace(replace(toLower(hub.name), '-', ''), '_', '')
 // cSpell:ignore vnet
 var vNetName = '${safeHubName}-vnet-${hub.location}'
-var nsgName = '${hub.routing.networkName}-nsg'
+var nsgName = '${vNetName}-nsg'
```
flanakin, line 22: I would not recommend this change. The hub.routing.networkName property is available to all apps. We should use that as the source of truth. If that's the same as vNetName here, then remove vNetName.
```suggestion
var nsgName = '${hub.routing.networkName}-nsg'
```

flanakin (changes requested):

src/templates/finops-hub/modules/infrastructure.bicep
```diff
@@ -303,7 +303,7 @@ resource scriptStorageAccount 'Microsoft.Storage/storageAccounts@2022-09-01' = i
       defaultAction: 'Deny'
```
flanakin, line 303: Add a dependency for `vNet::scriptSubnet` rather than using an explicit reference

flanakin (approved):

PR #1997, opened 2026-02-12
fix: Separate Event Grid subscription from status polling in Init-DataFactory

### LINKED ISSUE #1996: Init-DataFactory.ps1 fails when Event Grid subscription is already provisioning
## 🐛 Problem

When deploying a FinOps hub via `Deploy-FinOpsHub`, the deployment script `Init-DataFactory.ps1` fails if the Event Grid subscription is already being provisioned (e.g., from a previous deployment attempt or a redeployment).

In `Set-BlobTriggerSubscription`, the `Add-AzDataFactoryV2TriggerSubscription` call and the `Get-AzDataFactoryV2TriggerSubscriptionStatus` check are inside the same `Invoke-WithRetry` block. This causes two problems:

1. The status is checked immediately after subscribing, so it always finds `Provisioning` and throws.
2. On retry, `Add-AzDataFactoryV2TriggerSubscription` is called again while already provisioning, which fails with `Resource cannot be updated during provisioning` (BadRequest).

All 20 retries are exhausted and the deployment fails.

## 👣 Repro steps

1. Run `Deploy-FinOpsHub -Name MyFinOpsHub -ResourceGroupName my-resource-group -Location eastus`
2. The deployment script calls `Init-DataFactory.ps1 -StartTriggers`
3. `Set-BlobTriggerSubscription` calls `Add-AzDataFactoryV2TriggerSubscription`, then immediately checks status
4. Status is `Provisioning`, retry re-calls Add which gets `Resource cannot be updated during provisioning`
5. Deployment fails after exhausting all retries

## 🤔 Expected

The subscription operation and the status polling should be separate steps so the script waits for provisioning to complete instead of re-calling the subscribe operation.

## ℹ️ Additional context

Error trace from the deployment failure:

```
System.Management.Automation.RuntimeException: Subscription status is Provisioning, expected Enabled
...
Microsoft.Rest.Azure.CloudException: HTTP Status Code: BadRequest Error Code: BadRequest Error Message: Resource cannot be updated during provisioning
```

`src/templates/finops-hub/modules/fx/scripts/Init-DataFactory.ps1`, function `Set-BlobTriggerSubscription`.

## 🙋‍♀️ Ask for the community

We could use your help:

slapointe:
## 🛠️ Description

Fixes #1996

In `Set-BlobTriggerSubscription`, the `Add-AzDataFactoryV2TriggerSubscription` call and the `Get-AzDataFactoryV2TriggerSubscriptionStatus` check were inside the same `Invoke-WithRetry` block. This caused the status to be checked immediately after subscribing (always `Provisioning`), and on retry the subscribe call would fail with `Resource cannot be updated during provisioning`, exhausting all retries.

Split into two steps:
1. The subscription operation wrapped in a `try/catch` that tolerates an already-transitioning subscription
2. A separate status-polling retry loop that waits for provisioning to complete

> - [ ] 🤏 Lint tests
> - [ ] 💪 Unit tests
> - [ ] 🙌 Integration tests

### DIFF
--- a/src/templates/finops-hub/modules/fx/scripts/Init-DataFactory.ps1
+++ b/src/templates/finops-hub/modules/fx/scripts/Init-DataFactory.ps1
@@ -78,22 +78,39 @@ function Set-BlobTriggerSubscription([string]$TriggerName, [switch]$Subscribe)
     $action = if ($Subscribe) { 'Subscribing' } else { 'Unsubscribing' }
 
     Write-Log "$action $TriggerName to events..."
-    Invoke-WithRetry -Name "$action $TriggerName" -Delay 5 -Action {
-        if ($Subscribe)
-        {
-            Add-AzDataFactoryV2TriggerSubscription `
-                -ResourceGroupName $DataFactoryResourceGroup `
-                -DataFactoryName $DataFactoryName `
-                -Name $TriggerName | Out-Null
+    try
+    {
+        Invoke-WithRetry -Name "$action $TriggerName" -Delay 5 -Action {
+            if ($Subscribe)
+            {
+                Add-AzDataFactoryV2TriggerSubscription `
+                    -ResourceGroupName $DataFactoryResourceGroup `
+                    -DataFactoryName $DataFactoryName `
+                    -Name $TriggerName | Out-Null
+            }
+            else
+            {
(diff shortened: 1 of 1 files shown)


RolandKrummenacher:
Created a new PR #2001 .

PR #2001, opened 2026-02-13
Fix/init datafactory provisioning retry

### LINKED ISSUE #1996: Init-DataFactory.ps1 fails when Event Grid subscription is already provisioning

RolandKrummenacher:
## 🛠️ Description

Improves the fix from #1997 for Event Grid subscription handling in `Init-DataFactory.ps1`.

The original PR (#1997) correctly identified the problem (subscribe + status poll in the same retry loop), but wrapped the subscribe call in `Invoke-WithRetry` with an outer `try/catch` — meaning up to ~17 minutes of wasted retries before the "already provisioning" fallback kicks in.

This PR takes a check-first approach instead:

1. **Pre-check** the current subscription status before calling subscribe/unsubscribe
2. **Early return** if already at the target status (`Enabled`/`Disabled`)
3. **Skip the call** if already in a transitional state (`Provisioning`/`Deprovisioning`) and go straight to polling
4. **Poll separately** for completion using `Invoke-WithRetry`

Also adds the missing changelog entry under v14.

Fixes #1996
Supersedes #1997

> - [x] 🤏 Lint tests
> - [ ] 💪 Unit tests
> - [ ] 🙌 Integration tests

### DIFF
--- a/docs-mslearn/toolkit/changelog.md
+++ b/docs-mslearn/toolkit/changelog.md
@@ -33,6 +33,15 @@ The following section lists features and enhancements that are currently in deve
 
 <br><a name="latest"></a>
 
+## v14
+
+### [FinOps hubs](hubs/finops-hubs-overview.md) v14
+
+- **Fixed**
+  - Fixed Init-DataFactory deployment script failing when an Event Grid subscription is already provisioning by checking subscription status before attempting subscribe/unsubscribe and polling separately for completion ([#1996](https://github.com/microsoft/finops-toolkit/issues/1996)).
+
+<br>
+
 ## v13 Update 1
 
 _Released February 11, 2026_
--- a/src/templates/finops-hub/modules/fx/scripts/Init-DataFactory.ps1
+++ b/src/templates/finops-hub/modules/fx/scripts/Init-DataFactory.ps1
@@ -78,38 +78,43 @@ function Set-BlobTriggerSubscription([string]$TriggerName, [switch]$Subscribe)
     $action = if ($Subscribe) { 'Subscribing' } else { 'Unsubscribing' }
 
     Write-Log "$action $TriggerName to events..."
-    try
-    {
-        Invoke-WithRetry -Name "$action $TriggerName" -Delay 5 -Action {
-            if ($Subscribe)
-            {
-                Add-AzDataFactoryV2TriggerSubscription `
-                    -ResourceGroupName $DataFactoryResourceGroup `
-                    -DataFactoryName $DataFactoryName `
-                    -Name $TriggerName | Out-Null
-            }
-            else
-            {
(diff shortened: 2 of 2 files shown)


RolandKrummenacher:
@slapointe As I didn't could/wanted to push to your fork, I've created a new PR here.

flanakin (approved):

Elgeario:
@RolandKrummenacher When is this due to be fixed? We have upgraded to V13 today and hit this issue. Is there a work around / release date of the fix?

RolandKrummenacher:
@Elgeario  This is part of release 14 which is planned to be released this or next week.

Elgeario:
@RolandKrummenacher , Thank you for the prompt response!

RolandKrummenacher:
@Elgeario V14 is now fully released. We had some issues that delayed the release. Sorry about that.

PR #2011, opened 2026-02-18
[Recommendations] Add parameters for noisy recommendations and rename query files

flanakin:
## 🛠️ Description

Add `enableAHBRecommendations` and `enableSpotRecommendations` parameters to control noisy recommendation types independently. Also renames query files to follow `Recommendations-Microsoft-*` naming convention, adds a recommendations tab to createUiDefinition.json for the Azure portal deployment experience, and switches `enableRecommendations` to default off (preview).

Fixes # N/A

> - [x] 🤏 Lint tests
> - [ ] 💪 Unit tests
> - [ ] 🙌 Integration tests

> - [x] ➡️ Will add log in a future PR (feature branch PRs only)

> - [x] ➡️ Will add docs in a future PR (feature branch PRs only)


src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Recommendations/queries/Recommendations-Microsoft-BackendlessAppGateways.json
```diff
@@ -1,10 +1,10 @@
...
+    "queryEngine": "ResourceGraph",
+    "scope": "Tenant",
+    "source": "FinOps hubs",
+    "type": "Microsoft-BackendlessAppGateways",
```
helderpinto, line 8: Why are we prefixing the type as being "Microsoft-*"? What ties these ARG queries to Microsoft? Is it the author? Should we add an "author" property instead?

flanakin: I prefixed anticipating a future where we have non-Microsoft recommendations as well. I wanted to differentiate from AWS and GCP recommendations when we get there or when customers ingest them themselves.

src/templates/finops-hub/modules/hub.bicep
```diff
@@ -313,14 +319,16 @@ module analytics 'Microsoft.FinOpsHubs/Analytics/app.bicep' = if (useFabric || u
 // Recommendations app
 //------------------------------------------------------------------------------
 
-module recommendations 'Microsoft.FinOpsHubs/Recommendations/app.bicep' = if (enableRecommendations && (useFabric || useAzureDataExplorer)) {
+module recommendations 'Microsoft.FinOpsHubs/Recommendations/app.bicep' = if (enableRecommendations) {
   name: 'Microsoft.FinOpsHubs.Recommendations'
   dependsOn: [
     core
     analytics
   ]
```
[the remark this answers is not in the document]
flanakin: Technically, Recommendations don't require Analytics. It will work with a storage-only hub instance. It's designed to align with Analytics, but it's not blocked on having it.

RolandKrummenacher:
I don't want to start a fundamental discussion here, but is it really the right way to disable recommendations at deploy-time? Shouldn't we just deploy them and have a mechanism for users to configure/disable alarms by something like recommendation type or severity?

In the environments that I work with, at least the AHUB recommendation is not noise.

RolandKrummenacher (commented):
## VMsWithoutAHB query: WHERE clause appears incorrect

In `Recommendations-Microsoft-VMsWithoutAHB.json` there is a problematic filter condition:

```kusto
| where tostring(properties.['licenseType']) !has 'Windows'
    and tostring(properties.virtualMachineProfile.['licenseType']) == 'Windows_Server'
```

`properties.licenseType` applies to **VMs**, while `properties.virtualMachineProfile.licenseType` applies to **VMSS**. The `and` conjunction causes:

- **VMs are never matched**, because `virtualMachineProfile.licenseType` is empty for VMs, so `== 'Windows_Server'` always evaluates to false.
- **VMSS with AHB enabled are matched** (instead of without), because `== 'Windows_Server'` matches exactly those VMSS that already have AHB activated — the opposite of the intent.

### Suggestion

The condition should probably be (both properties without AHB):

```kusto
| where tostring(properties.['licenseType']) !has 'Windows'
    and tostring(properties.virtualMachineProfile.['licenseType']) !has 'Windows'
```

This way both VMs and VMSS that do **not** have AHB enabled are found.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Recommendations/queries/Recommendations-Microsoft-VMsWithoutAHB.json
```diff
@@ -1,10 +1,10 @@
...
...ws' and (properties.virtualMachineProfile.['licenseType']) !has 'Windows' , 'AHB-disabled', 'AHB-enabled'),        'Not Windows'    )) on subscriptionId | project     x_RecommendationId=strcat(tolower(WindowsId),'-CheckAHBWindows'),     x_ResourceGroupName=tolower(vmResourceGroup),     SubAccountId=subscriptionId,     x_RecommendationCategory='Cost',     x_RecommendationDescription='Windows virtual machine is not leveraging Azure Hybrid Benefit',     ResourceId = tolower(WindowsId),     ResourceName=tolower(split(WindowsId,'/')[-1]),     x_RecommendationDetails= strcat('{\"VMSku\": \"', VMSku, '\", \"CheckAHBWindows\": \"', CheckAHBWindows, '\", \"ActualCores\": \"', ActualCores, '\", \"Location\": \"', Location, '\", \"x_RecommendationProvider\": \"FinOps hubs\", \"x_RecommendationSolution\": \"Review the virtual machine licensing option\", \"x_RecommendationTypeId\": \"f326c065-b9f7-4a0e-a0f1-5a1c060bc25d\", \"x_ResourceType\": \"', vmType, '\"}'),     x_RecommendationDate = now() | join kind=leftouter ( resourcecontainers | where type == 'microsoft.resources/subscriptions' | project SubAccountName=name, SubAccountId=subscriptionId ) on SubAccountId | project-away SubAccountId1",
```
RolandKrummenacher: ## WHERE clause appears incorrect

The `and` conjunction of these two conditions is problematic:

```kusto
| where tostring(properties.['licenseType']) \!has 'Windows'
    and tostring(properties.virtualMachineProfile.['licenseType']) == 'Windows_Server'
```

`properties.licenseType` applies to **VMs**, while `properties.virtualMachineProfile.licenseType` applies to **VMSS** ([MS docs ref](https://learn.microsoft.com/en-us/javascript/api/@azure/arm-compute/virtualmachinescalesetvmprofile?view=azure-node-latest)). The `and` conjunction causes:

| Resource | (A) `licenseType \!has 'Windows'` | (B) `virtualMachineProfile.licenseType == 'Windows_Server'` | A and B |
|---|---|---|---|
| **VM without AHB** | `true` | `false` (property does not exist) | **false** — not matched |
| **VM with AHB** | `false` | `false` | **false** |
| **VMSS without AHB** | `true` (empty) | `false` | **false** — not matched |
| **VMSS with AHB** | `true` (empty) | `true` | **true** — matched\! |

**Result: the query only finds VMSS _with_ AHB enabled** — the opposite of the intent.

### Suggestion

The condition should probably be:

```kusto
| where tostring(properties.['licenseType']) \!has 'Windows'
    and tostring(properties.virtualMachineProfile.['licenseType']) \!has 'Windows'
```

This way both VMs and VMSS that do **not** have AHB enabled are found.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Recommendations/queries/Recommendations-Microsoft-BackendlessAppGateways.json
```diff
@@ -1,10 +1,10 @@
...
...backendIPCount = sum(backendIPCount) ,backendAddressesCount=sum(backendAddressesCount) by id) on id| project-away id1| where  (backendIPCount == 0 or isempty(backendIPCount)) and (backendAddressesCount==0 or isempty(backendAddressesCount))| project     x_RecommendationId=strcat(tolower(id),'-idle'),     x_ResourceGroupName=tolower(resourceGroup),     SubAccountId=subscriptionId,     x_RecommendationCategory='Cost',     x_RecommendationDescription='Application Gateway without any backend pool',     ResourceId = tolower(id),     ResourceName=tolower(AppGWName),     x_RecommendationDetails= strcat('{\"backendIPCount\": ', backendIPCount, ', \"Location\": \"', Location, '\", \"x_RecommendationProvider\": \"FinOps hubs\", \"x_RecommendationSolution\": \"Review and remove this resource if not needed\", \"x_RecommendationTypeId\": \"4f69df93-5972-44e0-97cf-4343c2bcf4b8\", \"x_ResourceType\": \"', type, '\", \"x_RecommendationMaturityLevel\": \"Preview\"}'),     x_RecommendationDate = now() | join kind=leftouter ( resourcecontainers | where type == 'microsoft.resources/subscriptions' | project SubAccountName=name, SubAccountId=subscriptionId ) on SubAccountId | project-away SubAccountId1",
```
RolandKrummenacher: ## `resourceGroup` is overwritten with full ARM path

The `extend` overwrites the built-in `resourceGroup` column:

```kusto
| extend ...
    resourceGroup=strcat('/subscriptions/',subscriptionId,'/resourceGroups/',resourceGroup)
```

This causes `x_ResourceGroupName` to contain the full ARM path (`/subscriptions/xxx/resourcegroups/yyy`) instead of just the resource group name. All other recommendation queries produce only the simple name via `tolower(resourceGroup)`.

This is likely a copy-paste leftover and should be removed to keep the output consistent.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Recommendations/queries/Recommendations-Microsoft-StoppedVMs.json
```diff
@@ -1,10 +1,10 @@
...
...ew.powerState.displayStatus) != 'VM running' | extend PowerState=tostring(properties.extended.instanceView.powerState.displayStatus) | extend Location=location, type| project id, PowerState, Location, resourceGroup, subscriptionId, VMName=name, type| project     x_RecommendationId=strcat(tolower(id),'-notDeallocated'),     x_ResourceGroupName=tolower(resourceGroup),     SubAccountId=subscriptionId,     x_RecommendationCategory='Cost',     x_RecommendationDescription='Virtual machine is powered off but not deallocated',     ResourceId = tolower(id),     ResourceName=tolower(VMName),     x_RecommendationDetails= strcat('{\"PowerState\": \"', PowerState, '\", \"Location\": \"', Location, '\", \"x_RecommendationProvider\": \"FinOps hubs\", \"x_RecommendationSolution\": \"Deallocate the virtual machine to ensure it does not incur in compute costs\", \"x_RecommendationTypeId\": \"ab2ff882-e927-4093-9d11-631be0219975\", \"x_ResourceType\": \"', type, '\"}'),    x_RecommendationDate = now() | join kind=leftouter ( resourcecontainers | where type == 'microsoft.resources/subscriptions' | project SubAccountName=name, SubAccountId=subscriptionId ) on SubAccountId | project-away SubAccountId1",
```
RolandKrummenacher: ## Consider using `bag_pack()` instead of `strcat` for JSON construction

All queries except AdvisorCost build `x_RecommendationDetails` as a raw JSON string via `strcat`:

```kusto
x_RecommendationDetails = strcat('{"key": "', value, '", ...}')
```

This is fragile — if any value contains quotes, backslashes, or other JSON special characters, the resulting JSON will be malformed. **AdvisorCost** already uses the safer approach with `bag_pack()`:

```kusto
x_RecommendationDetails = bag_pack('key', value, ...)
```

Consider aligning all recommendation queries to use `bag_pack()` for consistency and robustness.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Recommendations/queries/HubsRecommendations-StoppedVMs.json
```diff
@@ -1,10 +1,10 @@
...
...tate.displayStatus) != 'VM running' | extend PowerState=tostring(properties.extended.instanceView.powerState.displayStatus) | extend Location=location, type| project id, PowerState, Location, resourceGroup, subscriptionId, VMName=name, type| project     x_RecommendationId=strcat(tolower(id),'-notDeallocated'),     x_ResourceGroupName=tolower(resourceGroup),     SubAccountId=subscriptionId,     x_RecommendationCategory='Cost',     x_RecommendationDescription='Virtual machine is powered off but not deallocated',     ResourceId = tolower(id),     ResourceName=tolower(VMName),     x_RecommendationDetails= strcat('{\"PowerState\": \"', PowerState, '\", \"Location\": \"', Location, '\", \"x_RecommendationProvider\": \"Azure Resource Graph\", \"x_RecommendationSolution\": \"Deallocate the virtual machine to ensure it does not incur in compute costs\", \"x_RecommendationTypeId\": \"ab2ff882-e927-4093-9d11-631be0219975\", \"x_ResourceType\": \"', type, '\"}'),    x_RecommendationDate = now() | join kind=leftouter ( resourcecontainers | where type == 'microsoft.resources/subscriptions' | project SubAccountName=name, SubAccountId=subscriptionId ) on SubAccountId | project-away SubAccountId1",
```
helderpinto, line 4: The power state condition is based on the display status, which is more prone to changes than the power state code. Also, we should explicitly look for "stopped" power states instead of power states not equal to deallocated or running, which may in some rare situations include "starting", "deallocating" or "stopping" states. Use the condition below instead:

`tolower(properties.extended.instanceView.powerState.code) =~ 'PowerState/stopped' `

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Recommendations/queries/HubsRecommendations-UnattachedPublicIPs.json
```diff
@@ -1,10 +1,10 @@
...
...s.publicIPAllocationMethod), SKUName=sku.name, resourceGroup, Location=location, name, id ) on PublicIpId | extend PublicIpId,IPName, SKUName, resourceGroup, Location, AllocationMethod,name, subscriptionId )| project     x_RecommendationId=strcat(tolower(PublicIpId),'-idle'),     x_ResourceGroupName=tolower(resourceGroup),     SubAccountId=subscriptionId,     x_RecommendationCategory='Cost',     x_RecommendationDescription='Unattached (orphaned) public IP is incurring in networking costs',     ResourceId = tolower(PublicIpId),     ResourceName=tolower(name),     x_RecommendationDetails= strcat('{\"SKUName\": \"', SKUName, '\", \"AllocationMethod\": \"', AllocationMethod, '\", \"Location\": \"', Location, '\", \"x_RecommendationProvider\": \"Azure Resource Graph\", \"x_RecommendationSolution\": \"Review and remove this resource if not needed\", \"x_RecommendationTypeId\": \"3ecbf770-9404-4504-a450-cc198e8b2a7d\", \"x_ResourceType\": \"', type, '\"}'),     x_RecommendationDate = now() | join kind=leftouter ( resourcecontainers | where type == 'microsoft.resources/subscriptions' | project SubAccountName=name, SubAccountId=subscriptionId ) on SubAccountId | project-away SubAccountId1",
```
helderpinto, line 4: Dynamically allocated Public IPs are being left out by the `properties.publicIPAllocationMethod =~ 'Static'` condition. Although orphan dynamic public IPs do not incur in costs, we should include them in the recommendation, for the sake of housekeeping.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Recommendations/queries/Recommendations-Microsoft-VMsWithoutAHB.json
helderpinto: Additionally, the condition targeting Windows machines is inconsistent between VMs and VMSS. Also, it is not considering machines based on custom images.

```kql
where tostring(properties.storageProfile.imageReference.publisher ) == 'MicrosoftWindowsServer'
  or tostring(properties.virtualMachineProfile.storageProfile.osDisk.osType) == 'Windows' 
  or tostring(properties.storageProfile.imageReference.publisher ) == 'microsoftsqlserver'
```

We should probably make the overall filter more consistent and with a broader coverage (the proposal below includes @RolandKrummenacher's suggestion):

```kql
    | where (tostring(properties.virtualMachineProfile.storageProfile.osDisk.osType) == 'Windows'
               and tostring(properties.virtualMachineProfile.storageProfile.imageReference.publisher) !in~ ('microsoftwindowsdesktop','microsoftvisualstudio')
               and tostring(properties.virtualMachineProfile.licenseType) !startswith 'Windows')
        or (tostring(properties.storageProfile.osDisk.osType) == 'Windows' 
              and tostring(properties.storageProfile.imageReference.publisher) !in~ ('microsoftwindowsdesktop','microsoftvisualstudio')
              and tostring(properties.licenseType) !startswith 'Windows')
```

helderpinto (changes requested):

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Recommendations/queries/Recommendations-Microsoft-VMsWithoutAHB.json
flanakin: 🤖 [AI][Claude] ✅ **Implemented**

Implemented @helderpinto's comprehensive fix — the WHERE clause now uses separate VM/VMSS branches with `osType`-based Windows detection, excludes desktop/Visual Studio publishers (which can't use AHB), and uses `!startswith 'Windows'` for the licenseType check. Also converted `strcat` to `bag_pack()` per thread 5.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Recommendations/queries/Recommendations-Microsoft-BackendlessAppGateways.json
flanakin: 🤖 [AI][Claude] ✅ **Implemented**

Removed the `resourceGroup=strcat('/subscriptions/',subscriptionId,'/resourceGroups/',resourceGroup)` override. `x_ResourceGroupName` now gets just the resource group name via `tolower(resourceGroup)`, consistent with all other queries.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Recommendations/queries/Recommendations-Microsoft-StoppedVMs.json
flanakin: 🤖 [AI][Claude] ✅ **Implemented**

Converted all 9 non-AdvisorCost queries from `strcat`-based JSON construction to `bag_pack()`, aligning with AdvisorCost's existing pattern.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Recommendations/queries/HubsRecommendations-StoppedVMs.json
flanakin: 🤖 [AI][Claude] ✅ **Implemented**

Replaced the `displayStatus` checks with `tolower(tostring(properties.extended.instanceView.powerState.code)) =~ 'powerstate/stopped'` to explicitly match only stopped (not deallocated) VMs.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Recommendations/queries/HubsRecommendations-UnattachedPublicIPs.json
flanakin: 🤖 [AI][Claude] ✅ **Implemented**

Removed the `publicIPAllocationMethod =~ 'Static'` filter so all unattached public IPs are included for housekeeping. Updated the description to "Unattached (orphaned) public IP" (removed the cost-specific framing since dynamic IPs don't incur costs).

flanakin:
🤖 [AI][Claude] 🤔 **Needs discussion**

@RolandKrummenacher Great point — the deploy-time parameters were a quick workaround to give users some control, but ideally these should be configurable at runtime without requiring redeployment. That would be a larger change though.

One idea we're considering is adding an Informational severity column that could help distinguish recommendations that may be noisy in some environments from actionable ones.

Curious — can you share more about why AHB recommendations aren't noisy in your environments? We assumed they'd generate a high volume of results that could overwhelm the signal, but maybe we're thinking about it wrong. Understanding your use case would help us design the right approach.

flanakin:
🤖 [AI][Claude] **PR Update Summary**

**Addressed:** 6 thread(s)

- ✅ Implemented: 5
- 🤔 Needs discussion: 1
- ❓ Questions: 0

**Key changes:**
- Fixed VMsWithoutAHB WHERE clause (comprehensive fix covering VM/VMSS separately, osType detection, publisher exclusions)
- Fixed BackendlessAppGateways resourceGroup being overwritten with full ARM path
- Replaced displayStatus with powerState.code for StoppedVMs
- Included dynamic public IPs in UnattachedPublicIPs for housekeeping
- Converted all 9 queries from strcat to bag_pack() for robust JSON construction

RolandKrummenacher:
> 🤖 [AI][Claude] 🤔 **Needs discussion**
> 
> @RolandKrummenacher Great point — the deploy-time parameters were a quick workaround to give users some control, but ideally these should be configurable at runtime without requiring redeployment. That would be a larger change though.
> 
> One idea we're considering is adding an Informational severity column that could help distinguish recommendations that may be noisy in some environments from actionable ones.
> 
> Curious — can you share more about why AHB recommendations aren't noisy in your environments? We assumed they'd generate a high volume of results that could overwhelm the signal, but maybe we're thinking about it wrong. Understanding your use case would help us design the right approach.

In a lift-and-shift environment, you traditionally have a lot of standalone VMs for your Web-, App- and Database Servers with Windows and SQL Server running on top of it, running 24/7. For those environments, it's crucial to have AHUB applied to all of you servers (since those organizations usually have enough dual-use licenses). Once AHUB has been activated on all VMs, having a way to monitor that no **new** VM without AHUB gets deployed would be useful.

RolandKrummenacher (approved):

helderpinto (approved):

PR #2037, opened 2026-03-02
Fix: Prices transform cross-row operations broken by per-batch execution

### LINKED ISSUE #1625: Prices_transform_v1_0 function does not work as intended
## 🐛 Problem
The `Prices_transform_v1_0` function transforms `Prices_raw` into `Prices_final_v1_0`. However, the function references the entire `Prices_raw` table multiple times. If the pricesheet is ingested as multiple `parquet-snappy` files, the transformation function appears to be executed separately per file. As a result, it never has access to the full price table during transformation.

## 👣 Repro steps
1. Create a modified version of `Prices_transform_v1_0` with the following content:
   ```kql
   .create-or-alter function
     with (docstring='Transforms Prices_raw into FOCUS 1.0.', folder='Prices')
     Prices_transform_v1_0() {
       let prices = materialize(
         Prices_raw
       );
       prices
       | summarize count()
     }

2. Ingest a single pricesheet consisting of two `parquet-snappy` files.
3. Observe the resulting records in `Prices_final_v1_0`.

## 🤔 Expected

If the transformation would have access to the full dataset (i.e., the union of all rows across the multiple files) we should see a single summarized result based on the total record count. This is not the case. 

Therefore, the original transformation function does not work as intended. The transformation function can not reference the whole source table, for example to get a complete list of meterIds for SP or RIs

In particular the following statements don't work:

```KQL
let riMeters = prices | where x_SkuPriceType == 'ReservedInstance' | distinct x_SkuMeterId;
let spMeters = prices | where x_SkuPriceType == 'SavingsPlan' | distinct x_SkuMeterId;

prices
| where x_SkuPriceType == 'SavingsPlan'
| lookup kind=leftouter (prices | where x_SkuPriceType == 'Consumption'...
```

## 📷 Screenshots

Image

## ℹ️ Additional context

The screenshot demonstrates that the function was executed separately on each of the two files, producing two `count` outputs: 866,482 and 115,098. These two results align exactly with the expected row count of the individual files, but not the full dataset total of 981,580.

## 🙋‍♀️ Ask for the community

We could use your help:

RolandKrummenacher:
## Summary

Fixes #1625 — The `Prices_transform_v1_0` and `Prices_transform_v1_2` functions don't work as intended because ADX update policies execute per-ingestion batch, not across the full `Prices_raw` table. When a price sheet is split across multiple parquet files, cross-row operations only see the current file's data.

This PR fixes two issues:

### 1. Removed the savings plan → consumption price lookup

The update policy previously nulled out `ContractedUnitPrice` and `ListUnitPrice` for savings plan rows, then replaced them with on-demand prices from matching Consumption rows via a self-join. This lookup was broken by per-batch execution (SP and Consumption rows could land in different files).

**Why we dropped it rather than moving it:** Per the [EA](https://learn.microsoft.com/azure/cost-management-billing/dataset-schema/price-sheet-ea) and [MCA](https://learn.microsoft.com/azure/cost-management-billing/dataset-schema/price-sheet-mca) price sheet documentation, savings plan rows already have meaningful values:

- `UnitPrice` → the SP contracted rate (negotiated, including EA/MCA discounts)
- `MarketPrice` → the SP retail rate (before negotiated discounts)
- `BasePrice` → the rate at agreement signing

These are the correct prices **for the savings plan price type itself**. The old lookup was replacing them with on-demand reference prices to enable "SP vs. on-demand" savings analysis — but that conflated different price semantics on the same row. Comparing SP rates against on-demand rates is a cross-row analysis that belongs at query time, not baked into the data.

With this change, each price row now has its own correct prices regardless of price type (Consumption, ReservedInstance, or SavingsPlan), and discount columns show the EA/MCA negotiated discount for that specific price type.

### 2. Moved commitment discount eligibility to Hub view functions

The `riMeters`/`spMeters` distinct lookups (`prices | where x_SkuPriceType == 'ReservedInstance' | distinct x_SkuMeterId`) require visibility across all price data to determine which meters have RI/SP variants. These have been moved from the update policy to the `Prices_v1_2()` and `Prices_v1_0()` Hub view functions, which query the full final tables.

Additional fix: eligibility is now set only on `Consumption` rows (`x_SkuPriceType == 'Consumption'`). Previously, spend eligibility excluded RI rows but usage eligibility did not exclude SP rows. Since there are only three price types (Consumption, ReservedInstance, SavingsPlan), eligibility is meaningful only for on-demand rows — RI/SP rows *are* the commitment discounts themselves.

### Impact on downstream consumers

**`Prices_v1_2()` / `Prices_v1_0()` ADX functions** — these are the public-facing functions that users query directly. The semantic change for SP rows:

| Column | Before (when lookup worked) | After this PR |
|--------|---------------------------|---------------|
| `ContractedUnitPrice` | Consumption contracted rate | SP contracted rate |
| `ListUnitPrice` | Consumption retail rate | SP retail rate |
| `x_EffectiveUnitPrice` | SP contracted rate | SP contracted rate (unchanged) |
| `x_EffectiveUnitPriceDiscount` | Consumption - SP = SP savings | SP - SP = **0** |
| `x_ContractedUnitPriceDiscount` | Retail - Contracted (negotiated) | SP retail - SP contracted |
| `x_TotalUnitPriceDiscount` | Retail - SP = total savings | SP retail - SP contracted |

Note: the old behavior was already broken for most users due to the per-batch issue (#1625). The lookup only produced correct results when all price data landed in a single ingestion batch.

**Costs transform functions** — The `Costs_transform_v1_2` and `Costs_transform_v1_0` functions look up prices from `Prices_final` to populate missing `ContractedUnitPrice`/`ListUnitPrice` on cost rows. This lookup is **not affected** because it filters on `x_SkuPriceType == 'Consumption'`, and Consumption rows were never nulled — only SP rows were.

**Power BI reports** — Not affected:
- Storage-based reports read raw parquet files and do their own column mapping — they never go through the KQL transform.
- KQL-based reports have no Prices data table; the Costs table gets `ContractedUnitPrice`/`ListUnitPrice` from cost exports, not price sheets.

**ADX real-time dashboard** (`dashboard.json`) — Not directly affected. The dashboard's SKU prices and cost validation tiles query Costs data (`CostsByDay`, `Costs_v1_2`), not the Prices functions. The only Prices reference is a data ingestion monitoring query that checks table extents, not price values.

### Files changed

- **`IngestionSetup_v1_2.kql`** — Remove SP price nulling, cross-row lookup, and eligibility calculation from update policy
- **`IngestionSetup_v1_0.kql`** — Same changes (deprecated but kept consistent)
- **`HubSetup_v1_2.kql`** — Add eligibility computation with combined meter lookup across both final tables
- **`HubSetup_v1_0.kql`** — Same eligibility computation for FOCUS 1.0 consumers

## Test plan

- [ ] Deploy to a test hub with a price sheet split across multiple parquet files
- [ ] Verify `Prices_final_v1_2` has non-null `ContractedUnitPrice`/`ListUnitPrice` for all price types including SavingsPlan
- [ ] Verify `Prices_v1_2()` returns correct `x_CommitmentDiscountSpendEligibility` and `x_CommitmentDiscountUsageEligibility` on Consumption rows
- [ ] Verify RI and SP rows show `Not Eligible` for both eligibility columns
- [ ] Verify SP rows: `ContractedUnitPrice` = SP contracted rate, `ListUnitPrice` = SP retail rate
- [ ] Verify SP rows: `x_EffectiveUnitPriceDiscount` = 0 (expected, since ContractedUnitPrice and x_EffectiveUnitPrice are now the same value)
- [ ] Verify Consumption rows: discount columns unchanged
- [ ] Verify Costs transform still populates missing prices correctly (Consumption lookup unaffected)
- [ ] Verify ADX dashboard SKU prices tiles render correctly

### DIFF
--- a/docs-mslearn/toolkit/changelog.md
+++ b/docs-mslearn/toolkit/changelog.md
@@ -47,8 +47,9 @@ The following section lists features and enhancements that are currently in deve
   - Remote Hub configuration (storage URI, storage key, and purge protection) is now displayed in the Basics tab when Remote Hub mode is selected, making the mutual exclusivity clear.
   - Data Explorer SKU and retention settings are now only visible when Azure Data Explorer mode is selected.
 - **Fixed**
-  - Fixed prices transform function producing incorrect results when price sheet data is split across multiple ingestion batches by moving cross-row operations (commitment discount eligibility) from the ingestion-time update policy to the query-time Hub view functions ([#1625](https://github.com/microsoft/finops-toolkit/issues/1625)).
-    - Stopped nulling savings plan price columns (`ContractedUnitPrice`, `ListUnitPrice`) since the raw values from Cost Management (`UnitPrice`, `MarketPrice`) already contain the correct contracted and retail rates for all price types, including savings plans.
+  - Fixed price data ingestion bugs when price sheet data is split across multiple files ([#1625](https://github.com/microsoft/finops-toolkit/issues/1625)).
+    - Commitment discount prices are now consistently set correctly using a full price sheet lookup.
+    - Commitment discount eligibility columns moved to the Prices* Hub functions rather than hard-coded in the `Prices_final_v*` tables.
 
 ### [PowerShell module](powershell/powershell-commands.md) v14
 
--- a/docs-mslearn/toolkit/changelog.md
+++ b/docs-mslearn/toolkit/changelog.md
@@ -46,6 +46,9 @@ The following section lists features and enhancements that are currently in deve
   - Improved deployment UI to consolidate hub mode selection into a single radio button group with four mutually exclusive options: None (storage only for Power BI reports), Azure Data Explorer, Microsoft Fabric, or Remote Hub ([#1929](https://github.com/microsoft/finops-toolkit/issues/1929)).
   - Remote Hub configuration (storage URI, storage key, and purge protection) is now displayed in the Basics tab when Remote Hub mode is selected, making the mutual exclusivity clear.
   - Data Explorer SKU and retention settings are now only visible when Azure Data Explorer mode is selected.
+- **Fixed**
+  - Fixed prices transform function producing incorrect results when price sheet data is split across multiple ingestion batches by moving cross-row operations (commitment discount eligibility) from the ingestion-time update policy to the query-time Hub view functions ([#1625](https://github.com/microsoft/finops-toolkit/issues/1625)).
+    - Stopped nulling savings plan price columns (`ContractedUnitPrice`, `ListUnitPrice`) since the raw values from Cost Management (`UnitPrice`, `MarketPrice`) already contain the correct contracted and retail rates for all price types, including savings plans.
 
 ### [PowerShell module](powershell/powershell-commands.md) v14
 
--- a/docs-mslearn/toolkit/changelog.md
+++ b/docs-mslearn/toolkit/changelog.md
@@ -3,7 +3,7 @@ title: FinOps toolkit changelog
 description: Review the latest features and enhancements in the FinOps toolkit, including updates to FinOps hubs, Power BI reports, and more.
 author: MSBrett
 ms.author: brettwil
-ms.date: 03/01/2026
+ms.date: 03/02/2026
 ms.topic: reference
 ms.service: finops
 ms.subservice: finops-toolkit
--- a/src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts/HubSetup_v1_0.kql
+++ b/src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts/HubSetup_v1_0.kql
@@ -216,10 +216,22 @@ Costs_v1_0()
 
 
 // Prices_final_v1_0
+// Commitment discount eligibility is computed here instead of in the update policy because ADX
+// update policies execute per-ingestion batch, not across the full table.
+// See https://github.com/microsoft/finops-toolkit/issues/1625
 .create-or-alter function
 with (docstring = 'Gets all prices aligned to FOCUS 1.0.', folder = 'Prices')
 Prices_v1_0()
 {
+    // Commitment discount eligibility -- computed here because the update policy only sees per-batch data
+    let commitmentMeters = database('Ingestion').Prices_final_v1_0
+        | union (database('Ingestion').Prices_final_v1_2)
+        | where x_SkuPriceType in ('ReservedInstance', 'SavingsPlan')
+        | distinct x_SkuMeterId, x_SkuPriceType;
+    let riMeters = commitmentMeters | where x_SkuPriceType == 'ReservedInstance' | project x_SkuMeterId;
+    let spMeters = commitmentMeters | where x_SkuPriceType == 'SavingsPlan' | project x_SkuMeterId;
+    //
+    // Calculate commitment discount eligibility
     database('Ingestion').Prices_final_v1_0
     | union (
         database('Ingestion').Prices_final_v1_2
@@ -238,12 +250,15 @@ Prices_v1_0()
             x_SkuIncludedQuantity                = todecimal(x_SkuIncludedQuantity),
             x_SkuTier                            = todecimal(x_SkuTier),
             x_TotalUnitPriceDiscount             = todecimal(x_TotalUnitPriceDiscount),
-            x_TotalUnitPriceDiscountPercent      = todecimal(x_TotalUnitPriceDiscountPercent) 
+            x_TotalUnitPriceDiscountPercent      = todecimal(x_TotalUnitPriceDiscountPercent)
         // Rename columns
         | project-rename
             x_PricingCurrency = PricingCurrency,
             x_SkuMeterName = SkuMeter
     )
+    | extend x_CommitmentDiscountSpendEligibility = iff(x_SkuMeterId in (riMeters) and x_SkuPriceType == 'Consumption', 'Eligible', 'Not Eligible')
+    | extend x_CommitmentDiscountUsageEligibility = iff(x_SkuMeterId in (spMeters) and x_SkuPriceType == 'Consumption', 'Eligible', 'Not Eligible')
+    //
     | project
         BillingAccountId,
         BillingAccountName,
--- a/src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts/IngestionSetup_v1_0.kql
+++ b/src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts/IngestionSetup_v1_0.kql
@@ -78,11 +78,11 @@ Prices_transform_v1_0()
     );
     //
     // NOTE: Commitment discount eligibility requires cross-row lookups and is deferred to the Hub
-    // view (Prices_v1_2) because ADX update policies execute per-ingestion batch, not across the
+    // view (Prices_v1_0) because ADX update policies execute per-ingestion batch, not across the
     // full table. See https://github.com/microsoft/finops-toolkit/issues/1625
     prices
     //
-    // Commitment discount eligibility is computed in the Hub view (Prices_v1_2)
+    // Commitment discount eligibility is computed in the Hub view (Prices_v1_0)
     | extend x_CommitmentDiscountSpendEligibility = ''
     | extend x_CommitmentDiscountUsageEligibility = ''
     //
--- a/src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts/IngestionSetup_v1_0.kql
+++ b/src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts/IngestionSetup_v1_0.kql
@@ -61,8 +61,8 @@ Prices_transform_v1_0()
             x_SkuRegion = MeterRegion,
             x_SkuServiceFamily = ServiceFamily,
             x_SkuTier = TierMinimumUnits
-        | extend ContractedUnitPrice = iff(x_SkuPriceType != 'SavingsPlan', UnitPrice, todecimal(''))  // UnitPrice for savings plan is not the on-demand unit price
-        | extend ListUnitPrice = iff(x_SkuPriceType != 'SavingsPlan', MarketPrice, todecimal(''))  // MarketPrice for savings plan is not the list price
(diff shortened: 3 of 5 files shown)


src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts/IngestionSetup_v1_0.kql
```diff
@@ -72,32 +72,19 @@ Prices_transform_v1_0()
...
     );
     //
-    // Meters for reservations and savings plans to identify commitment eligibility
-    let riMeters = prices | where x_SkuPriceType == 'ReservedInstance' | distinct x_SkuMeterId;
-    let spMeters = prices | where x_SkuPriceType == 'SavingsPlan' | distinct x_SkuMeterId;
-    //
-    // Copy list/base/contracted prices from on-demand SKUs
+    // NOTE: Commitment discount eligibility requires cross-row lookups and is deferred to the Hub
+    // view (Prices_v1_2) because ADX update policies execute per-ingestion batch, not across the
+    // full table. See https://github.com/microsoft/finops-toolkit/issues/1625
```
flanakin, line 82: 🤖 [AI][Claude] ⚠️ **Should fix**

These comments reference `Prices_v1_2` but this is the v1_0 transform file. Should reference `Prices_v1_0` (or generically "the Hub view functions") for consistency. Same issue on line 85 below.

**Suggestion:**
```suggestion
    // NOTE: Commitment discount eligibility requires cross-row lookups and is deferred to the Hub
    // view (Prices_v1_0) because ADX update policies execute per-ingestion batch, not across the
    // full table. See https://github.com/microsoft/finops-toolkit/issues/1625
```

flanakin (commented):
🤖 [AI][Claude] **PR Review**

**Summary:** Well-structured fix that correctly addresses the per-batch execution limitation in ADX update policies. The approach of moving cross-row eligibility lookups to query-time Hub view functions is sound, and the decision to stop nulling SP price columns is well-justified by the documented schema semantics.

### ⚠️ Should fix (1)
1. Copy-paste comment in `IngestionSetup_v1_0.kql` references `Prices_v1_2` instead of `Prices_v1_0` (lines 81, 85)

### ✅ What's done well
- Thorough PR description with clear before/after semantic table and downstream impact analysis
- Correct architectural fix — moving cross-row operations to query-time Hub views
- Consistent eligibility logic across both v1_0 and v1_2 paths, with both Hub views unioning across both final tables
- Tightening eligibility to `x_SkuPriceType == 'Consumption'` only is semantically cleaner than the old asymmetric exclusion logic

docs-mslearn/toolkit/changelog.md
```diff
@@ -46,6 +46,9 @@ The following section lists features and enhancements that are currently in deve
   - Improved deployment UI to consolidate hub mode selection into a single radio button group with four mutually exclusive options: None (storage only for Power BI reports), Azure Data Explorer, Microsoft Fabric, or Remote Hub ([#1929](https://github.com/microsoft/finops-toolkit/issues/1929)).
   - Remote Hub configuration (storage URI, storage key, and purge protection) is now displayed in the Basics tab when Remote Hub mode is selected, making the mutual exclusivity clear.
   - Data Explorer SKU and retention settings are now only visible when Azure Data Explorer mode is selected.
+- **Fixed**
+  - Fixed prices transform function producing incorrect results when price sheet data is split across multiple ingestion batches by moving cross-row operations (commitment discount eligibility) from the ingestion-time update policy to the query-time Hub view functions ([#1625](https://github.com/microsoft/finops-toolkit/issues/1625)).
```
flanakin: nit: This is a bit verbose. 
```suggestion
  - Fixed price data ingestion bugs when price sheet data is split across multiple files ([#1625](https://github.com/microsoft/finops-toolkit/issues/1625)).
    - Commitment discount prices are now consistently set correctly using a full price sheet lookup.
    - Commitment discount eligibility columns moved to the Prices* Hub functions rather than hard-coded in the `Prices_final_v*` tables.
```

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts/IngestionSetup_v1_0.kql
```diff
@@ -61,8 +61,8 @@ Prices_transform_v1_0()
             x_SkuRegion = MeterRegion,
             x_SkuServiceFamily = ServiceFamily,
             x_SkuTier = TierMinimumUnits
-        | extend ContractedUnitPrice = iff(x_SkuPriceType != 'SavingsPlan', UnitPrice, todecimal(''))  // UnitPrice for savings plan is not the on-demand unit price
-        | extend ListUnitPrice = iff(x_SkuPriceType != 'SavingsPlan', MarketPrice, todecimal(''))  // MarketPrice for savings plan is not the list price
+        | extend ContractedUnitPrice = UnitPrice   // Negotiated rate for this price type (on-demand, RI, or SP)
+        | extend ListUnitPrice = MarketPrice       // Retail rate for this price type (on-demand, RI, or SP)
```
flanakin: Did you validate these are correct for savings plan prices? If memory serves, these didn't have valid values. If they do now for EA and MCA, we should be good to remove the extend. Fwiw, I would move these into the project-rename above, if we are removing this (in alpha order).

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts/IngestionSetup_v1_2.kql
```diff
@@ -52,8 +52,8 @@ Prices_transform_v1_2()
             x_SkuRegion = MeterRegion,
             x_SkuServiceFamily = ServiceFamily,
             x_SkuTier = TierMinimumUnits
-        | extend ContractedUnitPrice = iff(x_SkuPriceType != 'SavingsPlan', UnitPrice, real(null))  // UnitPrice for savings plan is not the on-demand unit price
-        | extend ListUnitPrice = iff(x_SkuPriceType != 'SavingsPlan', MarketPrice, real(null))  // MarketPrice for savings plan is not the list price
+        | extend ContractedUnitPrice = UnitPrice   // Negotiated rate for this price type (on-demand, RI, or SP)
+        | extend ListUnitPrice = MarketPrice       // Retail rate for this price type (on-demand, RI, or SP)
```
flanakin: Same comment as in the v1_0 file

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts/HubSetup_v1_0.kql
flanakin, line 1: Fwiw, I'm slightly torn on this... In one sense, I like providing this and I believe we'll have to do it for FOCUS 1.4 anyway, but I also hesitate to add this to EVERY Prices call. Given it's not working today, I would argue we could leave it empty and plan a separate fix based on priority. We just need to create a bug to track it.

I'm curious what others think.

flanakin (changes requested):

RolandKrummenacher:
## Data validation: Savings Plan price columns

Verified the PR's claims against actual EA and MCA price sheet data.

### EA account (price sheet in `Prices_raw`)

**E64as v7 (JA East)** — same meter across all price types:

| PriceType | Term | UnitPrice | MarketPrice | BasePrice | UnitOfMeasure |
|---|---|---|---|---|---|
| Consumption | — | 64.01 | 67.38 | 67.38 | 10 Hours |
| ReservedInstance | P1Y | 34,826 | 34,826 | null | 1 Hour |
| ReservedInstance | P3Y | 67,292 | 67,292 | null | 1 Hour |
| SavingsPlan | P1Y | 45.82 | 45.8184 | null | 10 Hours |
| SavingsPlan | P3Y | 31.67 | 31.6686 | null | 10 Hours |

- SP rows carry their own meaningful rates — `UnitPrice` (45.82) < Consumption (64.01), confirming these are the SP-discounted rates.
- `UnitPrice` ≈ `MarketPrice` with minor rounding (45.82 vs 45.8184) — no significant EA negotiated discount visible on SP rates.

### MCA account without MACC discount

**D2 v3 (NO West)**:

| PriceType | Term | UnitPrice | MarketPrice | BasePrice |
|---|---|---|---|---|
| Consumption | — | 0.172 | 0.172 | 0.172 |
| SavingsPlan | P1Y | 0.13072 | 0.13072 | null |
| SavingsPlan | P3Y | 0.0946 | 0.0946 | null |

- `UnitPrice` == `MarketPrice` exactly (no negotiated discount, as expected).
- SP P1Y rate (0.131) vs Consumption (0.172) = ~24% savings.

### Correction needed in PR description

The PR states:
> `BasePrice` → the rate at agreement signing

This is **incorrect for SP and RI rows** — `BasePrice` is **null** for both SavingsPlan and ReservedInstance price types on both EA and MCA accounts. `BasePrice` is only populated on Consumption rows. The PR description should be updated to reflect this.

### Conclusion

The core fix is validated — SP rows do carry their own correct prices in `UnitPrice` and `MarketPrice`, so removing the cross-row consumption lookup is the right approach.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts/IngestionSetup_v1_0.kql
RolandKrummenacher: Yes, validated on both EA and MCA — see my [data validation comment](https://github.com/microsoft/finops-toolkit/pull/2037#issuecomment-4060450185). SP rows have valid UnitPrice (SP contracted rate) and MarketPrice (SP retail rate).

Moved both into `project-rename` in alpha order and updated the downstream `UnitPrice` reference to use `ContractedUnitPrice`.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts/IngestionSetup_v1_2.kql
RolandKrummenacher: Done — same changes applied to both v1_0 and v1_2.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts/HubSetup_v1_0.kql
RolandKrummenacher: Agreed — let's defer this. Will remove the eligibility computation from the Hub views and create a separate issue to track it.

PR #2013, opened 2026-02-18
Add Claude Code plugin for FinOps hubs and Azure Cost Management

MSBrett:
## Summary

- Added Claude Code plugin with 4 agents (CFO, FinOps practitioner, database query, hubs agent), 4 commands, 2 skills (finops-toolkit, azure-cost-management), and an output style
- Skills include curated reference docs for Azure Advisor, savings plans, reservations, commitment discount decisions, budgets, cost exports, anomaly alerts, credits, and MACC
- KQL query catalog with 17 pre-built queries for cost analysis, anomaly detection, forecasting, and rate optimization
- Symlinked docs-mslearn and src/queries into the plugin for single-source-of-truth content
- Plugin located in src/templates/claude-plugin for packaging via Package-Toolkit.ps1

## Test plan

- [ ] Verify plugin loads correctly in Claude Code (`claude install-plugin`)
- [ ] Verify symlinks resolve correctly (queries, docs-mslearn)
- [ ] Verify SKILL.md reference links are valid relative paths
- [ ] Verify Package-Toolkit.ps1 packages the plugin into a zip
- [ ] Test KQL queries against a FinOps hub ADX cluster

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2043, opened 2026-03-04
Add Claude Code plugin for FinOps hubs and Azure Cost Management

MSBrett:
## Summary

Retry of #2013 after revert in #2023. Adds a Claude Code plugin with skills for FinOps hubs (KQL-based cost analysis) and Azure Cost Management (optimization, commitments, orphaned resources).

### What's included

- **2 skills**: FinOps toolkit (hubs queries, deployment, FOCUS) and Azure Cost Management (Advisor, reservations, savings plans, orphaned resources)
- **4 agents**: CFO, FinOps practitioner, database query, hubs agent
- **5 commands**: cost-optimization, hubs-connect, hubs-healthCheck, mom-report, ytd-report
- **1 output style**: executive formatting
- **Version sync**: `Update-Version.ps1` now updates `marketplace.json` plugin versions

### Issues from #2023 addressed

- CLAUDE.md restored (symlink to AGENTS.md)
- Packaging verified: `Build-Toolkit claude-plugin` resolves symlinks correctly (pwsh 7 `Copy-Item -Recurse` follows symlinks natively)
- Plugin versioning aligned at 13.0.0
- PR-2013-REVIEW-AUDIT.md removed

## Test plan

- [x] Clean build: `Build-Toolkit claude-plugin` produces 293 files, 0 symlinks
- [x] Symlinks resolved to real content in release output
- [x] End-to-end plugin validation
- [ ] ❎ Log not needed


RolandKrummenacher (commented):
Review of plugin structure, build integration, security, and test coverage. The domain knowledge in the reference docs is excellent and technically accurate. A few items to address below.

src/templates/claude-plugin/.claude-plugin/plugin.json
```diff
@@ -0,0 +1,23 @@
...
+  "repository": "https://github.com/microsoft/finops-toolkit",
+  "license": "MIT",
+  "keywords": ["finops", "cost-management", "reservations", "savings-plans", "cloud-optimization", "commitments", "credits", "macc"],
+  "commands": "./commands/",
+  "agents": ["./agents/chief-financial-officer.md", "./agents/finops-practitioner.md", "./agents/ftk-database-query.md", "./agents/ftk-hubs-agent.md"],
+  "skills": "./skills/",
+  "mcpServers": {
+    "azure-mcp-server": {
+      "command": "npx",
+      "args": ["-y", "@azure/mcp@latest", "server", "start", "--namespace", "kusto", "--read-only"]
```
RolandKrummenacher, line 19: Using `@azure/mcp@latest` is a supply-chain risk — a breaking or compromised release would silently affect all users. Consider pinning to a specific version (e.g., `@azure/mcp@0.2.1`) so upgrades are intentional and reviewable.

.gitignore
```diff
@@ -370,3 +370,9 @@ env/
 
 # AI
 .claude/settings.local.json
+
+# Internal planning docs
+TODO.md
```
RolandKrummenacher: These are very generic filenames. As written, they'll be ignored anywhere in the repo tree — if someone legitimately adds a `TODO.md` in a subdirectory, it would be silently excluded.

Consider scoping to the repo root:
```gitignore
/TODO.md
/PRD.md
/RTM.md
/RETROSPECTIVES.md
```

src/scripts/Update-Version.ps1
```diff
@@ -113,6 +113,32 @@ if ($update -or $Version)
...
+        Write-Verbose "- $($_.FullName.Replace($repoRoot + [IO.Path]::DirectorySeparatorChar, ''))"
+        $json = Get-Content $_ -Raw | ConvertFrom-Json
+        $json.version = $ver
+        $json | ConvertTo-Json -Depth 10 | Out-File $_ -Encoding utf8NoBOM -NoNewline
+    }
+
+    # Update version in marketplace.json plugin entries
+    Write-Verbose "Updating marketplace.json files..."
+    Get-ChildItem $repoRoot -Include "marketplace.json" -Recurse -Force `
+    | Where-Object { $_.FullName -like "*.claude-plugin*" } `
```
RolandKrummenacher: The glob `*.claude-plugin*` is loose — it matches any path containing `claude-plugin` anywhere. Today this only hits `.claude-plugin/marketplace.json`, but a future file like `docs/claude-plugin-guide.json` would also match.

Consider a more targeted filter, e.g.:
```powershell
| Where-Object { $_.Directory.Name -eq '.claude-plugin' }
```

src/scripts/Update-Version.ps1
```diff
@@ -113,6 +113,32 @@ if ($update -or $Version)
...
 
+    # Update version in plugin.json files
+    Write-Verbose "Updating plugin.json files..."
+    Get-ChildItem $repoRoot -Include "plugin.json" -Recurse -Force `
+    | Where-Object { $_.FullName -like "*claude-plugin*" } `
+    | ForEach-Object {
+        Write-Verbose "- $($_.FullName.Replace($repoRoot + [IO.Path]::DirectorySeparatorChar, ''))"
+        $json = Get-Content $_ -Raw | ConvertFrom-Json
+        $json.version = $ver
+        $json | ConvertTo-Json -Depth 10 | Out-File $_ -Encoding utf8NoBOM -NoNewline
```
RolandKrummenacher: `-Encoding utf8NoBOM` is PowerShell 6+ only. On Windows PowerShell 5.1 this will throw a `ParameterBindingException`. If the build system guarantees pwsh 7+, this is fine — but it may be worth adding a comment or a `$PSVersionTable` guard, since other parts of this script use `-Encoding ascii` / `-Encoding utf8` which work on both versions.

src/templates/agent-skills/.build.config
```diff
@@ -0,0 +1,3 @@
+{
+  "unversionedZip": true
```
RolandKrummenacher, line 2: Both `agent-skills/` and `claude-plugin/` have `.build.config` files, so `Build-Toolkit` will process both as separate templates. Since `claude-plugin/skills/` symlinks into `agent-skills/`, the release package may contain duplicate content (once standalone from `agent-skills/`, once embedded via symlink resolution in `claude-plugin/`).

Is the intent to ship both as separate release artifacts? If `agent-skills/` is only a source-of-truth consumed via symlinks, it may not need its own `.build.config`.

src/templates/agent-skills/azure-cost-management/references/azure-reservations.md
```diff
@@ -0,0 +1,353 @@
...
+description: Query the Azure Cost Management Benefit Recommendations API to retrieve reserved instance purchase recommendations. Analyze potential savings, utilization, coverage, and optimal commitment amounts for specific Azure resource types.
+---
+
+**Key Features:**
+- Up to 72% savings vs pay-as-you-go pricing for stable workloads
+- Resource-type specific (VMs, SQL DB, Cosmos DB, App Service, Synapse, Storage, etc.)
+- Instance size flexibility within VM series
+- Self-service returns (up to $50K/year) and exchanges within product family
+- 1-year and 3-year terms
+- Applied before savings plans in the benefit stack
```
RolandKrummenacher, line 12: Specific monetary limits like "up to $50K/year" for self-service returns tend to change over time. Consider linking to the authoritative Azure docs page for current limits, or adding a note like _"verify current limits in Azure documentation"_ so the reference doesn't silently go stale.

src/templates/agent-skills/azure-cost-management/references/Get-BenefitRecommendations.ps1
```diff
@@ -0,0 +1,76 @@
...
+    $Term = 'P3Y'
+)
+
+$url="https://management.azure.com/{0}/providers/Microsoft.CostManagement/benefitRecommendations?`$filter=properties/lookBackPeriod eq '{1}' AND properties/term eq '{2}'&`$expand=properties/usage,properties/allRecommendationDetails&api-version=2024-08-01" -f $BillingScope, $lookBackPeriod, $term
+$uri=[uri]::new($url)
+$result = Invoke-AzRestMethod -Uri $uri.AbsoluteUri -Method GET
+$jsonResult = $result.Content | ConvertFrom-Json
+
+Write-Output ""
+Write-Output "Raw output"
```
RolandKrummenacher, line 69: This script doesn't check the HTTP response status before parsing JSON — if the API returns a 4xx/5xx, `ConvertFrom-Json` will throw a confusing error. Since users will likely copy-paste this as a starting point, a basic guard would help:
```powershell
if ($result.StatusCode -ne 200) {
    Write-Error "API returned $($result.StatusCode): $($result.Content)"
    return
}
```
Also: `$lookBackPeriod` and `$term` in the format string (line 67) use lowercase, while the parameters are declared as `$LookBackPeriod` and `$Term`. PowerShell is case-insensitive so it works, but it's inconsistent with the rest of the script.

src/scripts/Update-Version.ps1
RolandKrummenacher: For additional context: every other `Out-File` call in `src/scripts/` uses `-Encoding utf8` (or `ascii`). No existing script uses `utf8NoBOM`. Switching to `-Encoding utf8` would be consistent with the rest of the codebase.

Note that `-Encoding utf8` itself has a subtle cross-version difference (writes BOM on PS 5.1, no BOM on pwsh 7+), but that's the existing pattern the repo already lives with.

src/templates/claude-plugin/.claude-plugin/plugin.json
MSBrett: Cargo cult claude recommendation - got the same one myself. It's the official MSFT MCP server - we always want the latest features.

.gitignore
MSBrett: Done

src/scripts/Update-Version.ps1
MSBrett: done

src/templates/agent-skills/.build.config
MSBrett: we need both so we can package skills separate

src/templates/agent-skills/azure-cost-management/references/azure-reservations.md
MSBrett: updated

src/templates/agent-skills/azure-cost-management/references/Get-BenefitRecommendations.ps1
MSBrett: updated

src/templates/claude-plugin/.claude-plugin/plugin.json
```diff
@@ -0,0 +1,23 @@
...
+    }
```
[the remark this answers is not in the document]
MSBrett: Pinning versions introduces security risks for zero benefit.

RolandKrummenacher (approved):

PR #2056, opened 2026-03-15
Add commitment discount eligibility open data set

RolandKrummenacher:
## Summary

Adds a new open data CSV (`CommitmentDiscountEligibility.csv`) listing Azure meters eligible for Reserved Instances and/or Savings Plans, sourced from the [Azure Retail Prices API](https://learn.microsoft.com/rest/api/cost-management/retail-prices/azure-retail-prices).

- **`src/scripts/Update-CommitmentDiscountEligibility.ps1`** — PowerShell script that pages through the Retail Prices API to collect RI-eligible meters (`priceType eq 'Reservation'`) and SP-eligible meters (Consumption items with `savingsPlan` arrays), then merges them into a single CSV.
- **`.github/workflows/opendata-commitment-eligibility.yml`** — GitHub Action that runs monthly (1st of each month) or on manual dispatch, executes the script, and creates a PR if data changed.
- **`src/open-data/CommitmentDiscountEligibility.csv`** — Initial dataset with ~84K meters.

### Motivation

FinOps Hub `Prices_v1_2()` / `Prices_v1_0()` functions currently compute commitment discount eligibility via cross-row lookups at query time, scanning the full prices table on every call. This open data set provides a pre-computed lookup that can replace those per-query scans, and also solves the per-batch ingestion issue described in #1625.

### Schema

| Column | Description |
|--------|-------------|
| `MeterId` | Azure meter GUID (region-specific) |
| `ReservationEligible` | Whether the meter has RI pricing |
| `SavingsPlanEligible` | Whether the meter has SP pricing |

### Scope

- Covers Reserved Instances and Savings Plans only
- Pre-Purchase Plans (e.g., Databricks, Sentinel SCUs) show as `Reservation` in the API and are included
- Commitment tiers (e.g., Sentinel GB/day, Log Analytics) are out of scope for now — they use a different pricing model

### Next steps

- Integrate into `Build-OpenData.ps1` to generate PowerShell functions and KQL lookup tables
- Use in Hub ingestion/view functions to replace cross-row eligibility computation

## Test plan

- [x] Full test run completed — script fetched ~142 pages of RI data and ~609 pages of Consumption data
- [x] Rate limit retry logic validated (exponential backoff with up to 5 retries)
- [x] Output CSV verified: 84,612 unique meters, no duplicates
- [x] Review GitHub Action workflow triggers and PR creation logic
- [x] Validate a sample of meter IDs against the Azure Pricing Calculator


.github/workflows/opendata-commitment-eligibility.yml
```diff
@@ -0,0 +1,64 @@
...
+          BRANCH="opendata/commitment-eligibility-$(date +%Y%m)"
+          git checkout -b "$BRANCH"
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add src/open-data/CommitmentDiscountEligibility.csv
+          git commit -m "chore: Update commitment discount eligibility data"
+          git push origin "$BRANCH"
+
+          gh pr create \
+            --title "Update commitment discount eligibility data ($(date +'%B %Y'))" \
```
flanakin:
```suggestion
            --title "[Open Data] Update commitment discount eligibility data ($(date +'%B %Y'))" \
```

.github/workflows/opendata-commitment-eligibility.yml
```diff
@@ -0,0 +1,64 @@
+name: Update Commitment Discount Eligibility
+
+on:
+  schedule:
+    - cron: '0 6 1 * *'  # 1st of each month at 06:00 UTC
```
flanakin: Any reason why 6am UTC? I'm thinking we should do this weekly.

.github/workflows/opendata-commitment-eligibility.yml
```diff
@@ -0,0 +1,64 @@
...
+  workflow_dispatch:
```
flanakin, line 6: Is this supposed to be empty?

src/scripts/Update-CommitmentDiscountEligibility.ps1
```diff
@@ -0,0 +1,144 @@
...
+# Step 3: Merge and output
+# -----------------------------------------------------------------------
+$allMeterIds = @($riMeters.Keys) + @($spMeters.Keys) | Select-Object -Unique | Sort-Object
+
+$rows = foreach ($meterId in $allMeterIds)
+{
+    [PSCustomObject]@{
+        MeterId             = $meterId
+        ReservationEligible = $riMeters.ContainsKey($meterId)
+        SavingsPlanEligible = $spMeters.ContainsKey($meterId)
```
flanakin: Should we use FOCUS-aligned names to match our dataset or the native Azure columns? I'm thinking we should use the Azure native MeterId, since that's what joins with price sheet and the FOCUS-aligned eligibility columns since that's what we bring into our data. Thoughts?
```suggestion
        MeterId             = $meterId
        x_CommitmentDiscountUsageEligibility = $riMeters.ContainsKey($meterId) ? 'Eligible' : 'Not Eligible'
        x_CommitmentDiscountSpendEligibility = $spMeters.ContainsKey($meterId) ? 'Eligible' : 'Not Eligible'
```

flanakin (approved):

.github/workflows/opendata-commitment-eligibility.yml
RolandKrummenacher: Done — added `[Open Data]` prefix.

.github/workflows/opendata-commitment-eligibility.yml
RolandKrummenacher: Changed to weekly (every Monday at 06:00 UTC).

.github/workflows/opendata-commitment-eligibility.yml
RolandKrummenacher: Yes — empty `workflow_dispatch:` is valid YAML and enables manual triggering with no input parameters.

src/scripts/Update-CommitmentDiscountEligibility.ps1
RolandKrummenacher: Done — renamed to `x_CommitmentDiscountSpendEligibility` and `x_CommitmentDiscountUsageEligibility` with "Eligible"/"Not Eligible" string values.

Note: your suggestion had RI → Usage and SP → Spend, but the existing KQL in `IngestionSetup_v1_2.kql` maps RI → Spend and SP → Usage. I matched the KQL. Worth double-checking which is correct.

.github/workflows/opendata-commitment-eligibility.yml
```diff
@@ -0,0 +1,64 @@
...
+            --title "[Open Data] Update commitment discount eligibility data ($(date +'%B %Y'))" \
+            --base dev \
+            --body "$(cat <<'EOF'
+          ## Summary
+
+          Weekly automated update of commitment discount eligibility data from the
+          [Azure Retail Prices API](https://learn.microsoft.com/rest/api/cost-management/retail-prices/azure-retail-prices).
+
+          This file lists Azure meters that are eligible for Reserved Instances and/or
+          Savings Plans, used by FinOps Hubs to populate `x_CommitmentDiscountSpendEligibility`
```
flanakin: 🤖 [AI][Claude] 🚫 **Blocker**

The PR body text in this heredoc (lines 48-64) has 10 spaces of leading whitespace on each line, matching the YAML indentation level. Bash `<<'EOF'` preserves all whitespace literally, so the generated PR body on GitHub will have every line indented with spaces, causing the markdown to render incorrectly (as a code block or with unwanted indentation).

**Fix:** Remove the leading whitespace from the heredoc content so the markdown lines start at column 1. Note that the `EOF` terminator must also start at column 1 (or use `<<-'EOF'` with tab indentation to allow stripping).

.github/workflows/opendata-commitment-eligibility.yml
```diff
@@ -0,0 +1,64 @@
...
+            exit 0
+          fi
+
+          BRANCH="opendata/commitment-eligibility-$(date +%Y%m)"
+          git checkout -b "$BRANCH"
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add src/open-data/CommitmentDiscountEligibility.csv
+          git commit -m "chore: Update commitment discount eligibility data"
+          git push origin "$BRANCH"
```
flanakin, line 58: 🤖 [AI][Claude] 💡 **Suggestion**

The branch name `opendata/commitment-eligibility-$(date +%Y%m)` is deterministic by month. If the schedule changes to weekly or if a manual dispatch overlaps with a scheduled run, this could create conflicts.

Consider either:
1. Including the day in the branch name (e.g., `$(date +%Y%m%d)`) if switching to more frequent runs
2. Checking for an existing branch/PR and updating it instead of creating a new one — this would be cleaner since it avoids orphaned branches and keeps the conversation in one PR

src/scripts/Update-CommitmentDiscountEligibility.ps1
```diff
@@ -0,0 +1,144 @@
...
+            {
+                $retries++
+                if ($retries -gt $maxRetries)
+                {
+                    throw "Failed after $maxRetries retries on page $page`: $_"
+                }
+                $wait = [Math]::Pow(2, $retries) * 10  # 20s, 40s, 80s, 160s, 320s
+                Write-Host "  Rate limited on page $page, retrying in ${wait}s (attempt $retries/$maxRetries)"
+                Start-Sleep -Seconds $wait
+            }
```
flanakin, line 84: 🤖 [AI][Claude] ⚠️ **Should fix**

The retry logic (lines 58-76) catches all exceptions and retries uniformly, but different HTTP errors warrant different handling:

- **429 (Too Many Requests):** Should retry, ideally respecting the `Retry-After` response header
- **5xx (Server errors):** Transient, safe to retry
- **4xx (Client errors like 400, 404):** Permanent failures — retrying won't help

Currently the log message says "Rate limited" for all errors, which could be misleading if the real issue is a 404 or 500.

**Suggestion:** Check `$_.Exception.Response.StatusCode` to differentiate, and parse `Retry-After` when available. Something like:

```powershell
$statusCode = [int]$_.Exception.Response.StatusCode
if ($statusCode -and $statusCode -lt 500 -and $statusCode -ne 429)
{
    throw "HTTP $statusCode on page $page: $_"
}
```

flanakin (commented):
🤖 [AI][Claude] **PR Review**

**Summary:** Solid addition of a pre-computed commitment discount eligibility dataset that will improve Hub query performance. The script and workflow are well-structured, but there are a few issues to address around workflow formatting, error handling, documentation, and scalability.

### 🚫 Blockers (1)
1. Workflow PR body will have leading whitespace due to heredoc indentation

### ⚠️ Should fix (2)
1. Retry logic doesn't differentiate error types or respect `Retry-After` headers
2. `src/open-data/README.md` not updated with the new dataset — please add a section for CommitmentDiscountEligibility following the same format as the existing datasets (PricingUnits, Regions, ResourceTypes, Services)

### 💡 Suggestions (2)
1. Workflow should handle existing branches/PRs to avoid conflicts on repeated runs
2. `Build-OpenData.ps1` auto-generates PowerShell functions from all CSVs in `src/open-data/`. With 84K rows, the generated file will be ~4MB. See #2069 for tracking this concern.

flanakin (changes requested):

.github/workflows/opendata-commitment-eligibility.yml
RolandKrummenacher: Fixed — moved the PR body into a YAML block scalar env var (`PR_BODY`) so indentation is handled cleanly. Also included the day in the branch name (`%Y%m%d`) for weekly runs.

.github/workflows/opendata-commitment-eligibility.yml
RolandKrummenacher: Fixed — branch name now uses `$(date +%Y%m%d)` to avoid conflicts with weekly runs.

src/scripts/Update-CommitmentDiscountEligibility.ps1
RolandKrummenacher: Fixed — retry logic now:
- Fails fast on 4xx errors (except 429)
- Respects `Retry-After` header when present
- Shows accurate error reason in log messages (e.g., "Rate limited" vs "HTTP 500")

flanakin (approved):

Issue #2057, opened 2026-03-16
Blank manifest.json does not trigger ingestion due to ignoreEmptyBlobs


RolandKrummenacher:
## Summary

The [custom ingestion documentation](https://learn.microsoft.com/en-us/cloud-computing/finops/toolkit/hubs/deploy?tabs=azure-portal%2Cadx-dashboard#ingest-from-other-data-sources) instructs users to create a blank `manifest.json` file to trigger ingestion. However, the ADF blob event trigger is configured with `ignoreEmptyBlobs: true`, which silently drops events for zero-byte blobs. This means a blank manifest.json never triggers the ingestion pipeline.

## Root cause

The trigger in [`hub-eventTrigger.bicep:79`](https://github.com/microsoft/finops-toolkit/blob/dev/src/templates/finops-hub/modules/fx/hub-eventTrigger.bicep#L79) has:

```bicep
ignoreEmptyBlobs: true
```

This setting has been present since the trigger module was first created in #100 (April 2023), when it was watching for `.csv` files. It was carried forward when the trigger was refactored to watch for `manifest.json`, but the custom ingestion docs were written assuming a blank file would work.

## Impact

Users following the docs for custom/multi-cloud ingestion (e.g., AWS, GCP) create an empty `manifest.json` and the pipeline never runs. There is no error — the event is silently ignored.

## Workaround

Write non-empty content to the manifest file, e.g.:

```json
{
  "generatedUtc": "<timestamp>"
}
```

The content must change on each write to generate a new `BlobCreated` event.

Credit to @DMarshall-1 for discovering this workaround in #2046.

## Suggested fix

Either:
1. **Update the docs** to specify that `manifest.json` must contain non-empty content, and provide a template
2. **Change `ignoreEmptyBlobs` to `false`** in the trigger — though this may introduce noise from empty blob artifacts
3. **Both**: keep `ignoreEmptyBlobs: true` for safety, update the docs, and provide a sample pipeline snippet that writes the manifest with a timestamp

## Related

- #2046 — Original issue where this was discovered

RolandKrummenacher:
Option 1 (doc fix) is covered in PR #2098, which rewrites the **Ingest from other data sources** section of `deploy.md` to explicitly require non-empty `manifest.json` content and calls out the `ignoreEmptyBlobs: true` behavior.

[...]

Issue #2046, opened 2026-03-04
FinOps Hub v13 + Power BI (Storage report) do not show AWS costs when parquet file is dropped directly under ingestion/


BoubakriMohamed404:
**Description:**

With FinOps Hub v13, AWS cost data does not appear in Power BI Storage‑based reports when the AWS parquet file (exported as FOCUS 1.0) is manually copied to the hub storage path:
ingestion/costs/\<year>/\<month>/awstest/

Azure costs render correctly. AWS costs ignored.

**Expected behavior:**

Storage‑based Power BI reports should display AWS costs  also 

**Actual behavior:**

Juste Azure costs are loaded successfully .

**Steps to reproduce**

1. Deploy FinOps Hub v13.
2. Export AWS costs as FOCUS 1.0 parquet (outside the hub pipelines).
3. Manually copy the parquet to ingestion/costs/\<year>/\<month>/awstest/ focusfile and add an empty manifest.json file
4. Open Power BI Storage reports and refresh.
5. Observe AWS data are not loaded .

**Environment**

FinOps Hub: v13
Power BI: FinOps toolkit Storage‑based reports
Data sources:
Azure data lake gen 2 → hub (OK)
with AWS parquet (FOCUS 1.0) manually placed in ingestion/costs/\<year>/\<month>/awstest/

**Doc link**
How to create and update FinOps hubs - Cloud Computing | Microsoft Learn
https://learn.microsoft.com/en-us/cloud-computing/finops/toolkit/hubs/deploy?tabs=azure-portal%2Cadx-dashboard#ingest-from-other-data-sources

Can you please provide me with the correcte steps to use FinOpsHub with AWS data in PowerBI storage report ?

flanakin:
@BoubakriMohamed404 The one thing that's jumping out to me is that the `Costs` folder should have an uppercase "C". I don't recall if that's a hard requirement for storage reports, but I know it is for Data Explorer ingestion. At first glance, everything else seems to be correct.

I would recommend renaming the folder to `Costs` (uppercase "C") and seeing if that works. As a quick workaround, you can also edit this line in the `ftk_Storage` function to change the casing of the folder name we're checking for:

```m
or (data = "focuscost"                  and "Costs"                   = Text.Split([Folder Path],"/"){containerOffset+1})
```

[...]

PR #2098, opened 2026-04-16
Docs: document filename convention and pipeline constraints for custom ingestion

### LINKED ISSUE #2096: Docs: 'Ingest from other data sources' missing critical filename/dedup details
## Summary

The [Ingest from other data sources](https://learn.microsoft.com/en-us/cloud-computing/finops/toolkit/hubs/deploy#ingest-from-other-data-sources) section of the deploy tutorial documents the folder path convention but omits several details that are required for the Hub's ingestion pipeline to work correctly. Following the doc literally led to silent data loss in our GCP → FOCUS → Hub pipeline.

Gaps we hit while integrating a GCP Cloud Billing FOCUS view into the Hub:

### 1. Filename convention is not documented (critical)

The doc says nothing about filenames, but the ingestion pipeline splits each filename on `__` (double underscore) to derive `ingestionId` and `originalFileName`. Source: [`Analytics/app.bicep:1806-1811`](https://github.com/microsoft/finops-toolkit/blob/dev/src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/app.bicep#L1806-L1811).

The pre-ingest cleanup ([`Analytics/app.bicep:1352`](https://github.com/microsoft/finops-toolkit/blob/dev/src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/app.bicep#L1352)) drops all extents tagged with the same \`drop-by:<folderPath>\` that don't share the current run's \`ingestionId\`. If you upload files without the \`__\` separator, every file gets its own derived \`ingestionId\` (the full filename), so each file's ingestion drops all previously-ingested files in the same folder — leaving only whatever got ingested last.

Concrete impact: we uploaded 12 Parquet files per month; 10 rows out of ~2100 survived in \`Costs_final_v1_2\` per run.

**Suggested doc fix:** Add a subsection:
- Filename pattern: \`<ingestionId>__<originalFileName>.parquet\`
- \`ingestionId\` must be stable across all files in a single logical run (e.g. a timestamp + GUID)
- Separator is two underscores

### 2. Dedup requires full-month replacement, not deltas

The doc hints at this with \"overriding datasets should consistently land in the same folder path so they're overwritten each time,\" but it's easy to miss. The implication isn't stated: if you write only incremental/delta rows to the month folder, the pre-ingest cleanup wipes the rest of the month's extents, and Kusto ends up with just the delta.

**Suggested doc fix:** Add an explicit note: \"When new rows arrive for an existing month, you must re-write the entire month's data in that folder (with a new ingestionId). Incremental deltas are not supported — they will result in data loss.\"

### 3. Tables referenced in step 4 are outdated

Step 4 says \"update the **Costs_raw** and **Costs_final_v1_0** tables, and **Costs_transform_v1_0**, **Costs_v1_0**, and **Costs** functions\". Current Hub deployments use the \`v1_2\` variants (\`Costs_final_v1_2\`, \`Costs_transform_v1_2\`, \`Costs_v1_2\`). The doc wasn't updated when the schema was bumped.

### 4. Empty Parquet shards cause 8-minute ingestion stalls

Not a documentation gap per se, but worth calling out: tools that produce partitioned Parquet output (BigQuery \`EXPORT DATA\`, Spark \`write.parquet\`, etc.) commonly emit header-only empty shards (~8 KB on a typical FOCUS schema). Kusto rejects these with \`BadRequest_NoRecordsOrWrongFormat: 0 bytes\`, and each failure is retried 3× at 120s intervals (\~8 minutes wasted per empty file). With \~50% empty shards a typical run's Hub pipeline takes 1–2 hours instead of minutes.

**Suggested doc fix:** Add a tip: \"Filter out empty shards (below \~10 KB on a typical FOCUS schema) before uploading to the ingestion container to avoid retry storms in the Hub pipeline.\"

### 5. Manifest content specification is ambiguous

The doc says \"Create an empty \`manifest.json\` file.\" The pipeline actually accepts any valid JSON. The \`IngestionQueries\` pipeline writes a \`settings.json\`-derived payload, so \"empty\" is a convention, not a requirement. Worth clarifying so integrators know they can embed audit metadata (e.g. \`ingestionId\`, exported timestamp).

### 6. Stale source-blob cleanup is the integrator's responsibility

When integrators use the \`<ingestionId>__<originalFileName>.parquet\` convention with a fresh \`ingestionId\` per run (recommended by point 1), each run's Parquet files have different filenames than previous runs. The Hub pipeline's drop-by cleanup operates on **Kusto extents** but **does not delete source blobs** in the ingestion container.

Consequence: stale Parquets from previous runs accumulate in the target folder. On the next manifest trigger, the Hub lists **all** Parquets in the folder (old + new), ingests every one, then uses drop-by-cleanup to keep only one \`ingestionId\`'s extents. Which ingestionId wins depends on blob processing order — often the older (lexicographically earlier) one, silently reverting the data to the prior run.

Concrete impact we observed: after changing the FOCUS view's \`BilledCost\` semantics, triggering a full re-export, and confirming the new Parquets were uploaded, \`Costs_final_v1_2\` still showed the old values. The Hub had picked up both old-ingestionId and new-ingestionId files, and the older-timestamped run's extents won.

**Suggested doc fix:** Add a step: \"Before uploading a new run's Parquets, delete all existing Parquets in the target folder that don't share the current run's ingestionId. The Hub pipeline does not clean up source blobs — leftover files from prior runs can race with new ones and silently revert your data.\"

Alternatively, integrators could use fixed filenames (\`part-0.parquet\`, \`part-1.parquet\`, …) that overwrite in place, matching the Azure Cost Management export pattern. This works only if the shard count is stable across runs; otherwise leftover shards cause the same issue.

## Repro context

GCP Cloud Billing detailed export → BigQuery FOCUS view → Parquet exports to GCS → streamed to Azure ingestion container → Hub pipeline → Kusto \`Costs_final_v1_2\`. Verified against \`Hub\` database on a FinCops-hosted Fabric eventhouse.

RolandKrummenacher:
## Summary

Closes #2096. Expands the **Ingest from other data sources** section of the Hub deploy tutorial with details required for the Data Explorer ingestion pipeline to work correctly. The existing text led integrators into silent data loss when uploading custom FOCUS datasets (e.g. from GCP or AWS).

Changes to [`docs-mslearn/toolkit/hubs/deploy.md`](/microsoft/finops-toolkit/blob/krummrol/docs-ingest-from-other-sources/docs-mslearn/toolkit/hubs/deploy.md):

- Document the `<ingestionId>__<originalFileName>.parquet` filename convention and explain how the pipeline derives `ingestionId` by splitting on `__` (see [`Analytics/app.bicep:1806-1810`](/microsoft/finops-toolkit/blob/dev/src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/app.bicep#L1806-L1810)).
- Add an IMPORTANT callout that each upload must rewrite the full month — incremental deltas silently wipe existing extents because of the pre-ingest cleanup at [`Analytics/app.bicep:1352`](/microsoft/finops-toolkit/blob/dev/src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/app.bicep#L1352).
- Add a TIP to filter empty parquet shards before upload (Data Explorer rejects them with `BadRequest_NoRecordsOrWrongFormat` and the pipeline retries 3× at 120s intervals, per `retry: 3, retryIntervalInSeconds: 120` on the Ingest Data activity).
- Correct the manifest instruction: the file must contain at minimum `{}` because the storage event trigger sets `ignoreEmptyBlobs: true` ([`fx/hub-eventTrigger.bicep:79`](/microsoft/finops-toolkit/blob/dev/src/templates/finops-hub/modules/fx/hub-eventTrigger.bicep#L79)) — a zero-byte file never triggers ingestion.
- Update table/function references in step 4 from `v1_0` to the current `v1_2` schema (confirmed against `HubSetup_Latest.kql`).

No code changes.

## Test plan

- [ ] Preview the rendered page locally (`docs-mslearn`) and confirm the IMPORTANT/TIP callouts render correctly under the numbered list item.
- [ ] Verify no broken links in the updated section via the existing doc-link test suite.

### DIFF
--- a/docs-mslearn/toolkit/changelog.md
+++ b/docs-mslearn/toolkit/changelog.md
@@ -56,6 +56,7 @@ _Released April 2026_
   - Improved deployment UI to consolidate hub mode selection into a single radio button group with four mutually exclusive options: None (storage only for Power BI reports), Azure Data Explorer, Microsoft Fabric, or Remote Hub ([#1929](https://github.com/microsoft/finops-toolkit/issues/1929)).
   - Remote Hub configuration (storage URI, storage key, and purge protection) is now displayed in the Basics tab when Remote Hub mode is selected, making the mutual exclusivity clear.
   - Data Explorer SKU and retention settings are now only visible when Azure Data Explorer mode is selected.
+  - Documented the `<ingestionId>__<originalFileName>.parquet` filename convention, full-folder replacement requirement, and manifest content rules for ingesting custom FOCUS datasets to prevent silent data loss ([#2096](https://github.com/microsoft/finops-toolkit/issues/2096)).
 - **Fixed**
   - Fixed Init-DataFactory deployment script failing when an Event Grid subscription is already provisioning by checking subscription status before attempting subscribe/unsubscribe and polling separately for completion ([#1996](https://github.com/microsoft/finops-toolkit/issues/1996)).
   - Added row count check in `msexports_ExecuteETL` pipeline to fix error when export files have no rows ([#1535](https://github.com/microsoft/finops-toolkit/issues/1535)).
--- a/docs-mslearn/toolkit/changelog.md
+++ b/docs-mslearn/toolkit/changelog.md
@@ -3,7 +3,7 @@ title: FinOps toolkit changelog
 description: Review the latest features and enhancements in the FinOps toolkit, including updates to FinOps hubs, Power BI reports, and more.
 author: MSBrett
 ms.author: brettwil
-ms.date: 04/04/2026
+ms.date: 04/21/2026
 ms.topic: reference
 ms.service: finops
 ms.subservice: finops-toolkit
--- a/docs-mslearn/toolkit/hubs/deploy.md
+++ b/docs-mslearn/toolkit/hubs/deploy.md
@@ -360,15 +360,13 @@ To ingest data from other data providers that support FOCUS, such as Amazon Web 
      - The pipeline derives the ingestion ID by splitting the file name on `__`. Every file that shares the same `ingestionId` is treated as part of the same run; extents from previous runs in the same folder are dropped once the new run succeeds.
    <!-- prettier-ignore-start -->
    > [!IMPORTANT]
-   > **Each run must rewrite the full contents of its folder path**
+   > **Each run must replace the full contents of its folder path**
    >
-   > The pre-ingest cleanup drops every extent tagged with the current folder path that doesn't share the run's `ingestionId`. Uploading only incremental rows into a folder that was already ingested results in silent data loss — only the new rows survive in Data Explorer.
+   > When re-ingesting into a folder that already contains data, you must both (1) include the complete dataset for the period in the new run (not just incremental rows), and (2) delete all `.parquet` files from previous runs before uploading the new ones. Otherwise, the pipeline processes old and new files together and each file's pre-ingest cleanup drops extents tagged with a different `ingestionId` — only the files sharing the last `ingestionId` processed survive in Data Explorer.
    >
    > If a provider emits nonoverlapping deltas, give each delta its own folder path using a `dd` or `dd/hh` subfolder as described above. Each folder is then independently replaced on its next run instead of being merged.
    >
-   > **Remove parquet files from previous runs before each upload**
-   >
-   > Before uploading a new run's files, delete every `.parquet` file already in the target folder. If files with a different `ingestionId` remain, the pipeline processes them alongside the new ones and each file's pre-ingest cleanup drops extents tagged with a different `ingestionId` — only the files sharing the last `ingestionId` processed survive in Data Explorer. The managed Cost Management export pipeline handles this cleanup automatically, but custom ingestion workflows must implement it themselves.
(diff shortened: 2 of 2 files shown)


RolandKrummenacher:
Cross-references for reviewer context:

- #2057 — Existing open issue reporting that a blank `manifest.json` never triggers ingestion due to `ignoreEmptyBlobs: true`. This PR addresses Option 1 from that issue (doc fix). Options 2 and 3 (changing the trigger setting) remain open for a separate code change.
- #2046 — Upstream user report where the integrator copied AWS parquet under `ingestion/costs/<year>/<month>/<scope>/` with a blank `manifest.json` and saw nothing ingested. The failure modes in that issue are covered by the filename-convention and manifest-content changes in this PR, plus the full-folder-replacement callout.

Closing #2096 (this PR's parent issue); leaving #2057 and #2046 for maintainers to decide whether an additional code-side fix is wanted.

flanakin (approved):

PR #1876, opened 2025-11-03
[Hubs] Enable custom resource naming

lcarli:
🛠️ Description
This PR adds optional custom resource naming and centralized networking support for FinOps hub deployments, enabling better integration with enterprise naming conventions and Hub & Spoke network topologies.

Key Features:
1. Custom Resource Naming (Optional)

Added 7 optional parameters to customize resource names:
storageAccountName - Storage Account
dataFactoryName - Data Factory
keyVaultName - Key Vault
virtualNetworkName - Virtual Network
managedIdentityName - Managed Identity
dataExplorerClusterName - Data Explorer Cluster
privateEndpointNamePrefix - Private Endpoint prefix
If parameters are empty, default names are auto-generated following Azure best practices
Simplified implementation: removed boolean flag complexity, using simple !empty() checks

2. Centralized Networking (Hub & Spoke)

Added 6 optional parameters to reference existing Private DNS Zones:
existingBlobDnsZoneId - Blob Storage DNS Zone
existingDfsDnsZoneId - Data Lake Storage DNS Zone
existingQueueDnsZoneId - Queue Storage DNS Zone
existingTableDnsZoneId - Table Storage DNS Zone
existingVaultDnsZoneId - Key Vault DNS Zone
existingDataExplorerDnsZoneId - Data Explorer DNS Zone
Prevents duplicate DNS Zone creation in centralized network environments
Automatically links Private Endpoints to existing DNS Zones via DNS Zone Groups

Changes:
Updated all Bicep modules with conditional logic for custom naming
Added new "Customization" tab in [createUiDefinition.json](vscode-file://vscode-app/c:/Users/lramoscostah/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) with organized sections
Updated README with comprehensive customization documentation
Added test cases for custom naming and centralized networking scenarios
Compiled ARM template (azuredeploy.json) for portal deployment
Updated metadata with current date and enhanced description
Fixes # 

📷 Screenshots
New "Customization" Tab in Azure Portal:
The new tab appears between "Advanced" and "Tags", providing a clean interface for optional customizations without cluttering the basic deployment experience.

Section 1: Resource naming - 7 optional name fields with validation
Section 2: Centralized networking (Hub & Spoke) - 6 DNS Zone resource ID fields

image

📋 Checklist

🔬 How did you test this change?

     💪 Unit tests

🙋‍♀️ Do any of the following that apply?
 
[...]

Issue #1105, opened 2024-11-07
[BICEP] FinOps Toolkit - Custom resource naming


plaplainabf:
## ⚠️ Problem
The bicep deployment doesn't allow the names of the resources to be set. This means that companies with naming standards can't deploy FinOps using bicep or the accept the resources won't adhere to their naming standards.

## 🛠️ Solution
Update the bicep code to include parameters for the names of the keyvault, data factory and storage accounts, this means that consumers can generate names using their own bicep naming modules and pass it through to the FinOps Toolkit deployment, providing a repeatable deployment.

## ℹ️ Additional context
Looking at the FinOps Toolkit bicep code the main.bicep should be fairly easy to update for the data factory and storage account as there are two variables that are set.
![Image](https://github.com/user-attachments/assets/862f0698-f841-4f44-9765-3d0aeb4b3437)

The keyvault name appears to be set within the keyvault.bicep so that would need a parameter defining and the main.bicep updating to pass through said parameter.

I believe parameters can have default values set, so these could be set to what the bicep deployment currently uses, that way it wouldn't break existing deployments.

This was originally posted in the wrong place;
https://github.com/Azure/bicep-registry-modules/issues/3740#issuecomment-2461047695

## 🙋‍♀️ Ask for the community
We could use your help:
1. Please vote this issue up (👍) to prioritize it.

MSBrett:
Thanks everyone for pushing on this — closing the loop here since three separate PRs (#1635, #2148, #1876) have proposed overlapping designs for this ask.

**Decision: rejecting free-form custom naming for globally-unique / DNS-namespaced resources** (Storage Account, Key Vault, Data Factory, and the Azure Data Explorer cluster).

Why: we already ship optional custom naming for the ADX cluster (`dataExplorerName` in `main.bicep`), and it takes the caller's value with zero uniqueness suffix. In production, this is our #1 cause of deployment failures — customers collide on cluster/storage/KV names across subscriptions, regions, or repeat deployments, because nothing enforces global DNS uniqueness once a user supplies their own name. This isn't hypothetical risk; it happened, repeatedly, with the one resource we already allow it for.

All three open PRs would extend that same pattern to more resources:
- #1635 and #1876 both **replace** (not append to) the uniqueness suffix the moment a custom name is supplied.
- #2148 drops the suffix **unconditionally**, even without opt-in — already flagged as a blocking bug in review.

[...]

PR #2148, opened 2026-05-16
Add BYO-VNet support and standardized resource naming for FinOps hubs

fatimataayeb:
# Add BYO-VNet support and standardized resource naming for FinOps hubs

## Summary

Enables FinOps hub deployments into **existing customer VNets** (hub-and-spoke scenarios) and applies a **standardized, deterministic resource naming convention** that aligns with the [Cloud Adoption Framework (CAF)](https://learn.microsoft.com/azure/cloud-adoption-framework/ready/azure-best-practices/resource-naming) and customer governance policies.

## Motivation

Today the FinOps hub template:
1. Always creates a brand-new VNet (`vnet-{hubName}`) and subnets — incompatible with hub-and-spoke topologies where private connectivity is centralized in a hub VNet owned by the network team
2. Uses non-deterministic / hash-suffixed resource names (e.g. `finopshub-c3orenemzlnm6`) — fails customer naming standards, breaks Azure Policy naming rules, and makes resources hard to identify in shared subscriptions

These two limitations have blocked enterprise deployments at multiple regulated customers (banking, public sector) where:
- A **hub-and-spoke network architecture** is mandated and private endpoints must land in pre-existing subnets
- A **resource naming policy** is enforced via Azure Policy and deployments are denied if names don't match

> Note:  These changes may break future updates, do it at your own risk 

### 1. BYO-VNet support

New optional parameters on `main.bicep`:

| Parameter | Description | Default |
|---|---|---|
| `existingVNetName` | Name of the customer's existing VNet | `''` (creates new VNet) |
| `existingVNetResourceGroupName` | RG of the existing VNet (supports cross-RG) | `''` |
| `peSubnetName` | Subnet for private endpoints | `snet-finops-pe-01` |
| `scriptSubnetName` | Subnet for deployment scripts (ACI) | `snet-finops-script-01` |
| `dataExplorerSubnetName` | Subnet for ADX (if used) | `snet-finops-adx-01` |

When `existingVNetResourceGroupName` is provided (or `networkName` does not start with `vnet-finops-`), the template **skips** creating NSG/VNet/subnets and uses cross-RG `resourceId(...)` references. All private endpoints and ACI containers land in the customer's existing subnets.

### 2. CAF-aligned resource naming

All resource names follow `{abbreviation}-finops-{hubName}` (storage follows `stgfinops{hubName}{instance}`):

| Resource | Old | New |
|---|---|---|
| Data Factory | `finopshub-<hash>` | `adf-finops-{hubName}` |
| Key Vault | `finopshubkv<hash>` | `kv-finops-{hubName}` |
| Storage (data lake) | `finopshubstg<hash>` | `stgfinops{hubName}01` |
| Storage (script) | `finopshubstg<hash>script` | `stgfinops{hubName}02` |
| NSG | `finopshub-nsg-<hash>` | `nsg-finops-{hubName}` |
| VNet | `finopshub-vnet-<hash>` | `vnet-finops-{hubName}` |
| Managed Identity | `finopshub-mi-<hash>` | `id-finops-{hubName}` |
| Event Grid System Topic | (implicit on trigger start) | `evg-finops-{hubName}` |
| PE (blob) | `finopshub-pep-<hash>` | `pep-finops-stgblob-{hubName}` |
| PE (DFS) | `finopshub-pep-<hash>` | `pep-finops-stgdfs-{hubName}` |
| PE (KV) | `finopshub-pep-<hash>` | `pep-finops-kv-{hubName}` |

All names respect per-resource character limits via `take()` in `hub-types.bicep`.

### 3. Explicit Event Grid system topic

`evg-finops-{hubName}` is now declared in Bicep, sourced from the hub data lake. Previously this was implicitly created when the storage event trigger was activated, making RBAC/tagging policies harder to apply. Explicit declaration gives customers IaC control.

## Files changed (5)

- `src/templates/finops-hub/main.bicep` — new BYO-VNet parameters
- `src/templates/finops-hub/modules/hub.bicep` — forwards new parameters
- `src/templates/finops-hub/modules/fx/hub-types.bicep` — `HubRoutingProperties` extended; `newHub`/`newApp` apply naming
- `src/templates/finops-hub/modules/fx/hub-app.bicep` — PE naming + Event Grid topic
- `src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Core/infrastructure.bicep` — BYO-VNet aware (gates resource creation, cross-RG references)

## Benefits

| Persona | Benefit |
|---|---|
| Enterprise architect | Deploy into existing hub-and-spoke topology without exception requests |
| Network admin | No new VNets, no peering, no IP planning — PEs land in pre-approved subnets |
| Cloud governance | Resource names pass Azure Policy naming rules out of the box |
| FinOps practitioner | Resources immediately identifiable in shared subscriptions (no random suffixes) |
| Toolkit maintainer | Backwards compatible — all new parameters optional with sensible defaults |

## Backwards compatibility

- All new parameters are **optional** with defaults that reproduce current behavior
- When BYO-VNet parameters are not specified, the template creates a VNet exactly as today
- Existing deployments would need resource recreation to pick up the new naming (Azure does not support most resource renames)

## Testing

- ✅ Bicep compiles cleanly (`az bicep build`)
- ✅ ARM template validation passes (`az deployment group validate`) for both BYO-VNet and create-new-VNet scenarios
- ✅ End-to-end deployed at a banking customer in `qatarcentral` into pre-existing hub-and-spoke subnets, using customer-mandated naming convention
- ✅ FOCUS Cost ingestion validated end-to-end: exports → msexports → ETL pipelines → ingestion container → Power BI

### DIFF
--- a/src/templates/finops-hub/main.bicep
+++ b/src/templates/finops-hub/main.bicep
@@ -162,6 +162,21 @@ param enablePublicAccess bool = true
 @description('Optional. Address space for the workload. Minimum /26 subnet size is required for the workload. Default: "10.20.30.0/26".')
 param virtualNetworkAddressPrefix string = '10.20.30.0/26'
 
+@description('Optional. Name of an existing VNet to deploy private endpoints into (BYO-VNet for hub-and-spoke). Leave empty to create a new VNet. Default: "".')
+param existingVNetName string = ''
+
+@description('Optional. Resource group of the existing VNet. Leave empty if same as deployment RG. Default: "".')
+param existingVNetResourceGroupName string = ''
+
+@description('Optional. Subnet name for private endpoints. Default: "snet-finops-pe-01".')
+param peSubnetName string = 'snet-finops-pe-01'
+
+@description('Optional. Subnet name for deployment scripts (delegated to Microsoft.ContainerInstance/containerGroups). Default: "snet-finops-script-01".')
+param scriptSubnetName string = 'snet-finops-script-01'
+
+@description('Optional. Subnet name for Azure Data Explorer. Default: "snet-finops-adx-01".')
+param dataExplorerSubnetName string = 'snet-finops-adx-01'
+
 
 //==============================================================================
 // Resources
@@ -196,6 +211,11 @@ module hub 'modules/hub.bicep' = {
     remoteHubStorageKey: remoteHubStorageKey
     enablePublicAccess: enablePublicAccess
     virtualNetworkAddressPrefix: virtualNetworkAddressPrefix
+    existingVNetName: existingVNetName
+    existingVNetResourceGroupName: existingVNetResourceGroupName
+    peSubnetName: peSubnetName
+    scriptSubnetName: scriptSubnetName
+    dataExplorerSubnetName: dataExplorerSubnetName
   }
 }
 
--- a/src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Core/infrastructure.bicep
+++ b/src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Core/infrastructure.bicep
@@ -16,14 +16,19 @@ param hub HubProperties
 // Variables
 //==============================================================================
 
-var nsgName = '${hub.routing.networkName}-nsg'
+var nsgName = 'nsg-finops-${hub.name}'
 
-// Workaround https://github.com/Azure/bicep/issues/1853
-var finopsHubSubnetName = 'private-endpoint-subnet'
-var scriptSubnetName = 'script-subnet'
-var dataExplorerSubnetName = 'dataExplorer-subnet'
+// True when the user supplies their own VNet (BYO / hub-and-spoke).
+// In BYO mode subnets must already exist in the existing VNet and we skip creating NSG/VNet here.
+var bringYourOwnNetwork = !empty(hub.routing.existingVNetResourceGroupName) || (!empty(hub.routing.networkName) && !startsWith(hub.routing.networkName, 'vnet-finops-'))
+var createNetwork = hub.options.privateRouting && !bringYourOwnNetwork
 
-var subnets = !hub.options.privateRouting ? [] : [
+// Use the subnet names from hub.routing so they always match what was passed to newHub.
+var finopsHubSubnetName = hub.routing.peSubnetName
+var scriptSubnetName = hub.routing.scriptSubnetName
+var dataExplorerSubnetName = hub.routing.dataExplorerSubnetName
+
+var subnets = !createNetwork ? [] : [
   {
     name: finopsHubSubnetName
     properties: {
@@ -80,7 +85,7 @@ var subnets = !hub.options.privateRouting ? [] : [
 // Network
 //------------------------------------------------------------------------------
 
-resource nsg 'Microsoft.Network/networkSecurityGroups@2023-11-01' = if (hub.options.privateRouting) {
+resource nsg 'Microsoft.Network/networkSecurityGroups@2023-11-01' = if (createNetwork) {
   name: nsgName
   location: hub.location
   tags: getHubTags(hub, 'Microsoft.Storage/networkSecurityGroups')
@@ -168,7 +173,7 @@ resource nsg 'Microsoft.Network/networkSecurityGroups@2023-11-01' = if (hub.opti
   }
 }
 
-resource vNet 'Microsoft.Network/virtualNetworks@2023-11-01' = if (hub.options.privateRouting) {
+resource vNet 'Microsoft.Network/virtualNetworks@2023-11-01' = if (createNetwork) {
   name: hub.routing.networkName
   location: hub.location
   tags: getHubTags(hub, 'Microsoft.Storage/virtualNetworks')
@@ -295,7 +300,7 @@ resource tablePrivateDnsZone 'Microsoft.Network/privateDnsZones@2024-06-01' = if
 resource scriptStorageAccount 'Microsoft.Storage/storageAccounts@2022-09-01' = if (hub.options.privateRouting) {
   name: hub.routing.scriptStorage
   dependsOn: [
-    vNet::scriptSubnet
+    vNet
   ]
   location: hub.location
   sku: {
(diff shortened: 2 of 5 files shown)


fatimataayeb:
@microsoft-github-policy-service agree

fajterini:
Wow, this is great idea for Enterprise deployment. But for us we use different naming standard like suggested. I would prefer if the resource naming standard would be possible to configure in parameter file. Options default and custom naming.

But I suspect the naming change will make this not compatible with deployment with older versions? In this case best would be to create it that way that the new naming would be optional. As a new option for new deployments.

I am just planning our PoC deployment, so take this as a light suggestion :)

RolandKrummenacher (changes requested):
## Code review — BYO-VNet support + standardized resource naming

**Overview:** Bundles two features: (1) deploy private endpoints into a customer's existing VNet (cross-RG hub-and-spoke), and (2) replace hash-suffixed resource names with CAF-style deterministic names. The BYO-VNet capability is genuinely valuable, but as written there are several correctness bugs plus a breaking change I'd block on.

### 🔴 Blocking

**1. Unconditional rename breaks every existing deployment — including data loss.** The naming change isn't gated behind any flag; it applies to all deployments. Storage accounts, Key Vault, and Data Factory **cannot be renamed in place** — on upgrade Azure creates new resources and orphans the old ones, and the data lake storage account *is the hub's data*. The PR's own note ("Existing deployments would need resource recreation"; "may break future updates, do it at your own risk") confirms this. New optional params don't make the default path compatible. The renaming must be **opt-in** (e.g. `resourceNaming: 'legacy' | 'caf'` defaulting to `legacy`) or split into its own proposal.

**2. Loss of global-uniqueness suffix → name collisions.** The existing `suffix` is `uniqueString(name, resourceGroup().id)` — it exists because storage / Key Vault / Data Factory names must be **globally unique across all of Azure**. The new names drop it (`kv-finops-{hubName}`, `stgfinops{hubName}01`, `adf-finops-{hubName}`). Two customers — or one customer in two regions/RGs — using hubName `finops` now collide → `StorageAccountAlreadyTaken` / KV / ADF global-name failures. CAF naming and global uniqueness aren't mutually exclusive — keep a short uniqueness token in globally-scoped names.

**3. Private-endpoint name collision via a broken string hack.** In `infrastructure.bicep` the script-storage PE is `name: 'pep-finops-stgblob-${replace(hub.name, '-01', '-02')}'`. `hub.name` doesn't contain `-01`, so `replace(...)` is a no-op and this resolves to `pep-finops-stgblob-{hubName}` — **identical** to the data-lake `blobEndpoint` in `hub-app.bicep`. Two PEs with the same name. The instance number lives on the storage account name, not the hub name, so the differentiator was applied to the wrong variable.

### 🟠 Significant

**4. Fragile BYO detection + two inconsistent conditions.** `bringYourOwnNetwork = !empty(existingVNetResourceGroupName) || (!empty(networkName) && !startsWith(networkName, 'vnet-finops-'))`. A same-RG existing VNet that happens to be named `vnet-finops-*` defeats the heuristic and the template tries to create it. Worse, `infrastructure.bicep` decides *whether to create* off `bringYourOwnNetwork` while `hub-types.bicep` decides *cross-RG vs local resourceId* off `empty(existingVNetResourceGroupName)` — the two can disagree. Drive everything off one explicit BYO flag.

**5. Event Grid system topic: likely duplicate, possibly unwired.** `evg-finops-${app.hub.name}` is declared explicitly, but (a) existing deployments already have an implicit GUID-named topic → you end up with two, and (b) I don't see the storage event trigger rewired to bind to this topic, so ADF may still auto-create its own. Confirm trigger wiring + migration from the implicit topic.

**6. `vNetAddressSpace` output now reports the param, not reality.** Changed to `[hub.options.networkAddressPrefix]`; in BYO mode the customer VNet's real address space is unknown, so this output is now misleading.

### 🟡 Minor / scope
- **No `createUiDefinition.json` update** — portal deployments can't set BYO-VNet params (template/CLI only).
- **No changelog entry.**
- Subnet delegation (script subnet → `Microsoft.ContainerInstance/containerGroups`) can't be validated for BYO subnets — documented in the param description, which is the right call.
- **Collides with #2163** (NAT Gateway): both rewrite `infrastructure.bicep` network gating and `hub-types.bicep` routing; whichever merges second needs a non-trivial rebase, and #2163's `vnetIntegration`/`privateRouting` split interacts directly with this PR's `createNetwork`/`bringYourOwnNetwork` logic.

**Verdict: Request changes.** The feature is worth landing, but currently this would break existing hubs (rename + lost uniqueness), has a real PE name collision, and relies on a fragile heuristic. Suggested path: (1) split BYO-VNet from the renaming; (2) make naming opt-in and preserve a global-uniqueness token on storage/KV/ADF; (3) fix the `pep-finops-stgblob-*` collision (derive the instance suffix from the storage account); (4) one explicit BYO flag used consistently; (5) verify Event Grid wiring/migration and coordinate the rebase against #2163.

RolandKrummenacher:
Recusing from this review — I'm not the right person to assess the hub Bicep / networking and resource-naming changes here, so please reassign to a maintainer who owns that area.

Note: my earlier review landed as **Changes requested**, and repo branch protection won't let me dismiss my own review. Could a maintainer/admin please dismiss it so it doesn't block the merge decision? The substance of the earlier comments still stands as input for whoever picks this up — I just shouldn't be the gating reviewer.

MSBrett:
+1 to @RolandKrummenacher's review — specifically the global-uniqueness point (#2 in the list). Adding the production data point behind why that's a hard blocker rather than a nice-to-have: we already ship custom naming for the ADX cluster (`dataExplorerName`), which takes the caller's name with no uniqueness suffix at all, and it's our #1 cause of deployment failures today (name collisions across subscriptions/regions/redeployments). This PR would extend that same unprotected pattern to Storage, Key Vault, Data Factory, and networking resources — unconditionally, not even opt-in.

Also flagging: #1635 and #1876 are proposing overlapping/competing custom-naming designs for the same underlying ask (#1105), with the same uniqueness gap. Worth coordinating so we land one design, not three.

MSBrett:
Thank you for the implementation and deployment validation. We are closing this PR as not planned because the BYO-VNet design conflicts with the FinOps hub network ownership boundary.

FinOps Toolkit must exclusively create and manage the hub virtual network, its three subnets, private DNS, routing, and related resources. PR #2270 documents the supported customer-connectivity architecture: customer-managed private endpoints in the customer's own VNet are preferred, with VNet peering retained only as a secondary option.

This disposition applies to the bundled BYO-VNet design. Resource-naming changes should be evaluated separately rather than coupled to customer-owned hub networking. The existing review findings are acknowledged, and no implementation revision is requested because the product direction is not being accepted.

PR #2149, opened 2026-05-18
Fix HTML rendering in Infrastructure encryption description

RolandKrummenacher:
## Summary
- `Microsoft.Common.TextBlock` does not render inline HTML, so the `<a>` tags in the Infrastructure encryption description on the Advanced tab of the FinOps hub deployment UI appeared as raw markup.
- Split the description into two text blocks, each using the supported `options.link` (`label` + `uri`) instead of embedded HTML.

## Test plan
- [ ] Deploy the FinOps hub template via the portal and open the Advanced tab.
- [ ] Confirm the Infrastructure encryption section shows clean text with clickable "Learn more" and "Blob Storage pricing" links (no raw HTML).

### DIFF
--- a/src/templates/finops-hub/createUiDefinition.json
+++ b/src/templates/finops-hub/createUiDefinition.json
@@ -850,7 +850,23 @@
                 "type": "Microsoft.Common.TextBlock",
                 "visible": true,
                 "options": {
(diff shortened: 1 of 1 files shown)


RolandKrummenacher:
image

flanakin (approved):

PR #2154, opened 2026-05-19
fix(docs): skip ms.date updates and republish for metadata-only changes

MSBrett:
## Problem

The `update-mslearn-dates` GitHub Action and `Publish-Toolkit.ps1` both treated any change to a docs markdown file as a content change. Editing only YAML frontmatter (or other metadata-only fields) was enough to bump `ms.date` on PR merge and to produce a docs republish PR full of metadata-only churn.

## Fix

Three coordinated changes that gate updates on actual article-body changes.

### Workflow — `.github/workflows/update-mslearn-dates.yml`
- Strip YAML frontmatter from both the working copy and the `pull_request.base.sha` version of each changed file and `cmp` the remaining article body.
- Skip the `ms.date` rewrite when the bodies match.
- Skip files that no longer exist.

### Publish-Toolkit — `src/scripts/Publish-Toolkit.ps1`
- Add `Get-DocsArticleBody` — strips frontmatter, markdownlint and prettier ignore comments, normalizes trailing whitespace and blank-line runs.
- Add `Remove-DocsMetadataOnlyChanges` — walks every `docs/*.md` in the publish target and `git restore`s files whose normalized body matches `HEAD`.
- Invoked from `Set-RepoContent` when `-Template` is `'docs'`.

### Tests — `src/powershell/Tests/Unit/Action.UpdateMsLearnDates.Tests.ps1`
- Three Pester assertions pinning the workflow to `github.event.pull_request.base.sha`, the `strip_frontmatter` helper, and the `cmp -s` body-diff guard.

## Effect

- PRs that touch only frontmatter no longer bump `ms.date`.
- Docs republish branches no longer contain metadata-only file changes.
- Pester tests fail loudly if the workflow guards are removed.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


RolandKrummenacher (commented):
A couple of additional notes on top of Copilot's review:

### 1. Asymmetric "what counts as body" between workflow and publish script
The workflow's `strip_frontmatter` (awk) only removes the `---...---` block. The PowerShell `Get-DocsArticleBody` in `Publish-Toolkit.ps1` also strips:
- `<!-- markdownlint-disable-next-line MD025 -->`
- `<!-- prettier-ignore-start -->` / `<!-- prettier-ignore-end -->`
- Trailing whitespace
- Triple+ blank lines collapsed to two

If a PR changes only ignore-comments or whitespace, the workflow treats it as a body change and bumps `ms.date`, then at publish time `Remove-DocsMetadataOnlyChanges` reverts the file to HEAD (with the OLD ms.date). Net result is correct, but the ms.date bump on the PR is wasted churn — and the two definitions will drift over time.

Worth extracting body normalization to a single source of truth (either both call the same helper, or both apply the same rule set).

### 2. `Get-DocsArticleBody` / `Remove-DocsMetadataOnlyChanges` have no behavior tests
The new Pester tests follow the repo convention of regex-matching workflow YAML, which is fine for the workflow guard. But the two new PowerShell functions are pure-ish and easy to unit test:

```powershell
It 'Get-DocsArticleBody strips frontmatter'
It 'Get-DocsArticleBody strips markdownlint/prettier ignore lines'
It 'Get-DocsArticleBody normalizes blank-line runs to max two'
It 'Get-DocsArticleBody trims trailing whitespace per line'
It 'Get-DocsArticleBody returns "" for empty or frontmatter-only input'
```

These would also be the most useful guard against the asymmetry in #1 — if both sides are required to satisfy a shared test fixture, drift becomes visible.

src/scripts/Publish-Toolkit.ps1
```diff
@@ -205,6 +209,64 @@ function Set-RepoContent($repo, [string]$branchPrefix, [string]$sourceDir)
...
+        "<!-- markdownlint-disable-next-line MD025 -->",
+        "<!-- prettier-ignore-start -->",
+        "<!-- prettier-ignore-end -->"
+    )
+
+    $body = $lines[$start..($lines.Count - 1)] `
+    | Where-Object { $ignoreLines -notcontains $_.Trim() } `
+    | ForEach-Object { $_.TrimEnd() }
+
+    return (($body -join "`n") -replace "`n{3,}", "`n`n").Trim()
```
flanakin: 🤖 [AI][Claude Opus 4.7] ⚠️ **Should fix**

The workflow's `strip_frontmatter` (awk) only removes the `---...---` block, but `Get-DocsArticleBody` here additionally strips markdownlint/prettier ignore comments and normalizes trailing whitespace + blank-line runs. The two definitions disagree on "what counts as body."

A PR that changes only an ignore-comment or whitespace is seen as a real body change by the workflow (so `ms.date` gets bumped), then reverted to `HEAD` at publish time (with the OLD `ms.date`). Net result is correct today, but:

1. The PR carries a wasted `ms.date` bump that will be silently undone.
2. The two normalization implementations will drift over time.

Consider a single source of truth — e.g., a small helper script the workflow `source`s, or matching the workflow's awk to also drop the ignore-comment lines. (Re-raising @RolandKrummenacher's top-level point at the line level for visibility.)

src/scripts/Publish-Toolkit.ps1
```diff
@@ -205,6 +209,64 @@ function Set-RepoContent($repo, [string]$branchPrefix, [string]$sourceDir)
...
+}
```
flanakin, line 286: 🤖 [AI][Claude Opus 4.7] ⚠️ **Should fix**

`Get-DocsArticleBody` is pure and easy to unit test, but there are no Pester tests covering it. The new tests in `Action.UpdateMsLearnDates.Tests.ps1` regex-match the workflow YAML (fine for the workflow guard) but don't exercise this function at all.

Minimal coverage worth adding (paraphrasing @RolandKrummenacher's suggestion):

```powershell
It 'Get-DocsArticleBody strips frontmatter'
It 'Get-DocsArticleBody strips markdownlint/prettier ignore lines'
It 'Get-DocsArticleBody collapses 3+ blank lines to 2'
It 'Get-DocsArticleBody trims trailing whitespace per line'
It 'Get-DocsArticleBody returns "" for empty or frontmatter-only input'
```

If both the workflow's `strip_frontmatter` and this function are required to satisfy a shared fixture, the asymmetry above becomes visible the moment one drifts.

src/scripts/Publish-Toolkit.ps1
```diff
@@ -205,6 +209,64 @@ function Set-RepoContent($repo, [string]$branchPrefix, [string]$sourceDir)
...
+        }
+    }
+
+    $ignoreLines = @(
+        "<!-- markdownlint-disable-next-line MD025 -->",
+        "<!-- prettier-ignore-start -->",
+        "<!-- prettier-ignore-end -->"
+    )
+
+    $body = $lines[$start..($lines.Count - 1)] `
```
flanakin: 🤖 [AI][Claude Opus 4.7] 💡 **Suggestion**

Edge case: if a file is frontmatter-only (closing `---` is the last line), `$start` equals `$lines.Count` and `$lines[$start..($lines.Count - 1)]` becomes `$lines[Count..(Count-1)]`, which PowerShell evaluates as a reversed range and returns the last element instead of an empty slice. The function would then return that line (likely empty), which happens to still compare equal across versions, so the visible blast radius is small — but the logic is wrong.

Guard with a length check, e.g.:

```suggestion
    if ($start -ge $lines.Count)
    {
        return ""
    }

    $body = $lines[$start..($lines.Count - 1)] `
```

(Existing pipeline continues from `| Where-Object ...`.)

src/scripts/Publish-Toolkit.ps1
```diff
@@ -205,6 +209,64 @@ function Set-RepoContent($repo, [string]$branchPrefix, [string]$sourceDir)
...
+    return (($body -join "`n") -replace "`n{3,}", "`n`n").Trim()
+}
+
+function Remove-DocsMetadataOnlyChanges([string]$docsPath)
+{
+    Push-Location
+    Set-Location $docsPath
+    $repoRoot = git rev-parse --show-toplevel
+    Set-Location $repoRoot
+
```
flanakin, line 296: 🤖 [AI][Claude Opus 4.7] 💡 **Suggestion**

The function changes the process cwd (`Set-Location $repoRoot`) so that `git restore --source HEAD -- $repoFile` later resolves `$repoFile` against the correct repo. It works, but it's implicit — anyone reading the loop has to remember the cwd was changed 15 lines earlier.

Passing `-C $repoRoot` to each `git` call would make the intent explicit and remove the need to change cwd:

```powershell
git -C $repoRoot cat-file -e "HEAD:$repoFile" 2>$null
...
$previousContent = [string]::Join("`n", (git -C $repoRoot show "HEAD:$repoFile"))
...
git -C $repoRoot restore --source HEAD -- $repoFile
```

Minor — purely a readability/robustness improvement.

flanakin (commented):
🤖 [AI][Claude Opus 4.7] **PR Review**

**Summary:** Focused, well-scoped fix that achieves its stated goal. Workflow guards and the publish-side cleanup work together correctly. No blockers. Re-emphasizing two items already raised by @RolandKrummenacher at the line level for code-review visibility, plus two additional suggestions.

### ⚠️ Should fix (2)

1. Asymmetric body normalization between workflow (`strip_frontmatter` only) and `Get-DocsArticleBody` (frontmatter + ignore comments + whitespace) — already raised in @RolandKrummenacher's top-level comment; pinning at the line for context.
2. No behavior tests for `Get-DocsArticleBody` / `Remove-DocsMetadataOnlyChanges` — also from @RolandKrummenacher's comment.

### 💡 Suggestions (2)

3. `Get-DocsArticleBody` produces a reversed range when a file is frontmatter-only (`$start == $lines.Count`).
4. `Remove-DocsMetadataOnlyChanges` relies on cwd for `git restore` to resolve paths — `git -C $repoRoot` would be more robust and self-documenting.

RolandKrummenacher (commented):
## Code review — skip ms.date updates & republish for metadata-only changes

**Overview:** Gates both the `update-mslearn-dates` Action and `Publish-Toolkit.ps1` on actual article-body changes, so frontmatter-only edits stop bumping `ms.date` and stop generating metadata-only churn in docs republish branches. Clean, well-targeted fix.

### What's done well
- **Workflow guard is correct.** `strip_frontmatter` (awk) + `cmp -s` against `pull_request.base.sha` is the right approach. I confirmed the workflow triggers on `pull_request` only and checks out with `fetch-depth: 0`, so `base.sha` is always populated and reachable — `git cat-file -e`/`git show "$BASE_SHA:$file"` will resolve. New/deleted files are handled (`cat-file -e` miss → treat as changed; `[ ! -f ]` → skip). Process substitution is fine since GH `run:` blocks are bash.
- **`Remove-DocsMetadataOnlyChanges` logic is sound** — per-file `HEAD` comparison with `git restore` to drop metadata-only diffs, new files (`cat-file -e` miss) correctly left alone, line-ending differences absorbed by the `\r?\n` split + `.Trim()`.

### Issues

**1. (Medium) PowerShell descending-range bug in `Get-DocsArticleBody` for degenerate files.**
```powershell
$body = $lines[$start..($lines.Count - 1)]
```
If `$start` lands past the last index — e.g. a **frontmatter-only** file where the closing `---` is the last line (`$start == $lines.Count`), or an empty file — `$start..($lines.Count-1)` becomes a *descending* range (PowerShell quirk: `5..4` → `@(5,4)`), so this indexes out of bounds / reverses lines and yields garbage body text rather than empty. Guard it:
```powershell
$body = if ($start -lt $lines.Count) { $lines[$start..($lines.Count - 1)] | ... } else { @() }
```
Unlikely for real articles (they have bodies), but it's a silent-wrong rather than loud-fail, so worth the one-liner.

**2. (Low) Divergent definitions of "body" between the two code paths.**
The workflow's `strip_frontmatter` removes *only* frontmatter, while `Get-DocsArticleBody` additionally strips markdownlint/prettier ignore comments and normalizes whitespace + blank-line runs. So a whitespace-only or ignore-comment-only edit would be treated as a content change by the Action (bumps `ms.date`) but as metadata-only by Publish-Toolkit (restored). Probably acceptable since they run at different stages, but the asymmetry is a future foot-gun — worth a comment noting the two are intentionally different, or factoring toward one definition.

**3. (Low) Test coverage is on the wrong layer.**
The three new Pester assertions are *string-match pins* on the workflow YAML (`base.sha`, `strip_frontmatter`, `cmp -s`, the skip message). They'll catch a guard being deleted, but validate no behavior. Meanwhile the **new PowerShell functions — where the actual logic and the range bug live — have no tests at all.** `Get-DocsArticleBody` is pure and trivially unit-testable; a few cases (normal article, frontmatter-only, no-closing-`---`, ignore-comment-only, whitespace-only) would cover the risky surface and pin the range fix above.

### Nits
- `Get-DocsArticleBody` doesn't handle a frontmatter block with no closing `---` (keeps the whole file as body). Errs safe (treats as content change), just noting.
- awk `$0 == "---"` is an exact match — a `---` line with trailing whitespace or CRLF wouldn't be recognized as a fence. Docs are LF so fine in practice.

**Verdict: Approvable once the range guard (#1) is added.** The workflow side is correct and the overall approach is right; my only real ask is the one-line guard in `Get-DocsArticleBody`, ideally with a couple of unit tests for that function since it's the untested part carrying the logic.

src/scripts/Publish-Toolkit.ps1
```diff
@@ -205,6 +209,64 @@ function Set-RepoContent($repo, [string]$branchPrefix, [string]$sourceDir)
...
+        $previousContent = [string]::Join("`n", (git show "HEAD:$repoFile"))
+        $currentContent = Get-Content $_.FullName -Raw
+
+        if ((Get-DocsArticleBody $previousContent) -eq (Get-DocsArticleBody $currentContent))
+        {
+            git restore --source HEAD -- $repoFile
+        }
+    }
+
+    Pop-Location
```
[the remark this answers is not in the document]
MSBrett: Verified in `pwsh` that this is not the case — `return` inside a `ForEach-Object` scriptblock skips the current pipeline item (per-iteration), not the enclosing function. The pipeline keeps processing later files and `Pop-Location` is reached.

```pwsh
function Test-ReturnBehavior {
    1..3 | ForEach-Object {
        if ($_ -eq 2) { return }
        Write-Host "  Saw: $_"
    }
    Write-Host "  After pipeline (should reach)"
}
# Output: Saw: 1 / Saw: 3 / After pipeline (should reach)
```

That said, the underlying concern that a terminating error could skip `Pop-Location` is valid, so I wrapped the body in `try { ... } finally { Pop-Location }` in 7d3d3ec8 and added a comment noting the per-iteration semantics.

src/scripts/Publish-Toolkit.ps1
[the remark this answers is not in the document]
MSBrett: Confirmed and fixed in 7d3d3ec8 — added a guard `if ($start -ge $lines.Count) { return "" }` before the slice. Also covered by new Pester tests in `Publish-Toolkit.GetDocsArticleBody.Tests.ps1` (frontmatter-only and empty-input cases).

src/scripts/Publish-Toolkit.ps1
```diff
@@ -205,6 +209,64 @@ function Set-RepoContent($repo, [string]$branchPrefix, [string]$sourceDir)
...
+        $repoFile = [System.IO.Path]::GetRelativePath($repoRoot, $_.FullName).Replace("\", "/")
+        git cat-file -e "HEAD:$repoFile" 2>$null
+
+        if ($LASTEXITCODE -ne 0)
+        {
+            return
+        }
+
+        $previousContent = [string]::Join("`n", (git show "HEAD:$repoFile"))
+        $currentContent = Get-Content $_.FullName -Raw
```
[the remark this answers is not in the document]
MSBrett: Leaving as-is. `Remove-DocsMetadataOnlyChanges` only runs when `-Template -eq "docs"` during a release publish — it is not on any hot path. `git cat-file -e` is cheap (object-database lookup, no working-copy work) and the docs tree is bounded. Optimizing to `git status --porcelain` first would add complexity for negligible savings on a once-per-release script. Happy to revisit if publish runtime ever becomes an issue.

src/scripts/Publish-Toolkit.ps1
```diff
@@ -205,6 +209,64 @@ function Set-RepoContent($repo, [string]$branchPrefix, [string]$sourceDir)
...
+    Get-ChildItem $docsPath -Recurse -Filter "*.md" | ForEach-Object {
+        $repoFile = [System.IO.Path]::GetRelativePath($repoRoot, $_.FullName).Replace("\", "/")
+        git cat-file -e "HEAD:$repoFile" 2>$null
+
+        if ($LASTEXITCODE -ne 0)
+        {
+            return
+        }
+
+        $previousContent = [string]::Join("`n", (git show "HEAD:$repoFile"))
```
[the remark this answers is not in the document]
MSBrett: Not a correctness issue here. PowerShell does not shell-split bare variable arguments when invoking native commands — `git restore --source HEAD -- $repoFile` passes `$repoFile` as a single argv element even if the path contains spaces or special characters. Verified in `pwsh`. Real `docs-mslearn/**/*.md` paths in this repo have no spaces, so the quoting would be cosmetic.

src/scripts/Publish-Toolkit.ps1
```diff
@@ -205,6 +209,64 @@ function Set-RepoContent($repo, [string]$branchPrefix, [string]$sourceDir)
...
+        $currentContent = Get-Content $_.FullName -Raw
+
+        if ((Get-DocsArticleBody $previousContent) -eq (Get-DocsArticleBody $currentContent))
+        {
+            git restore --source HEAD -- $repoFile
```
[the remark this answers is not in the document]
MSBrett: Same as the comment above — PowerShell passes bare variable arguments to native commands as a single argv element, so `$repoFile` is not shell-split even with spaces or special characters. No correctness issue.

src/scripts/Publish-Toolkit.ps1
MSBrett: Addressed in 7d3d3ec8 by extending the workflow's `strip_frontmatter` awk to also drop the same three ignore-comment lines (`<!-- markdownlint-disable-next-line MD025 -->`, `<!-- prettier-ignore-start -->`, `<!-- prettier-ignore-end -->`), with `gsub` whitespace trimming so leading/trailing space does not defeat the match. The new workflow test `Should ignore markdownlint and prettier directives in body comparison` pins this so the workflow and `Get-DocsArticleBody` cannot silently drift on the directive list.

The residual asymmetry (PowerShell additionally trims trailing whitespace and collapses 3+ blank lines to 2) is intentional and not load-bearing for the workflow — those normalizations exist to absorb publish-pipeline line-ending and copy-time whitespace shifts that the workflow's git-vs-working-copy comparison does not encounter. Extracting a single helper for the workflow's bash and PowerShell environments would add deployment complexity that is not justified for those two cosmetic deltas.

src/scripts/Publish-Toolkit.ps1
MSBrett: Added in 7d3d3ec8 — new file `src/powershell/Tests/Unit/Publish-Toolkit.GetDocsArticleBody.Tests.ps1` with 11 behavior tests covering:

- normal article with frontmatter
- frontmatter-only file (range guard)
- empty input
- file with no closing `---` (errs safe — keeps whole file as body)
- each ignore directive stripped (with surrounding whitespace too)
- trailing whitespace trimming
- 3+ blank lines collapsed to 2
- CRLF input matches LF
- only-frontmatter-change yields equal body

Because `Publish-Toolkit.ps1` has top-level executable code, the test file AST-extracts the `Get-DocsArticleBody` function definition and defines it in the test scope instead of dot-sourcing the whole script. All 40 tests in the unit suite pass.

src/scripts/Publish-Toolkit.ps1
MSBrett: Fixed in 7d3d3ec8 with `if ($start -ge $lines.Count) { return "" }` before the slice. New Pester tests cover frontmatter-only (returns `""`) and empty input (returns `""`) so the guard is pinned.

src/scripts/Publish-Toolkit.ps1
MSBrett: Not adopting the `git -C` rewrite, but the underlying robustness concern is now addressed — in 7d3d3ec8 the function wraps its body in `try { ... } finally { Pop-Location }`, so a terminating error inside the loop cannot leave the caller in the wrong cwd. That is the property `git -C $repoRoot` would have given by removing the need to change cwd in the first place. The remaining choice is purely stylistic and the current shape stays consistent with `Push-Location`/`Pop-Location` usage elsewhere in this script.

src/scripts/Publish-Toolkit.ps1
MSBrett: Done in `67dc5518`. Both `strip_frontmatter` (awk) and `Get-DocsArticleBody` (PowerShell) now apply the same six-step normalization in the same order:

1. Strip a trailing `\r` per line (CRLF tolerance).
2. Whitespace-tolerant `---` fence detection; missing closing fence → treat whole file as body (errs safe).
3. Drop the three ignore-directive lines (whitespace-tolerant trim equality).
4. `TrimEnd` per line.
5. Collapse runs of 2+ blank lines down to 1 (matches the old `\n{3,}` → `\n\n` regex behavior exactly).
6. Strip leading and trailing blank lines.

The new `Action.UpdateMsLearnDates.AwkParity.Tests.ps1` extracts the awk script from the workflow YAML and asserts equivalence-class parity between the two implementations across fixtures for ms.date-only edits, ignore-directive-only edits, trailing whitespace, CRLF input, frontmatter-only files, whitespace-only bodies, fences with surrounding whitespace, and files with no closing fence — so if either side drifts again the parity test will fail.

MSBrett:
Thanks, Roland — addressed your asks in the latest commits.

1. **Range guard:** `Get-DocsArticleBody` now returns an empty body when the calculated body start is past the end of the file, so empty/frontmatter-only files no longer hit PowerShell's descending-range behavior.
2. **Workflow/publish alignment:** updated the workflow `strip_frontmatter` implementation and comments to match the publish-time `Get-DocsArticleBody` semantics for frontmatter, ignore directives, CRLF handling, and leading/trailing blank body lines.
3. **Behavior coverage:** added focused `Get-DocsArticleBody` unit tests for normal articles, frontmatter-only, empty input, no closing fence, ignore directives, CRLF, and metadata-only changes. Also added awk-vs-PowerShell parity tests so the workflow and publish logic stay aligned.

Targeted validation passes locally: `Invoke-Pester -Output Detailed -Path ./src/powershell/Tests/Unit/Publish-Toolkit.GetDocsArticleBody.Tests.ps1,./src/powershell/Tests/Unit/Action.UpdateMsLearnDates.AwkParity.Tests.ps1` → 19/19 passed. GitHub checks are green as well.

Re-review already requested — appreciate another look when you have a chance.

RolandKrummenacher:
## Re-review — skip ms.date updates & republish for metadata-only changes

**Verdict: Ready to merge.** Both concerns from the prior review round are resolved, and I verified the fix by running the tests rather than by inspection alone.

### Verified
Checked out the branch and ran all three test suites locally (awk + Pester 5.7.1): **48/48 pass**, including:
- **Cross-engine parity** — the new `Action.UpdateMsLearnDates.AwkParity.Tests.ps1` proves the workflow's `strip_frontmatter` (awk) and `Get-DocsArticleBody` (PowerShell) agree on equivalence for *every pair* of fixtures. This closes the **awk/PowerShell asymmetry** I raised earlier: both paths now share an identical, documented rule list (fence detection, MD025/prettier-ignore stripping, blank-line trimming, whitespace preservation).
- **Range guard** — frontmatter-only and empty files return `""` cleanly; the descending-range issue is gone and explicitly tested.
- CRLF parity, trailing hard-break detection, and blank-line-runs-in-code-blocks all covered.

CI is green (Pester, CodeQL, CLA); `MERGEABLE`, blocked only by required review.

### Code correctness
The publish-side flow is sound: `Remove-DocsMetadataOnlyChanges` runs after copy / before commit, skips files absent from `HEAD` (new files kept), and `git restore`s only body-unchanged files — so metadata-only churn never reaches the republish commit. CRLF/LF differences normalize away in both engines. The `AGENTS.md` guidance is updated to match the new behavior.

### Non-blocking notes
1. **Ignore-directive stripping is unscoped** — `<!-- prettier-ignore-* -->` / MD025 lines are dropped anywhere, including inside fenced code blocks. If a doc ever shows those directives as literal example content, a metadata-only edit could mask a body change. Very low risk, and **symmetric** across both engines so it can't cause workflow/publish disagreement. Worth a comment at most.
2. **New/renamed files always bump `ms.date`** (absent from base SHA → treated as changed). Intentional and acceptable; noting for visibility.
3. **Parity test is O(n²)** with one awk process spawned per cell (~13.5s locally). Hoisting the per-input body computation out of the double loop would speed it up if fixtures grow — minor test-perf nit.

RolandKrummenacher (approved):
Approving. Re-reviewed and verified the fix end-to-end (ran all three test suites locally, 48/48 pass, including the cross-engine awk↔PowerShell parity test). Both prior-round concerns — body-normalization asymmetry and the frontmatter-only range guard — are resolved. CI green. The three notes in my re-review comment are non-blocking.

src/scripts/Publish-Toolkit.ps1
```diff
@@ -205,6 +209,116 @@ function Set-RepoContent($repo, [string]$branchPrefix, [string]$sourceDir)
...
+
+    if ($s -gt $e)
+    {
+        return ""
+    }
+
+    return ($body[$s..$e] -join "`n")
+}
+
+function Remove-DocsMetadataOnlyChanges([string]$docsPath)
```
flanakin, line 288: 🤖 [AI][Claude Opus 4.8 (1M context)] ⚠️ **Should fix**

`Remove-DocsMetadataOnlyChanges` overwrites working-tree files via `git restore --source HEAD` — a destructive, state-changing operation — but does not support `-WhatIf`/`-Confirm`. `CLAUDE.md` requires `ShouldProcess` support for destructive operations, and PSScriptAnalyzer flags this as `PSUseShouldProcessForStateChangingFunctions`.

Consider adding `[CmdletBinding(SupportsShouldProcess)]` and gating the restore, e.g.:

```powershell
function Remove-DocsMetadataOnlyChanges
{
    [CmdletBinding(SupportsShouldProcess)]
    param([string]$docsPath)
    ...
    if ((Get-DocsArticleBody $previousContent) -eq (Get-DocsArticleBody $currentContent))
    {
        if ($PSCmdlet.ShouldProcess($repoFile, 'Restore metadata-only change from HEAD'))
        {
            git restore --source HEAD -- $repoFile
        }
    }
}
```

The existing call site (`Remove-DocsMetadataOnlyChanges $repo.path`) keeps working unchanged; `-WhatIf` then lets a publish dry-run show which files would be reverted before touching them.

_Note: this file isn't covered by `Lint.Tests.ps1` (which scans only `src/powershell/`), so it won't fail CI — advisory._

flanakin (commented):
🤖 [AI][Claude Opus 4.8 (1M context)] **PR Review**

**Summary:** Focused, correct fix. I re-derived findings from the current working tree (not thread history) and independently verified: all 48 tests pass; the awk and PowerShell body-normalization paths agree on equivalence classes end-to-end (metadata-only edits, ignore directives, CRLF, blank-line trimming); the range guard for frontmatter-only/empty files is present; and the `git show`-vs-`Get-Content -Raw` trailing-newline asymmetry is correctly absorbed by trailing-blank-line stripping. The earlier review concerns (asymmetric normalization, missing range guard, missing behavior tests) are all genuinely resolved, and the new `AwkParity` test is a strong guard against future drift. No blockers.

### ⚠️ Should fix (1)

1. `Remove-DocsMetadataOnlyChanges` performs a destructive working-tree operation (`git restore`) without `-WhatIf`/`-Confirm` support.

_Note: this does not fail CI — `Lint.Tests.ps1` scans only `src/powershell/`, not `src/scripts/`, so `Publish-Toolkit.ps1` (including its pre-existing `Write-Host` usages) is unlinted. Advisory, not blocking._

PR #2162, opened 2026-05-29
NAT Gateway for private subnets (fixes #2161)

### LINKED ISSUE #2161: FinOps Hub private deployment fails under 'Subnets should be private' policy / Sept 2025 default-outbound retirement
## Summary

FinOps Hub private deployments (`enablePublicAccess: false`) fail in environments enforcing the Azure built-in policy **"Subnets should be private"** (policy definition `fe505f54d90b47b3b60de101`). That policy denies any subnet whose `defaultOutboundAccess` is not explicitly set to `false`, which becomes the platform default after the Sept 2025 retirement of implicit subnet outbound (see https://aka.ms/defaultoutboundaccess).

## Repro

1. Assign the built-in policy "Subnets should be private" at subscription or management-group scope.
2. Deploy the FinOps Hub template with `enablePublicAccess: false` and any address prefix.
3. Deployment fails with `RequestDisallowedByPolicy` on `Microsoft.Network/virtualNetworks/<hub>-vnet-<region>`:

```
Resource '<hub>-vnet-<region>' was disallowed by policy.
Policy identifiers: 'Subnets should be private' (fe505f54d90b47b3b60de101)
```

## Root cause

`src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Core/infrastructure.bicep` defines three subnets (`private-endpoint-subnet`, `script-subnet`, `dataExplorer-subnet`) without setting `defaultOutboundAccess: false`. The hub's NSG has an explicit `AllowInternetOutBound` rule (priority 200), so the template was designed around implicit subnet outbound — which the policy and the upcoming platform default no longer permit.

## Why the trivial fix is insufficient

Naively adding `defaultOutboundAccess: false` to the subnets satisfies the policy but **breaks the hub** because the private deployment depends on outbound internet from two subnets:

| Subnet | Outbound dependency | Why |
|--------|--------------------|----- |
| `script-subnet` | `mcr.microsoft.com` | `Microsoft.Resources/deploymentScripts` runs ACI containers using the `azuredeploymentscripts-powershell` image. Microsoft does not expose a private alternative for this image and the schema has no `containerSettings.image` override (verified against the ARM schema and the AVM `deployment-script` module). |
| `dataExplorer-subnet` | `raw.githubusercontent.com` | The `config_InitializeHub` ADF pipeline runs four `.set-or-replace` KQL commands against the ADX cluster using `externaldata()` to pull `PricingUnits.csv`, `Regions.csv`, `ResourceTypes.csv`, and `Services.csv` from the toolkit's open-data folder on GitHub. ADX makes the outbound HTTPS call directly. |

`private-endpoint-subnet` is inbound-only (hosts private endpoints) and does not need outbound.

Explicit egress is therefore mandatory in private mode. The supported pattern per Microsoft docs for ACI-in-VNet is a **NAT Gateway with a Standard Public IP** (https://learn.microsoft.com/azure/container-instances/container-instances-virtual-network-concepts).

## Proposed fix

1. Set `defaultOutboundAccess: false` on all three subnets.
2. Add a Standard NAT Gateway + Standard static Public IP to the hub template, attached to `script-subnet` and `dataExplorer-subnet`. Both gated on the existing private-routing flag so public deployments are unaffected.
3. While here: introduce a tri-state network mode parameter (`public` / `vnet` / `private`) that replaces the binary `enablePublicAccess` boolean (kept as a deprecated back-compat shim). The new middle `vnet` mode deploys only the free VNet + NSG scaffold (no NAT, no PE) so customers can stage a future private upgrade without paying for unused egress. The portal UI (`createUiDefinition.json`) is updated to surface this via an `OptionsGroup` with per-mode notes documenting cost and downgrade behavior.
4. Fix six pre-existing tag-namespace bugs in `infrastructure.bicep` where `Microsoft.Storage/*` was used for network resources that should be `Microsoft.Network/*`, and add the four new network resource types to the portal's `TagsByResource` resource list.

## Notes

- Mode downgrades (private → vnet or private → public) do not delete orphaned NAT Gateway / Public IP / private endpoints because deployments are incremental by default. This is documented in the new param description and in the portal UI privateModeNote.
- The new `vnet` mode delivers a free VNet scaffold only. Storage / Key Vault / ADX retain public endpoints in this mode; this is documented in the UI vnetModeNote so customers don't mistake it for private isolation.
- Verified end-to-end in three independent Azure deployments (public, vnet, private). Three rubber-duck reviews (gpt-5.5, claude-opus-4.6, lark) converged on remediation findings that are included in the PR.

## Related

- Azure built-in policy: https://www.azadvertizer.net/azpolicyadvertizer/fe505f54d90b47b3b60de101.html
- Default outbound retirement: https://aka.ms/defaultoutboundaccess
- ACI VNet concepts (NAT Gateway requirement): https://learn.microsoft.com/azure/container-instances/container-instances-virtual-network-concepts

MSBrett:
## Closes #2161

Adds a tri-state network mode to the FinOps Hub Bicep template and the portal UI, and adds a NAT Gateway to the private deployment path so it works under the Azure built-in policy *Subnets should be private* (and after the Sept 2025 implicit subnet outbound retirement).

## Summary of changes

| File | Change |
|------|--------|
| `src/templates/finops-hub/main.bicep` | New `networkMode` param (`public` \| `vnet` \| `private`) with `@allowed`. Existing `enablePublicAccess` bool kept and marked DEPRECATED. `effectiveNetworkMode` derives from explicit `networkMode` when set, otherwise from the legacy bool. |
| `src/templates/finops-hub/modules/hub.bicep` | Replaces `enablePublicAccess` with `networkMode` (passed in from `main.bicep`). Telemetry suffix gains `'V'` for vnet (was `''` for public, `'P'` for private). |
| `src/templates/finops-hub/modules/fx/hub-types.bicep` | `newHub()` and `newHubInternal()` take `networkMode` instead of `enablePublicAccess`. Adds `vnetIntegration: bool` to `HubProperties.options` (true when networkMode != public). `privateRouting` stays as it was (true when networkMode == private). `routing.subnets.*` populates when VNet exists (`networkMode != 'public'`). |
| `src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Core/infrastructure.bicep` | Sets `defaultOutboundAccess: false` on all 3 subnets. Adds Standard NAT Gateway + Standard static Public IP, gated on `privateRouting`, attached to script + dataExplorer subnets via conditional `union()` on subnet properties (PE subnet correctly excluded). VNet + NSG move from `privateRouting` to `vnetIntegration`. Fixes 6 pre-existing wrong tag namespaces (`Microsoft.Storage/*` → `Microsoft.Network/*` for NSG / VNet / privateDnsZones). VNet outputs re-gated on `vnetIntegration`. |
| `src/templates/finops-hub/createUiDefinition.json` + `docs/deploy/finops-hub-latest.ui.json` | Replaces 2-option Public/Private DropDown with 3-option `networkMode` `OptionsGroup`. Conditional `TextBlock`s per mode explain vnet vs private tradeoffs (cost, what gets isolated, downgrade behavior). Outputs emit both `networkMode` (new) and a derived `enablePublicAccess` (back-compat for ARM/CLI callers). `TagsByResource` list gains `Microsoft.Network/{virtualNetworks,networkSecurityGroups,natGateways,publicIPAddresses}`. |

## Validation

- `az bicep build` clean (only pre-existing unused-param warnings in `Recommendations/app.bicep`).
- `az deployment group what-if` passed for all three modes against a real RG, no policy denials.
- Real end-to-end deployments succeeded for all three variants:
  - `networkMode=public`: 12m11s, 17 resources, 0 network resources (no VNet).
  - `networkMode=vnet`: 30m57s, VNet + NSG only (no NAT, no PE, no DNS — `defaultOutboundAccess: false` on all subnets, no `natGateway` reference).
  - `networkMode=private`: 33m17s, full stack — VNet + NSG + NAT Gateway + Public IP + 4 PEs (blob, dfs, script-blob, ADX cluster) + 5 private DNS zones (blob, dfs, queue, table, kusto). ADX `publicNetworkAccess: Disabled`. All 11 deployment-script ACI containers ran successfully, proving NAT Gateway → MCR egress works.

## Back-compat

- `enablePublicAccess` remains a valid parameter. `true` (default) maps to `public`. `false` maps to `private`. CLI / ARM callers that pass only the legacy bool continue to behave identically.
- New `vnet` middle tier requires explicitly passing `networkMode='vnet'` and cannot be expressed via the legacy bool.
- Direct callers of `modules/hub.bicep` (a module that is only consumed by this template's `main.bicep` per repo grep) must migrate to `networkMode`.

## Operational notes (also surfaced in the portal UI)

- Mode downgrade (private → vnet → public) does NOT delete orphaned VNet / NAT / Public IP / private endpoints because deployments are incremental by default. Documented in the `networkMode` param description and in the portal `privateModeNote`.
- `vnet` mode delivers a *free* VNet scaffold only — storage / KV / ADX keep public endpoints in this mode. This is intentional: it lets customers pre-stage the VNet for a later flip to `private` without paying for unused NAT / Public IP hours. The portal `vnetModeNote` documents this so customers don't mistake it for private isolation.

## Reviewers

The design was reviewed by three independent rubber-duck reviews (gpt-5.5, claude-opus-4.6, lark). All three converged on two HIGH findings, both addressed in this PR before submission:
1. `vnet` mode originally created NAT Gateway + Public IP that had no workload routing through them ($36/mo dead infrastructure) — fixed by gating NAT/PIP on `privateRouting` only and conditionally omitting the `natGateway` property from subnets in vnet mode.
2. Mode downgrades silently leave orphaned resources (incremental deployment) — addressed by explicit documentation in param description + UI note.

### DIFF
--- a/docs/deploy/finops-hub-latest.ui.json
+++ b/docs/deploy/finops-hub-latest.ui.json
@@ -890,33 +890,52 @@
             "label": "Networking",
             "elements": [
               {
-                "name": "infraEncryptionIntro",
+                "name": "networkingIntro",
                 "type": "Microsoft.Common.TextBlock",
                 "visible": true,
                 "options": {
-                  "text": "For enhanced security, consider blocking external access to your FinOps hub instance to enable secure, private connectivity to your cloud resources by eliminating exposure to the public internet. Private endpoints can incur significant additional cost. Please understand the cost impact before selecting this option."
+                  "text": "Select the network isolation mode. Private vNet uses private endpoints to block all public access and incurs significant additional cost (private endpoint hours, data processing, plus a NAT Gateway with a public IP for outbound dependencies). Please understand the cost impact before selecting Private vNet."
                 }
               },
               {
-                "name": "enablePublicAccess",
-                "type": "Microsoft.Common.DropDown",
-                "label": "Access",
-                "toolTip": "Indicate whether FinOps hubs resources should be accessible by the internet or not. If set to private, all resources will be hidden behind a virtual network private endpoint. This will incur additional costs for data crossing virtual network boundaries and for virtual network peering, if configured.",
-                "defaultValue": "Public",
-                "multiLine": true,
+                "name": "networkMode",
+                "type": "Microsoft.Common.OptionsGroup",
+                "label": "Network access mode",
+                "defaultValue": "Public (no vNet)",
+                "toolTip": "Choose how FinOps hubs resources are exposed on the network. Public: no virtual network, all resources reachable from the internet. vNet: deploy resources inside a virtual network with a NAT Gateway for outbound traffic, but keep public endpoints (segmentation without private-endpoint cost). Private vNet: all resources are hidden behind private endpoints; public access is blocked. This will incur additional costs for private endpoints, data processing across virtual network boundaries, and for virtual network peering if configured.",
                 "constraints": {
                   "allowedValues": [
                     {
-                      "label": "Public",
-                      "description": "Allow external access to storage and Data Explorer.",
-                      "value": true
+                      "label": "Public (no vNet)",
+                      "value": "public"
                     },
                     {
-                      "label": "Private",
-                      "description": "Block all external access. This will incur additional charges.",
-                      "value": false
+                      "label": "vNet (public endpoints)",
+                      "value": "vnet"
+                    },
+                    {
+                      "label": "Private vNet (private endpoints only)",
+                      "value": "private"
                     }
-                  ]
+                  ],
+                  "required": true
+                },
+                "visible": true
+              },
+              {
+                "name": "vnetModeNote",
+                "type": "Microsoft.Common.TextBlock",
+                "visible": "[equals(steps('advanced').networking.networkMode, 'vnet')]",
+                "options": {
+                  "text": "A virtual network + NSG + 3 subnets are deployed as a scaffold for future expansion, but public endpoints remain. No NAT Gateway and no private endpoints are created, so this mode adds no Azure cost beyond the free VNet primitives. Useful if you want a VNet ready to flip to Private vNet later without recreating resources."
+                }
+              },
+              {
+                "name": "privateModeNote",
+                "type": "Microsoft.Common.TextBlock",
+                "visible": "[equals(steps('advanced').networking.networkMode, 'private')]",
+                "options": {
+                  "text": "Private endpoints and private DNS zones will be created for storage, Key Vault, and Data Explorer. A NAT Gateway with a static public IP is also deployed so deployment scripts and Data Explorer can reach Microsoft Container Registry and the FinOps toolkit open data on GitHub. This incurs charges for private endpoint hours, data processing, NAT Gateway hours, and the public IP. Switching back to vNet or Public later does NOT delete these resources — clean them up manually if you downgrade."
                 }
               },
               {
@@ -933,7 +952,7 @@
                     }
                   ]
                 },
-                "visible": "[equals(steps('advanced').networking.enablePublicAccess, false)]"
+                "visible": "[not(equals(steps('advanced').networking.networkMode, 'public'))]"
               }
             ],
             "visible": true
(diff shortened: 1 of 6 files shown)


MSBrett:
Superseded by #2163 — same branch name and contents, head moved to microsoft/finops-toolkit:fix/subnet-default-outbound-access. #2163 also includes the tooltip/param-description fix addressing the three copilot-pull-request-reviewer findings on this PR (vnet-mode text incorrectly claimed a NAT Gateway is deployed).

PR #2183, opened 2026-06-16
fix(powershell): return full export run history in Get-FinOpsCostExport

### LINKED ISSUE #2063: Get-FinOpsCostExport -RunHistory returns incomplete run history
## Summary

`Get-FinOpsCostExport -RunHistory` returns fewer runs than expected because it uses the Cost Management **list** API (`GET .../exports?$expand=runHistory`), which truncates run history per export. The individual **get** API (`GET .../exports/{name}?$expand=runHistory`) returns the full run history.

## Repro

1. Create an export with `-Backfill 3` (generates multiple runs)
2. Run `Get-FinOpsCostExport -Scope $scope -RunHistory`
3. Compare the `RunHistory` count with what the Azure portal shows

**Expected:** All runs returned (e.g., 2 runs)
**Actual:** Only 1 run returned

## Root cause

The cmdlet calls the list endpoint for all exports at the scope, which returns a reduced `runHistory.value` array per export. The individual GET endpoint returns the complete history.

**List API** (used by cmdlet):
```
GET {scope}/providers/Microsoft.CostManagement/exports?api-version=2025-03-01&$expand=runHistory
```
→ Returns 1 run per export

**Individual GET API** (returns full history):
```
GET {scope}/providers/Microsoft.CostManagement/exports/{name}?api-version=2025-03-01&$expand=runHistory
```
→ Returns all runs (e.g., 2)

## Possible fix

When `-RunHistory` is specified, after the initial list call, make an additional individual GET call per export to retrieve the full run history. This adds N API calls (one per export) but ensures complete data.

Alternatively, when `-Name` is specified without wildcards, use the individual GET endpoint directly.

## File

- `src/powershell/Public/Get-FinOpsCostExport.ps1`

RolandKrummenacher:
## 🪲 Fixes

Fixes #2063

`Get-FinOpsCostExport -RunHistory` returned fewer runs than expected because it relied solely on the Cost Management **list** endpoint (`GET .../exports?$expand=runHistory`), which truncates the `runHistory.value` array per export. The individual **get** endpoint (`GET .../exports/{name}?$expand=runHistory`) returns the complete run history.

## 🛠️ Changes

- When `-RunHistory` is specified, after the list call the cmdlet now makes an additional individual `GET` per export to retrieve the full run history (one extra API call per export).
- If the individual `GET` fails, it falls back to the run history returned by the list response so the cmdlet still returns data.
- Behavior without `-RunHistory` is unchanged (single list call, no `$expand`, no per-export calls).

## ✔️ Testing

Updated/added unit tests in `Get-FinOpsCostExport.Tests.ps1`:
- Verifies the individual GET endpoint is queried and the full run history (not the truncated list) is returned.
- Verifies no per-export call or `$expand` is made when `-RunHistory` is omitted.
- Verifies fallback to list run history when the individual GET fails.
- Verifies empty run history is handled.

All 5 tests pass; PSScriptAnalyzer reports no findings.

## 📃 Documentation

- Added a v15 PowerShell module entry to the changelog.


src/powershell/Public/Get-FinOpsCostExport.ps1
```diff
@@ -168,10 +168,32 @@ function Get-FinOpsCostExport
...
+
+            # The list endpoint truncates runHistory per export, so when -RunHistory is
+            # requested we re-fetch each export individually to get the full run history.
+            # https://github.com/microsoft/finops-toolkit/issues/2063
+            $runHistoryValue = $export.properties.runHistory.value
+            if ($RunHistory)
+            {
+                Write-Verbose -Message "fetching full run history for export $($export.name)"
+                $exportPath = "$($export.id.Trim('/'))?api-version=$ApiVersion&`$expand=runHistory"
+                $runResponse = Invoke-Rest -Method GET -Uri $exportPath -CommandName "Get-FinOpsCostExport"
```
flanakin: 🤖 [AI][Claude Code] ⚠️ **Should fix**

`Invoke-Rest` doesn't always return `Success=$false` on failure — it `throw`s when the error response has no structured `error` body (network failures, 5xx with non-JSON payloads, token acquisition errors). See [Invoke-Rest.ps1](src/powershell/Private/Invoke-Rest.ps1) (`throw $_.Exception.Message`). Since this call isn't wrapped in try/catch, such a failure on **any single export** propagates uncaught and aborts the entire `Get-FinOpsCostExport` call — which defeats the fallback's stated purpose ("so the cmdlet still returns data"). Wrapping the call lets the existing `else` branch handle thrown errors too.

```suggestion
                try
                {
                    $runResponse = Invoke-Rest -Method GET -Uri $exportPath -CommandName "Get-FinOpsCostExport"
                }
                catch
                {
                    $runResponse = $null
                    Write-Verbose -Message "individual GET threw for export $($export.name): $($_.Exception.Message)"
                }
```

flanakin: 🤖 [AI][Claude Code] ⚠️ **Should fix**

This adds one extra GET per export with no throttling awareness. The issue being fixed (#2063) is about exports with large run histories, which tend to live in busy accounts that also have many exports — so the N+1 fan-out here can realistically hit Cost Management rate limits mid-loop. The sibling [Start-FinOpsCostExport.ps1](src/powershell/Public/Start-FinOpsCostExport.ps1) already handles 429s with a wait-and-retry loop. `Invoke-Rest` surfaces `$runResponse.Throttled`; at minimum consider honoring it (retry on throttle) rather than silently falling back to the truncated list history, which would quietly reintroduce the very bug this PR fixes.

flanakin (commented):

**Summary:** Focused, well-tested fix for the truncated run-history problem. Tests pass (5/5), PSScriptAnalyzer is clean, the changelog targets the correct version (v15), and the `$_`-collision-safe extraction pattern is preserved. No blockers. Two error-handling/resilience items are worth addressing since the PR's stated goal is to keep returning data when the per-export call fails.

### ⚠️ Should fix (2)

1. Per-export `Invoke-Rest` isn't wrapped in try/catch. `Invoke-Rest` can `throw` (not just return `Success=$false`) on network errors, 5xx with a non-JSON body, or token failures — that throw propagates uncaught and aborts the whole cmdlet, defeating the documented fallback intent.
2. N+1 calls with no throttle handling. One extra GET per export with no 429 awareness, unlike the sibling `Start-FinOpsCostExport` which retries on throttling. Accounts with many exports could be throttled mid-loop.

### 💡 Suggestions (1)

3. `$_` vs `$export` inconsistency — line 171 aliases `$_` to `$export`, but the `$item` hashtable still uses `$_`. Works correctly; using `$export` throughout would read more consistently.

src/powershell/Public/Get-FinOpsCostExport.ps1
```diff
@@ -168,10 +168,32 @@ function Get-FinOpsCostExport
         }
         $exportdetails = @()
         $content | ForEach-Object {
+            $export = $_
```
flanakin, line 171: 🤖 [AI][Claude Code] 💡 **Suggestion**

This aliases `$_` to `$export`, but the `$item` hashtable further down (lines 215-241) still reads from `$_` throughout. It resolves correctly because those references are outside the inner `foreach`, but switching the `$item` block to use `$export` consistently would make the intent clearer and avoid future confusion about which `$_` is in scope. Optional.

docs-mslearn/toolkit/changelog.md
```diff
@@ -60,6 +60,11 @@ _Released June 2026_
 - **Fixed**
   - Removed call to Azure Classic administrators endpoint (deprecated on May 1, 2026) from Azure RBAC assignments exports ([#2142](https://github.com/microsoft/finops-toolkit/issues/2142)).
 
+### [PowerShell module](powershell/powershell-commands.md) v15
+
+- **Fixed**
+  - Fixed [Get-FinOpsCostExport](powershell/cost/get-finopscostexport.md) `-RunHistory` returning incomplete run history by querying each export's individual GET endpoint, which returns the full run history instead of the truncated history from the list endpoint ([#2063](https://github.com/microsoft/finops-toolkit/issues/2063)).
```
flanakin:
```suggestion
  - Fixed [Get-FinOpsCostExport](powershell/cost/get-finopscostexport.md) `-RunHistory` to return the complete run history ([#2063](https://github.com/microsoft/finops-toolkit/issues/2063)).
```

flanakin (approved):

src/powershell/Public/Get-FinOpsCostExport.ps1
RolandKrummenacher: Fixed in d9b7f5c8 — wrapped the per-export GET in try/catch. On a throw (network failure, non-JSON 5xx, token error) it sets $runResponse = $null so the existing else-branch falls back to the list run history instead of aborting the cmdlet. Added a test ("Should fall back to list run history when the individual GET throws").

RolandKrummenacher: Fixed in d9b7f5c8 — 429s now get a bounded wait-and-retry (3 attempts, 60s wait, mirroring Start-FinOpsCostExport) before falling back, so a throttle no longer silently drops to the truncated list history. Added a test ("Should retry on throttling and then use the full run history", Start-Sleep mocked).

src/powershell/Public/Get-FinOpsCostExport.ps1
RolandKrummenacher: Done in d9b7f5c8 — the output hashtable now reads $export consistently instead of $_.

PR #2220, opened 2026-07-30
Use case-insensitive KQL operators instead of tolower() comparisons

### LINKED ISSUE #2213: Use has instead of contains, and drop redundant tolower(), in hub KQL comparisons
## ⚠️ Problem

KQL's string comparison operators are already case-insensitive — `contains`, `startswith`, `endswith`, `has`, `=~`, `in~` (the `_cs` suffixed forms are the case-sensitive ones). Wrapping a column in `tolower()` before such a comparison adds a per-row function call and prevents the engine from using an index, with no behavioural benefit.

Across the source KQL there are 59 `tolower()` calls; 17 sit in comparison position and are redundant:

| Pattern | Sites | Equivalent |
| --- | ---: | --- |
| `tolower(X) contains '…'` | 12 | `X has '…'` |
| `tolower(X) == '…'` | 3 | `X =~ '…'` |
| `tolower(a) != tolower(b)` | 2 | `a !~ b` |

The 12 `contains` sites are all the same two expressions, in the `x_AmortizationClass` derivation:

```kusto
tolower(ResourceId) contains '/microsoft.capacity/reservationorders/'
tolower(ResourceId) contains '/microsoft.billingbenefits/savingsplanorders/'
```

across `HubSetup_v1_2.kql`, `IngestionSetup_v1_0.kql` and `IngestionSetup_v1_2.kql`. The equality sites are `costs-enriched-base.kql:52`, `HubSetup_v1_2.kql:106`, `IngestionSetup_v1_2.kql:511`, `IngestionSetup_v1_0.kql:518` and `IngestionSetup_v1_2.kql:625`.

These run over every ingested row, so the cost is proportional to dataset size.

## 🛠️ Solution

Drop `tolower()` in comparison position and use the case-insensitive operator directly. For the `ResourceId` matches, prefer `has` over `contains` — `has` uses the term index rather than scanning for a substring:

```kusto
ChargeCategory == 'Purchase' and (ResourceId has 'reservationorders'
                               or ResourceId has 'savingsplanorders'), 'Principal',
```

## ℹ️ Additional context

Verified on a hub with ~36M rows (890 `Purchase` rows): the current expression, the same without `tolower()`, `has` on the bare terms, and `has` on the full paths **all return the identical 720 rows**, matching `x_AmortizationClass == 'Principal'` exactly. No borderline cases in either direction.

Two cautions for whoever picks this up:

- `has` matches whole terms, not substrings, so it is not a blanket replacement for `contains`. It is safe at these sites because `reservationorders` and `savingsplanorders` are complete tokens under KQL's tokenizer.
- `costs-enriched-base.kql:52` compares a dynamic field (`x_SkuDetails.AHB`). `=~` expects strings, so that one may need `tostring()` rather than a plain operator swap.

The remaining ~42 `tolower()` calls write lowercased values into columns and are not affected.

Scope note: this audit covers `src/`. The `docs/deploy/*.json` templates carry the same expressions but are build output. The `Recommendations-*.json` queries were not scanned (checkout hit a path-length limit locally) and may be worth a look.

RolandKrummenacher:
## 🛠️ Description

KQL string comparison operators are already case-insensitive, so wrapping a column in `tolower()` before a comparison adds a per-row function call and prevents the engine from using the term index, with no behavioral benefit. This PR replaces all 17 `tolower()` calls in comparison position across the hub KQL scripts and the query catalog:

- `tolower(ResourceId) contains '/microsoft.capacity/reservationorders/'` (and the savings plan variant) → `ResourceId has '...'` — uses the term index instead of a lowercase + substring scan over every ingested row (12 sites in `HubSetup_v1_2.kql`, `IngestionSetup_v1_0.kql`, `IngestionSetup_v1_2.kql`).
- `tolower(x_ResourceType) == '...'` → `x_ResourceType =~ '...'` (2 sites).
- `tolower(ResourceType) != tolower(x_ResourceType)` → `ResourceType !~ x_ResourceType` (2 sites).
- `tolower(x_SkuDetails.AHB) == 'true'` → `x_SkuDetails.AHB =~ 'true'` in `costs-enriched-base.kql` — `=~` accepts the dynamic operand directly; verified across all payload shapes, see comments (1 site).

Second follow-up (3d51d3f1): swept the remaining `contains` sites where the needle is a whole term or slash-bounded phrase — `/providers/` in `parse_resourceid`, the `Windows Server BYOL` / `Azure Hybrid Benefit` / `Windows` license checks, `Trial`/`Preview` in `x_FreeReason`, and `Input`/`Output` in the AI token query (14 sites). All validated against production data with identical counts (see comments). The only `contains` left in source KQL are the two-char `ConsumedUnit` checks (`PB`/`TB`/`MB`), kept deliberately.

Follow-up commit: also switched the remaining `ResourceType !contains '/'` in the display-name guard to `ResourceType !has '/'` (both ingestion scripts). Since `/` is a separator and not a term, `has` can't use the term index there — measured identical duration/CPU and identical matches in both directions over 36.5M rows on a production hub — so this one is for consistency, not performance.

The full resource paths are kept as `has` phrases (not bare terms), so over-matching is impossible — the phrase form requires the surrounding separators exactly like `contains` did.

**Validation** (details in the issue comments): cross-tabulated old vs. new predicates on two production hubs (36.3M + 28.0M rows) — zero divergence in either direction across 37,618 matching rows; also probed 20 adversarial synthetic resource IDs and mixed-case forms with identical results. The remaining ~42 `tolower()` calls write lowercased values into columns and are intentionally untouched.

Fixes #2213

> - [x] 👍 Manually deployed + verified — predicates cross-validated on two production hubs (36.3M + 28.0M rows), zero divergence

### 📦 Deploy to test?

> - [ ] Hubs + ADX (managed)

> - [x] Changelog updated (Unreleased section)

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     paths:
       - 'src/powershell/**'
+      # KQL sources are covered by unit tests (HubsKqlOperators.Tests.ps1)
+      - 'src/templates/finops-hub/**/*.kql'
+      - 'src/queries/**/*.kql'
 jobs:
   run_pester_tests:
     name: Pester
--- a/docs-mslearn/toolkit/changelog.md
+++ b/docs-mslearn/toolkit/changelog.md
@@ -29,6 +29,8 @@ The following section lists features and enhancements that are currently in deve
 
 - **Added**
   - Added VNet and private network modes, including opt-in NAT Gateway support for private mode; NAT Gateway incurs additional cost when enabled ([#2163](https://github.com/microsoft/finops-toolkit/pull/2163)).
+- **Changed**
+  - Replaced redundant `tolower()` comparisons in hub KQL with case-insensitive operators (`has`, `=~`, `!~`) so the engine can use the term index instead of scanning every row ([#2213](https://github.com/microsoft/finops-toolkit/issues/2213)).
 
 -->
 
--- a/docs-mslearn/toolkit/changelog.md
+++ b/docs-mslearn/toolkit/changelog.md
@@ -31,6 +31,7 @@ The following section lists features and enhancements that are currently in deve
   - Added VNet and private network modes, including opt-in NAT Gateway support for private mode; NAT Gateway incurs additional cost when enabled ([#2163](https://github.com/microsoft/finops-toolkit/pull/2163)).
 - **Changed**
   - Replaced redundant `tolower()` comparisons in hub KQL with case-insensitive operators (`has`, `=~`, `!~`) so the engine can use the term index instead of scanning every row ([#2213](https://github.com/microsoft/finops-toolkit/issues/2213)).
+  - Replaced whole-term `contains` matches with `has` across hub KQL and the query catalog (resource ID paths, licensing phrases, SKU description terms) and added a per-row operator-equivalence regression harness with unit test coverage ([#2220](https://github.com/microsoft/finops-toolkit/pull/2220)).
 
 -->
 
--- a/docs-mslearn/toolkit/changelog.md
+++ b/docs-mslearn/toolkit/changelog.md
@@ -3,7 +3,7 @@ title: FinOps toolkit changelog
 description: Review the latest features and enhancements in the FinOps toolkit, including updates to FinOps hubs, Power BI reports, and more.
 author: MSBrett
 ms.author: brettwil
-ms.date: 07/29/2026
+ms.date: 07/30/2026
 ms.topic: reference
 ms.service: finops
 ms.subservice: finops-toolkit
--- a/src/powershell/Tests/Unit/HubsKqlOperators.Tests.ps1
+++ b/src/powershell/Tests/Unit/HubsKqlOperators.Tests.ps1
@@ -1,0 +1,134 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+<#
+    Regression coverage for the string operator conventions in hub KQL (#2213 / PR #2220):
+
+    1. tolower() must never appear in comparison position — KQL comparison operators are case-insensitive
+       (use =~, !~, has, in~; append _cs for case-sensitive matching).
+    2. contains is reserved for genuine substring matching (needle may be fused inside a larger token).
+       Every usage must be listed in the allowlist below with a justification.
+    3. The sites converted from contains to has stay converted (operator pins).
+    4. Every pinned needle has per-row fixtures in the executable equivalence harness
+       (Tests/assets/StringOperatorEquivalence.kql), which verifies old-vs-new behavior row by row,
+       including the whole-term vs. substring boundary cases, on any Kusto database.
+#>
+
+Describe 'HubsKqlOperators' {
+
+    BeforeDiscovery {
+        $repoRoot = (Resolve-Path "$PSScriptRoot/../../../..").Path
+        $kqlDirs = @(
+            (Join-Path $repoRoot 'src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts'),
+            (Join-Path $repoRoot 'src/queries/catalog')
+        )
+        $kqlFiles = @($kqlDirs | ForEach-Object { Get-ChildItem -Path $_ -Filter '*.kql' -ErrorAction SilentlyContinue } | ForEach-Object {
+                @{ Name = $_.Name; FullName = $_.FullName }
+            })
+
+        # Sites converted from contains/tolower() to case-insensitive term operators.
+        # Each entry pins the literal expression so a future edit cannot silently regress the operator choice.
+        $operatorPins = @(
+            @{ File = 'IngestionSetup_HubInfra.kql'; Literal = "!has '/providers/'";                                                        Needle = '/providers/' }
+            @{ File = 'HubSetup_v1_2.kql';           Literal = "has '/microsoft.capacity/reservationorders/'";                              Needle = '/microsoft.capacity/reservationorders/' }
+            @{ File = 'HubSetup_v1_2.kql';           Literal = "has '/microsoft.billingbenefits/savingsplanorders/'";                       Needle = '/microsoft.billingbenefits/savingsplanorders/' }
+            @{ File = 'IngestionSetup_v1_0.kql';     Literal = "has '/microsoft.capacity/reservationorders/'";                              Needle = '/microsoft.capacity/reservationorders/' }
+            @{ File = 'IngestionSetup_v1_2.kql';     Literal = "has '/microsoft.capacity/reservationorders/'";                              Needle = '/microsoft.capacity/reservationorders/' }
+            @{ File = 'IngestionSetup_v1_0.kql';     Literal = "!has '/'";                                                                  Needle = '/' }
+            @{ File = 'IngestionSetup_v1_2.kql';     Literal = "!has '/'";                                                                  Needle = '/' }
+            @{ File = 'HubSetup_v1_2.kql';           Literal = "has 'Windows Server BYOL'";                                                 Needle = 'Windows Server BYOL' }
+            @{ File = 'costs-enriched-base.kql';     Literal = "has 'Windows Server BYOL'";                                                 Needle = 'Windows Server BYOL' }
+            @{ File = 'HubSetup_v1_2.kql';           Literal = "has 'Azure Hybrid Benefit'";                                                Needle = 'Azure Hybrid Benefit' }
+            @{ File = 'IngestionSetup_v1_2.kql';     Literal = "has 'Azure Hybrid Benefit'";                                                Needle = 'Azure Hybrid Benefit' }
+            @{ File = 'costs-enriched-base.kql';     Literal = "has 'Azure Hybrid Benefit'";                                                Needle = 'Azure Hybrid Benefit' }
+            @{ File = 'HubSetup_v1_2.kql';           Literal = "has 'Windows'";                                                             Needle = 'Windows' }
+            @{ File = 'IngestionSetup_v1_2.kql';     Literal = "has 'Windows'";                                                             Needle = 'Windows' }
+            @{ File = 'costs-enriched-base.kql';     Literal = "has 'Windows'";                                                             Needle = 'Windows' }
+            @{ File = 'costs-enriched-base.kql';     Literal = "has 'Trial'";                                                               Needle = 'Trial' }
+            @{ File = 'costs-enriched-base.kql';     Literal = "has 'Preview'";                                                             Needle = 'Preview' }
+            @{ File = 'ai-token-usage-breakdown.kql'; Literal = 'has "Input"';                                                              Needle = 'Input' }
+            @{ File = 'ai-token-usage-breakdown.kql'; Literal = 'has "Output"';                                                             Needle = 'Output' }
+            @{ File = 'HubSetup_v1_2.kql';           Literal = "=~ 'microsoft.compute/capacityreservationgroups/capacityreservations'";     Needle = 'microsoft.compute/capacityreservationgroups/capacityreservations' }
+            @{ File = 'IngestionSetup_v1_2.kql';     Literal = "=~ 'microsoft.compute/capacityreservationgroups/capacityreservations'";     Needle = 'microsoft.compute/capacityreservationgroups/capacityreservations' }
+            @{ File = 'costs-enriched-base.kql';     Literal = "x_SkuDetails.AHB =~ 'true'";                                                Needle = 'AHB' }
+        )
+        # Resolve pin file names to full paths
+        $pinsByPath = @{}
+        foreach ($f in $kqlFiles) { $pinsByPath[$f.Name] = $f.FullName }
+        $operatorPins = @($operatorPins | ForEach-Object { $_.FullName = $pinsByPath[$_.File]; $_ })
+
+        $pinnedNeedles = @($operatorPins | ForEach-Object { $_.Needle } | Sort-Object -Unique | ForEach-Object { @{ Needle = $_ } })
+    }
+
+    BeforeAll {
+        $repoRoot = (Resolve-Path "$PSScriptRoot/../../../..").Path
+        $harnessPath = Join-Path $repoRoot 'src/powershell/Tests/assets/StringOperatorEquivalence.kql'
+        $kqlFileCount = @(
+            (Join-Path $repoRoot 'src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts'),
+            (Join-Path $repoRoot 'src/queries/catalog')
+        ) | ForEach-Object { Get-ChildItem -Path $_ -Filter '*.kql' -ErrorAction SilentlyContinue } | Measure-Object | Select-Object -ExpandProperty Count
+
+        # contains is only allowed where the needle may be fused inside a larger token (substring semantics required).
+        # To add an entry: prove the substring requirement, add per-row fixtures to StringOperatorEquivalence.kql,
+        # run the harness against a Kusto database, and document the justification here.
+        $containsAllowlist = @(
+            # ConsumedUnit carries fused unit forms (e.g. 'Mbps'); has would not match them and the needles are
+            # below the 3-character term index minimum anyway, so has offers no benefit here.
+            @{ File = 'storage-tier-distribution.kql'; Needle = 'PB' }
+            @{ File = 'storage-tier-distribution.kql'; Needle = 'TB' }
+            @{ File = 'storage-tier-distribution.kql'; Needle = 'MB' }
+        )
+    }
+
+    Context 'Case-insensitive comparisons' {
+
+        It 'Should have at least one KQL file' {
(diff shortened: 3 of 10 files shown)


RolandKrummenacher:
Follow-up on the `ResourceType !contains '/'` sites (the one place this PR previously left `contains` in place): benchmarked the alternatives on a production hub (36.5M rows, hot cache, 3 interleaved runs each, server-side stats from `.show queries`):

| Predicate | Duration (avg) | CPU (avg) |
| --- | ---: | ---: |
| `contains '/'` | 1.54 s | 3.09 s |
| `contains_cs '/'` | 1.52 s | 3.04 s |
| `has '/'` | 1.51 s | 3.08 s |
| `indexof(...) >= 0` | 2.40 s | 5.35 s |

`/` is a separator, not a term, so `has` can't use the term index here and all three non-`indexof` forms take the same scan path — differences are within noise. Correctness: all four predicates return identical matches in both directions (36,517,179 of 36,517,444 rows on `x_ResourceType`; 38,881 on `ResourceType`).

Switched to `!has '/'` anyway in 41c48b88 for consistency with the rest of the PR — one operator family for all these checks — with no behavioral or performance delta.

RolandKrummenacher:
## Synthetic edge-case validation — all 4 replacement patterns

In addition to the production-data cross-checks above, all four replacement patterns in this PR were validated against a synthetic dataset designed to hit the edge cases where the operators *could* diverge: case variants, swapped separators (`/`↔`.`), embedded/split terms, missing trailing slashes, non-string dynamic values, and empties.

**Result: 36/36 rows — old and new expression agree on every single case, in both directions.**

| # | Check | Cases | Mismatches |
| --- | --- | ---: | ---: |
| 1 | `x_AmortizationClass`: `tolower(ResourceId) contains '...'` → `ResourceId has '...'` | 14 | 0 |
| 2 | `CapacityReservationStatus`: `tolower(x) == '...'` → `x =~ '...'` | 6 | 0 |
| 3 | `ResourceType` guard: `tolower(a) != tolower(b) and a !contains '/'` → `a !~ b and a !has '/'` | 7 | 0 |
| 4 | SQL AHB flag: `tolower(x_SkuDetails.AHB) == 'true'` → `tostring(x_SkuDetails.AHB) =~ 'true'` | 9 | 0 |

Noteworthy cases:

- **`has` phrase respects separators exactly like `contains`**: `.../microsoft.capacity.reservationorders/...` (dot for slash) and `.../microsoft/capacity/reservationorders/...` (slash for dot) are rejected by both old and new — the phrase form does not degrade to bare term-adjacency matching.
- **Dynamic bool AHB**: `{"AHB": true}` (JSON bool, not string) evaluates `true` under both `tolower(...) == 'true'` and `tostring(...) =~ 'true'` — the `tostring()` wrapper introduces no behavioral change for non-string payloads (`false`, `1`, missing key, `null` all agree too).
- **Case variants** (`True`/`TRUE`, uppercase resource IDs and types) match under both forms, confirming the case-insensitive operators are exact replacements for the `tolower()` pattern.

<details>
<summary>Full 36-row result table</summary>

| Check | Case | Old | New | Match |
| 1 AmortizationClass | RI canonical | true | true | ✅ |
| 1 AmortizationClass | RI UPPERCASE | true | true | ✅ |
| 1 AmortizationClass | RI MixedCase | true | true | ✅ |
| 1 AmortizationClass | RI no trailing slash | false | false | ✅ |
| 1 AmortizationClass | RI dot for slash | false | false | ✅ |
| 1 AmortizationClass | RI slash for dot | false | false | ✅ |
| 1 AmortizationClass | RI embedded term (`xreservationordersx`) | false | false | ✅ |
| 1 AmortizationClass | RI term split (`reservation/orders`) | false | false | ✅ |
| 1 AmortizationClass | SP canonical | true | true | ✅ |
| 1 AmortizationClass | SP UPPERCASE | true | true | ✅ |
| 1 AmortizationClass | SP no trailing slash | false | false | ✅ |
| 1 AmortizationClass | SP wrong provider (`microsoft.capacity/savingsplanorders`) | false | false | ✅ |
| 1 AmortizationClass | ordinary VM resource | false | false | ✅ |
| 1 AmortizationClass | empty | false | false | ✅ |
| 2 CapacityReservationStatus | exact lowercase | true | true | ✅ |
| 2 CapacityReservationStatus | MixedCase | true | true | ✅ |
| 2 CapacityReservationStatus | UPPERCASE | true | true | ✅ |
| 2 CapacityReservationStatus | different type | false | false | ✅ |
| 2 CapacityReservationStatus | prefix only | false | false | ✅ |
| 2 CapacityReservationStatus | empty | false | false | ✅ |
| 3 ResourceType guard | display name vs internal id | true | true | ✅ |
| 3 ResourceType guard | internal id, case diff | false | false | ✅ |
| 3 ResourceType guard | internal id, identical | false | false | ✅ |
| 3 ResourceType guard | display name, case diff only | false | false | ✅ |
| 3 ResourceType guard | display name with slash | false | false | ✅ |
| 3 ResourceType guard | empty ResourceType | false | false | ✅ |
| 3 ResourceType guard | empty x_ResourceType | true | true | ✅ |
| 4 SQL AHB flag | bool true | true | true | ✅ |
| 4 SQL AHB flag | bool false | false | false | ✅ |
| 4 SQL AHB flag | string true | true | true | ✅ |
| 4 SQL AHB flag | string True | true | true | ✅ |
| 4 SQL AHB flag | string TRUE | true | true | ✅ |
| 4 SQL AHB flag | string false | false | false | ✅ |
| 4 SQL AHB flag | number 1 | false | false | ✅ |
| 4 SQL AHB flag | missing key | false | false | ✅ |
| 4 SQL AHB flag | null details | false | false | ✅ |

</details>

<details>
<summary>Reproduction query (self-contained datatable, runs on any ADX/Eventhouse database)</summary>

```kusto
let c1 = datatable(label:string, ResourceId:string) [
    'RI canonical',            '/providers/microsoft.capacity/reservationorders/00000000-aaaa/reservations/1111',
    'RI UPPERCASE',            '/PROVIDERS/MICROSOFT.CAPACITY/RESERVATIONORDERS/00000000-AAAA',
    'RI MixedCase',            '/providers/Microsoft.Capacity/ReservationOrders/00000000-aaaa',
    'RI no trailing slash',    '/providers/microsoft.capacity/reservationorders',
    'RI dot for slash',        '/providers/microsoft.capacity.reservationorders/xxx',
    'RI slash for dot',        '/providers/microsoft/capacity/reservationorders/xxx',
    'RI embedded term',        '/providers/microsoft.capacity/xreservationordersx/xxx',
    'RI term split',           '/providers/microsoft.capacity/reservation/orders/xxx',
    'SP canonical',            '/providers/microsoft.billingbenefits/savingsplanorders/2222/savingsplans/3333',
    'SP UPPERCASE',            '/PROVIDERS/MICROSOFT.BILLINGBENEFITS/SAVINGSPLANORDERS/2222',
    'SP no trailing slash',    '/providers/microsoft.billingbenefits/savingsplanorders',
    'SP wrong provider',       '/providers/microsoft.capacity/savingsplanorders/2222',
    'ordinary VM resource',    '/subscriptions/s1/resourcegroups/rg/providers/microsoft.compute/virtualmachines/vm1',
    'empty',                   '',
]
| extend old_expr = tolower(ResourceId) contains '/microsoft.capacity/reservationorders/' or tolower(ResourceId) contains '/microsoft.billingbenefits/savingsplanorders/'
| extend new_expr = ResourceId has '/microsoft.capacity/reservationorders/' or ResourceId has '/microsoft.billingbenefits/savingsplanorders/'
| extend chk = '1 AmortizationClass (contains->has)';
let c2 = datatable(label:string, xrt:string) [
    'exact lowercase',  'microsoft.compute/capacityreservationgroups/capacityreservations',
    'MixedCase',        'Microsoft.Compute/CapacityReservationGroups/CapacityReservations',
    'UPPERCASE',        'MICROSOFT.COMPUTE/CAPACITYRESERVATIONGROUPS/CAPACITYRESERVATIONS',
    'different type',   'microsoft.compute/virtualmachines',
    'prefix only',      'microsoft.compute/capacityreservationgroups',
    'empty',            '',
]
| extend old_expr = tolower(xrt) == 'microsoft.compute/capacityreservationgroups/capacityreservations'
| extend new_expr = xrt =~ 'microsoft.compute/capacityreservationgroups/capacityreservations'
| extend chk = '2 CapacityReservationStatus (==tolower -> =~)';
let c3 = datatable(label:string, rt:string, xrt:string) [
    'display name vs internal id', 'Virtual machine',                  'microsoft.compute/virtualmachines',
    'internal id, case diff',      'MICROSOFT.COMPUTE/VIRTUALMACHINES','microsoft.compute/virtualmachines',
    'internal id, identical',      'microsoft.compute/virtualmachines','microsoft.compute/virtualmachines',
    'display name, case diff',     'Virtual Machine',                  'virtual machine',
    'display name with slash',     'Disk / Snapshot',                  'microsoft.compute/snapshots',
    'empty ResourceType',          '',                                 'microsoft.compute/virtualmachines',
    'empty x_ResourceType',        'Virtual machine',                  '',
]
| extend old_expr = isnotempty(rt) and tolower(rt) != tolower(xrt) and rt !contains '/'
| extend new_expr = isnotempty(rt) and rt !~ xrt and rt !has '/'
| extend chk = '3 ResourceType guard (!=tolower,!contains -> !~,!has)';
let c4 = datatable(label:string, sku:dynamic) [
    'bool true',    dynamic({"AHB": true}),
    'bool false',   dynamic({"AHB": false}),
    'string true',  dynamic({"AHB": "true"}),
    'string True',  dynamic({"AHB": "True"}),
    'string TRUE',  dynamic({"AHB": "TRUE"}),
    'string false', dynamic({"AHB": "false"}),
    'number 1',     dynamic({"AHB": 1}),
    'missing key',  dynamic({"other": 1}),
    'null details', dynamic(null),
]
| extend old_expr = tolower(sku.AHB) == 'true'
| extend new_expr = tostring(sku.AHB) =~ 'true'
| extend chk = '4 SQL AHB flag (==tolower -> tostring =~)';
union (c1 | project chk, label, old_expr, new_expr),
      (c2 | project chk, label, old_expr, new_expr),
      (c3 | project chk, label, old_expr, new_expr),
      (c4 | project chk, label, old_expr, new_expr)
| extend match = old_expr == new_expr
| order by chk asc, label asc
```

</details>

RolandKrummenacher:
## Follow-up: full `contains` sweep (3d51d3f1)

Scanned all remaining `contains` usages in the source KQL under the rule that `contains` should only be used where genuine substring search is required. 14 more sites qualified for `has` (whole-term or slash-bounded-phrase needles); 3 were intentionally left as `contains`.

**Switched to `has`** — validated on a production hub (36.5M rows), `contains` and `has` return identical counts for every needle, in both directions:

| Needle | Column | Sites | contains | has |
| --- | --- | ---: | ---: | ---: |
| `/providers/` | `ResourceId` (parse_resourceid) | 2 | 36,517,179 | 36,517,179 |
| `Windows Server BYOL` | `ImageType` | 3 | 1,012,161 | 1,012,161 |
| `Azure Hybrid Benefit` | `x_SkuMeterSubcategory` | 3 | 52,016 | 52,016 |
| `Windows` | `x_SkuMeterSubcategory` | 4 | 62,698 | 62,698 |
| `Trial` | `x_SkuDescription` | 1 | 14,488 | 14,488 |
| `Preview` | `x_SkuDescription` | 1 | 37,532 | 37,532 |
| `Input` / `Output` | `Model` (ai-token-usage-breakdown) | 2 | 2,270 / 2,267 | 2,270 / 2,267 |

Also verified that `has` accepts a dynamic left operand directly (`x_SkuDetails.ImageType has '...'` — no `tostring()` wrapper needed; probe returned identical results for `contains`, raw `has`, and `tostring()+has`).

**Intentionally kept as `contains`** — the `ConsumedUnit contains 'PB'/'TB'/'MB'` unit-conversion checks in `storage-tier-distribution.kql`:

- Two-character needles are below the term-index minimum, so `has` takes the same scan path — no performance benefit.
- Unit strings are the one place where fused forms exist in the wild (`Mbps`, `MBps`). `contains 'MB'` matches those today; `has` would not. Whether matching `Mbps` in a storage-tier query is even *desirable* is a separate question (arguably a latent bug — a bandwidth unit would be treated as megabyte-months), but that's a behavioral decision that doesn't belong in an operator-equivalence PR. On the validation hub the counts are identical (895,755 / 561 / 0 rows for MB/TB/PB), so nothing diverges today either way.

After this sweep, those 3 unit checks are the only `contains` left in the source KQL.

RolandKrummenacher:
One more simplification (9681dc42): `=~` accepts a dynamic left operand directly, just like `has`, so the `tostring()` wrapper around the AHB comparison was unnecessary — `x_SkuDetails.AHB =~ 'true'` is now used as-is. Verified on a live cluster across all payload shapes (JSON bool `true`/`false`, string casings `true`/`True`/`TRUE`/`false`, number, missing key, `null`): the original `tolower(...) == 'true'`, the `tostring(...) =~` form, and the raw `=~` form agree on every case. The issue's caution that this site might need `tostring()` turned out to be unfounded.

MSBrett (changes requested):
Request changes for `9681dc423de5b10977d69eb13d8a32c65a4ebfce`.

The follow-up commits expand this from the original 17 redundant `tolower()` comparisons into 14 standalone `contains` -> `has` substitutions across `ai-token-usage-breakdown.kql`, `costs-enriched-base.kql`, and `IngestionSetup_HubInfra.kql`. That is a semantic filter change, not merely removal of redundant case normalization.

Please add durable, per-row regression coverage for the expanded sweep (or narrow the PR back to #2213's original scope). In particular, cover the `/providers/` guard in `parse_resourceid` and the multiword licensing phrases (`Windows Server BYOL`, `Azure Hybrid Benefit`). The existing initial 36-row query does not cover these later substitutions, and equal aggregate production counts cannot detect offsetting false positives and false negatives. Microsoft documents `has` as whole-term matching and `contains` as arbitrary-substring matching, so these boundaries need an executable regression artifact rather than only one-time PR-comment evidence.

If retaining the expanded sweep, update the changelog to distinguish the separate `contains` -> `has` optimization from the original `tolower()` comparison cleanup.

I found no security or dependency/supply-chain issue. The diff is whitespace-clean. Independent live KQL execution was not available in this review environment (no local Kusto client and the public help cluster requires authentication).

MSBrett (approved):
Approved at the current head. I am superseding my earlier request for changes: it elevated a non-required testing preference into a block without a confirmed defect. The KQL predicate changes are supported by the author''s targeted production and synthetic equivalence evidence, and this review found no functional, security, or supply-chain issue.

MSBrett (approved):

RolandKrummenacher:
@MSBrett Thanks — all three points addressed in df17ed45.

**1. Per-row production evidence (closes the offsetting-FP/FN gap).** You're right that equal aggregate counts can't rule out offsetting errors, so I re-ran the production check as row-level inequality instead: `countif(old_expression != new_expression)` for every substituted predicate pair — including the `/providers/` guard, both multiword licensing phrases, and the full ResourceType guard expression. Result on a 36,517,444-row production hub: **0 divergent rows for every pair.** A row-level `countif(old != new)` cannot offset by construction.

**2. Durable, executable per-row regression artifact.** Added `src/powershell/Tests/assets/StringOperatorEquivalence.kql` — a self-contained datatable harness (no table access; runs on any ADX or Fabric eventhouse database) that evaluates every replaced expression old-vs-new, row by row. It covers all substitutions from the sweep, specifically including the `/providers/` guard in `parse_resourceid` and the `Windows Server BYOL` / `Azure Hybrid Benefit` phrases, plus the original #2213 sites. Because `has` is whole-term and `contains` is arbitrary-substring, the harness also encodes the boundary **explicitly**: `expectedDivergent` fixtures (`XWindows Server BYOL`, `SQLAzure Hybrid Benefit`, `WindowsServer`, `Free Trials`, `InputTokens`, …) assert that the two operators *do* differ there — so a pass means both "equivalent where claimed" and "divergent exactly where documented". PASS = 0 returned rows; verified green against a live cluster before committing.

**3. CI enforcement.** `src/powershell/Tests/Unit/HubsKqlOperators.Tests.ps1` (142 tests, Pester) now guards the conventions on every PR: no `tolower()` in comparison position, `contains` restricted to a justified allowlist (the three `ConsumedUnit` unit-conversion checks, where fused forms like `Mbps` genuinely require substring semantics), literal pins for every converted site, and a sync check that every pinned needle has fixtures in the harness. `dev.yml` now also triggers the Pester job on `**/*.kql` changes so these actually run when hub KQL is edited.

**4. Changelog** now carries two separate entries: the `tolower()` comparison cleanup (#2213) and the `contains` → `has` term-matching change (this PR).

One honest caveat for the record: the harness proves *operator semantics* at the boundaries, which are stable engine behavior. What no static test can cover is future *data* drift — e.g. Microsoft introducing a fused meter name like `WindowsServer` tomorrow, which would change matching under either operator family (it would have silently over-matched under the old `contains` too). That residual risk is unchanged by this PR; it's now at least documented and fenced by the allowlist policy.

RolandKrummenacher:
#needs-review

PR #2221, opened 2026-07-30
Compare ContractedCost with a tolerance instead of exact float equality

### LINKED ISSUE #2216: ContractedCost recompute uses exact float comparison, producing no-op rewrites
## ⚠️ Problem

`IngestionSetup_v1_2.kql:592` recomputes `ContractedCost` when it disagrees with price × quantity:

```kusto
| extend ContractedCost = iff(ProviderName == 'Microsoft'
    and isnotempty(PricingQuantity) and isnotempty(x_PricingBlockSize)
    and ContractedCost != ContractedUnitPrice * PricingQuantity,
    ContractedUnitPrice * PricingQuantity, ContractedCost)
```

The guard compares two floats with `!=`. Any last-bit difference satisfies it, so the recompute fires and writes back a value differing by ~1e-7. Measured across three production hubs:

| Hub | Recorded `ContractedCost` mutations | Sub-1e-6 no-ops | Material (≥ 0.01) |
| --- | ---: | ---: | ---: |
| A | 2,424,691 | 2,422,952 (99.93%) | 1,653 |
| B | 3,426,227 | 3,425,697 (99.98%) | 509 |
| C | 7,898,060 | 7,693,047 (97.4%) | 203,122 |

**13.5 million no-op rewrites**, and every one is recorded in `x_SourceValues`, which makes that column useless for finding the changes that do matter — on hub C the 203,122 real edits sit inside 7.9M entries.

The value written is correct; only the trigger is wrong. Verified that the block size is not implicated: every material change occurs at `x_PricingBlockSize = 1`, and for rows with larger block sizes the *original* value already matched `ContractedUnitPrice * PricingQuantity`, so no block-size division is missing.

## 🛠️ Solution

Compare with a tolerance rather than exact equality. The same file already uses `0.0001` for this purpose in the savings derivations:

```kusto
and abs(ContractedCost - ContractedUnitPrice * PricingQuantity) >= 0.0001
```

That leaves the material corrections in place and drops the no-ops.

## ℹ️ Additional context

`checkReal` (line 300) also tests exact equality, which is what turns each no-op into an `x_SourceValues` entry:

```kusto
let checkReal = (column: string, oldValue: real, newValue: real) {
    iff(oldValue == newValue, dynamic({}), bag_pack(column, oldValue))
};
```

Fixing the guard at line 592 resolves this case. A tolerance in `checkReal` itself would harden the audit trail against the same pattern elsewhere.

RolandKrummenacher:
## 🛠️ Description

The `ContractedCost` recompute guard in the ingestion scripts compares two floats with `!=`:

```kusto
... and ContractedCost != ContractedUnitPrice * PricingQuantity, ...
```

Any last-bit floating-point difference satisfies the guard, so the recompute fires and writes back a value differing by ~1e-7. Measured across three production hubs: **13.5 million of 13.7 million recorded `ContractedCost` mutations (97.4–99.98%) are such no-ops**, and every one is recorded in `x_SourceValues`, burying the material corrections — on one hub 509 real edits sit inside 3.4M entries.

This PR replaces the exact comparison with the same `0.0001` tolerance the file already uses in the savings derivations:

```kusto
... and abs(ContractedCost - ContractedUnitPrice * PricingQuantity) >= 0.0001, ...
```

Material corrections (≥ 0.01) still fire exactly as before; the no-op rewrites stop, and `x_SourceValues` becomes useful for finding real changes again. Applied to both `IngestionSetup_v1_2.kql` (line 592) and `IngestionSetup_v1_0.kql` (line 489), which carries the identical guard.

Not included (possible follow-up): adding a tolerance inside `checkReal()` itself would harden the `x_SourceValues` audit trail against the same pattern elsewhere, but that changes audit semantics for all tracked columns, so it's left for a maintainer decision.

Fixes #2216

> - [x] 💪 Unit tests — per-row equivalence harness (Tests/assets/ContractedCostTolerance.kql) + HubsContractedCostGuard.Tests.ps1
> - [x] 👍 Manually deployed + verified — guard behavior measured on three production hubs (36M/28M/15M rows); verified every material change occurs at `x_PricingBlockSize = 1` and no block-size division is missing

### 📦 Deploy to test?

> - [ ] Hubs + ADX (managed)

> - [x] Changelog updated (Unreleased section)

### DIFF
--- a/docs-mslearn/toolkit/changelog.md
+++ b/docs-mslearn/toolkit/changelog.md
@@ -29,6 +29,8 @@ The following section lists features and enhancements that are currently in deve
 
 - **Added**
   - Added VNet and private network modes, including opt-in NAT Gateway support for private mode; NAT Gateway incurs additional cost when enabled ([#2163](https://github.com/microsoft/finops-toolkit/pull/2163)).
+- **Fixed**
+  - Fixed the `ContractedCost` recompute guard to compare with a tolerance instead of exact float equality, eliminating millions of no-op rewrites that polluted the `x_SourceValues` audit trail ([#2216](https://github.com/microsoft/finops-toolkit/issues/2216)).
 
 -->
 
--- a/docs-mslearn/toolkit/changelog.md
+++ b/docs-mslearn/toolkit/changelog.md
@@ -3,7 +3,7 @@ title: FinOps toolkit changelog
 description: Review the latest features and enhancements in the FinOps toolkit, including updates to FinOps hubs, Power BI reports, and more.
 author: MSBrett
 ms.author: brettwil
-ms.date: 07/29/2026
+ms.date: 07/30/2026
 ms.topic: reference
 ms.service: finops
 ms.subservice: finops-toolkit
--- a/docs-mslearn/toolkit/changelog.md
+++ b/docs-mslearn/toolkit/changelog.md
@@ -30,7 +30,7 @@ The following section lists features and enhancements that are currently in deve
 - **Added**
   - Added VNet and private network modes, including opt-in NAT Gateway support for private mode; NAT Gateway incurs additional cost when enabled ([#2163](https://github.com/microsoft/finops-toolkit/pull/2163)).
 - **Fixed**
-  - Fixed the `ContractedCost` recompute guard to compare with a tolerance instead of exact float equality, eliminating millions of no-op rewrites that polluted the `x_SourceValues` audit trail ([#2216](https://github.com/microsoft/finops-toolkit/issues/2216)).
+  - Fixed the `ContractedCost` recompute guard to compare with a null-safe tolerance instead of exact float equality, eliminating millions of no-op rewrites that polluted the `x_SourceValues` audit trail while preserving the null-cost backfill and no longer overwriting an existing cost when the unit price is missing ([#2216](https://github.com/microsoft/finops-toolkit/issues/2216)).
 
 -->
 
--- a/src/powershell/Tests/Unit/HubsContractedCostGuard.Tests.ps1
+++ b/src/powershell/Tests/Unit/HubsContractedCostGuard.Tests.ps1
@@ -1,0 +1,59 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+<#
+    Regression coverage for the ContractedCost recompute guard (#2216 / PR #2221):
+    the guard must compare with a 0.0001 tolerance, not exact float equality, so floating-point
+    noise does not trigger no-op rewrites that pollute the x_SourceValues audit trail.
+
+    Per-row behavioral coverage lives in the executable harness
+    Tests/assets/ContractedCostTolerance.kql (PASS = 0 returned rows on any Kusto database).
+#>
+
+Describe 'HubsContractedCostGuard' {
+
+    BeforeDiscovery {
+        $repoRoot = (Resolve-Path "$PSScriptRoot/../../../..").Path
+        $scriptsPath = Join-Path $repoRoot 'src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts'
+        $guardFiles = @('IngestionSetup_v1_0.kql', 'IngestionSetup_v1_2.kql') | ForEach-Object {
+            @{ Name = $_; FullName = (Join-Path $scriptsPath $_) }
+        }
+    }
+
+    BeforeAll {
+        $repoRoot = (Resolve-Path "$PSScriptRoot/../../../..").Path
+        $harnessPath = Join-Path $repoRoot 'src/powershell/Tests/assets/ContractedCostTolerance.kql'
+        $toleranceGuard = 'abs(ContractedCost - ContractedUnitPrice * PricingQuantity) >= 0.0001'
+    }
+
+    Context 'Tolerance guard' {
+
+        It 'Should compare ContractedCost with a tolerance: <Name>' -ForEach $guardFiles {
+            $content = Get-Content -Path $FullName -Raw
+            $content.Contains($toleranceGuard) | Should -BeTrue -Because 'the ContractedCost recompute must use the 0.0001 tolerance (#2216); exact float equality fires on last-bit noise and floods x_SourceValues with no-op rewrites'
+        }
+
+        It 'Should not compare ContractedCost with exact inequality: <Name>' -ForEach $guardFiles {
+            $content = Get-Content -Path $FullName -Raw
+            $content.Contains('ContractedCost != ContractedUnitPrice * PricingQuantity') | Should -BeFalse -Because 'exact float inequality was replaced by the tolerance comparison in #2216'
+        }
+    }
+
+    Context 'Equivalence harness' {
+
+        It 'Should have the tolerance harness asset' {
+            Test-Path $harnessPath | Should -BeTrue
+        }
+
+        It 'Should assert expected divergence per row' {
+            $harness = Get-Content -Path $harnessPath -Raw
+            $harness | Should -Match '\| where \(old_fires != new_fires\) != expectedDivergent' -Because 'the harness must return only rows that violate the recorded old-vs-new expectation'
+        }
+
+        It 'Should cover both column typings' {
+            $harness = Get-Content -Path $harnessPath -Raw
+            $harness | Should -Match 'qty:real' -Because 'Costs_final_v1_2 columns are real'
+            $harness | Should -Match 'qty:decimal' -Because 'Costs_final_v1_0 columns are decimal'
+        }
+    }
+}
--- a/src/powershell/Tests/Unit/HubsContractedCostGuard.Tests.ps1
+++ b/src/powershell/Tests/Unit/HubsContractedCostGuard.Tests.ps1
@@ -23,14 +23,14 @@ Describe 'HubsContractedCostGuard' {
     BeforeAll {
         $repoRoot = (Resolve-Path "$PSScriptRoot/../../../..").Path
         $harnessPath = Join-Path $repoRoot 'src/powershell/Tests/assets/ContractedCostTolerance.kql'
-        $toleranceGuard = 'abs(ContractedCost - ContractedUnitPrice * PricingQuantity) >= 0.0001'
+        $toleranceGuard = 'isnotempty(ContractedUnitPrice) and (isempty(ContractedCost) or abs(ContractedCost - ContractedUnitPrice * PricingQuantity) >= 0.0001)'
     }
 
     Context 'Tolerance guard' {
 
-        It 'Should compare ContractedCost with a tolerance: <Name>' -ForEach $guardFiles {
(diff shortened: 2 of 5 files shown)


MSBrett (changes requested):
## Confirmed functional regression

The new guards at `IngestionSetup_v1_2.kql:592` and `IngestionSetup_v1_0.kql:489` change null behavior in addition to adding a float tolerance.

I executed the old and new expressions against the public ADX help cluster (`Samples`). For a valid computed cost of `75`:

| Case | Old condition/result | New condition/result |
| --- | --- | --- |
| `ContractedCost = null`, product = `75` | `true` / `75` | `null` / `null` |
| `ContractedCost = 50`, product = `null` | `true` / `null` | `null` / `50` |
| Delta ≈ `1e-7` | `true` / recomputed | `false` / retained |
| Delta = `0.0001` | `true` / recomputed | `true` / recomputed |

The first row is reachable: `v1_2:460-462` and `v1_0:456-458` retain the original (possibly null) `ContractedCost` whenever `EffectiveCost == 0` and `BilledCost == 0`, even when `ContractedUnitPrice` and `PricingQuantity` are populated. The old guard then backfilled the missing cost; the new null-propagating expression leaves it missing. That contradicts the PR's claim that only insignificant float-trigger behavior changes.

A null-safe form that I also executed successfully is:

```kusto
and isnotempty(ContractedUnitPrice)
and (isempty(ContractedCost) or abs(ContractedCost - ContractedUnitPrice * PricingQuantity) >= 0.0001)
```

It backfills a null cost when the product is valid, preserves an existing cost when unit price is null, suppresses the reported `1e-7` noise, and still fires at the boundary. Please apply equivalent handling to both script versions and attach ADX evidence for null cost, null unit price, sub-threshold, exact-boundary, and material-delta cases.

## Verification and risk review

- `bicep build src/templates/finops-hub/main.bicep --stdout` completed with exit code 0; only two unrelated pre-existing warnings were emitted.
- No automated test references either ingestion transform. The opt-in Hubs + ADX deployment check was skipped, so the changed KQL has no reproducible CI regression coverage.
- No confirmed security or supply-chain findings. No dependencies, workflows, permissions, or deployment configuration changed.

## Requirements traceability matrix

| Requirement | Evidence | Status |
| --- | --- | --- |
| Replace exact float comparison with `0.0001` tolerance | Both changed KQL guards | Pass |
| Suppress insignificant recomputes | Live ADX `1e-7` case retained the original value | Pass |
| Preserve numeric material corrections and `>=` boundary | Live ADX boundary case recomputed | Pass |
| Preserve valid missing-cost repair behavior | Live ADX null-cost case plus reachable transform path | **Fail** |
| Update both deployed script versions | v1.0 and v1.2 changed | Pass |
| Document the fix and update `ms.date` | Unreleased changelog entry; `07/30/2026` | Pass |
| Provide reproducible behavioral coverage | No KQL tests; deployment check skipped | **Incomplete** |
| Avoid new security/supply-chain risk | Static KQL arithmetic only; dependency manifests unchanged | Pass |

**Verdict:** Changes requested before merge to make null handling explicit and verify it with executable ADX cases.

RolandKrummenacher:
## Per-row synthetic validation — committed as an executable regression harness (1ea52e35)

The tolerance change is now covered by `src/powershell/Tests/assets/ContractedCostTolerance.kql` — a self-contained datatable harness (runs on any ADX/Fabric database, no table access) that evaluates the old guard (`ContractedCost != ContractedUnitPrice * PricingQuantity`) against the new one (`abs(...) >= 0.0001`) row by row, plus `Tests/Unit/HubsContractedCostGuard.Tests.ps1` (7 tests) pinning the expression in both ingestion scripts and keeping the harness in sync.

**Result: 19/19 fixtures behave exactly as recorded — verified green against a live cluster.** `Divergent = yes` rows are the point of the fix (the old guard fired, the new one correctly doesn't); `no` rows assert everything else is unchanged.

| Typing | Case | Old fires | New fires | Divergent |
| real (v1_2) | exact match | no | no | — |
| real (v1_2) | float noise `0.1 * 3` (product `0.30000000000000004`) | **yes** | no | expected ✅ |
| real (v1_2) | float noise `1.1 * 3` | **yes** | no | expected ✅ |
| real (v1_2) | sub-tolerance `+0.00005` | **yes** | no | expected ✅ |
| real (v1_2) | at tolerance `+0.0001` (see note) | **yes** | no | expected ✅ |
| real (v1_2) | above tolerance `+0.00011` | yes | yes | — |
| real (v1_2) | material `+0.01` | yes | yes | — |
| real (v1_2) | material large | yes | yes | — |
| real (v1_2) | negative sub-tolerance | **yes** | no | expected ✅ |
| real (v1_2) | negative material | yes | yes | — |
| real (v1_2) | zero product, zero cost | no | no | — |
| real (v1_2) | zero product, dust cost `0.00005` | **yes** | no | expected ✅ |
| real (v1_2) | empty `PricingQuantity` | no | no | — |
| real (v1_2) | empty `x_PricingBlockSize` | no | no | — |
| real (v1_2) | non-Microsoft provider | no | no | — |
| decimal (v1_0) | decimal exact `0.1 * 3` | no | no | — |
| decimal (v1_0) | decimal sub-tolerance | **yes** | no | expected ✅ |
| decimal (v1_0) | decimal material `+0.01` | yes | yes | — |

Key takeaways:

- **Only the no-op class changes.** Every divergent row is a difference below 0.0001 — exactly the 13.5M no-op rewrites measured in the issue. All material corrections (`>= 0.00011`) fire identically under both guards, in both directions and for both signs.
- **The guards are untouched.** `ProviderName`, `isnotempty(PricingQuantity)`, and `isnotempty(x_PricingBlockSize)` gate the recompute exactly as before.
- **Both engine typings covered.** `Costs_final_v1_2` columns are `real`; `Costs_final_v1_0` columns are `decimal`. Decimal arithmetic is exact (`decimal(0.1) * 3 == decimal(0.3)`), so the float-noise class only exists on the v1_2 path — but the sub-tolerance class applies to both, which is why the fix targets both scripts.
- **One boundary documented in the harness:** the literal `8.0001` is `8.00009999…` in IEEE 754 and lands marginally *below* the threshold, so a difference intended as exactly 0.0001 may be skipped. That's a 1e-4-scale delta — immaterial by design, but recorded as an explicit fixture rather than hidden.

Harness usage: paste into any Kusto database — **PASS = 0 returned rows**; any returned row is either a behavioral regression or a change in engine semantics.

RolandKrummenacher:
@MSBrett Confirmed — good catch, and thank you for the null-safe form. My synthetic fixtures covered nulls in the gate columns (`PricingQuantity`, `x_PricingBlockSize`) but not in `ContractedCost`/`ContractedUnitPrice` themselves, so the `null != x` special case slipped through. Fixed in f4320297.

**Applied your expression verbatim to both scripts** (`IngestionSetup_v1_2.kql:592`, `IngestionSetup_v1_0.kql:489`):

```kusto
and isnotempty(ContractedUnitPrice)
and (isempty(ContractedCost) or abs(ContractedCost - ContractedUnitPrice * PricingQuantity) >= 0.0001)
```

**Executed ADX evidence for the requested cases** — run on a live cluster, for both column typings (`real` = v1_2, `decimal` = v1_0). `old`/`new value` is the resulting `ContractedCost` after the `iff()`:

| Case | Old fires → value | New fires → value | Verdict |
| --- | --- | --- | --- |
| null cost, product 75 (real) | yes → 75 | yes → **75** | backfill preserved |
| null cost, product 75 (decimal) | yes → 75 | yes → **75** | backfill preserved |
| null unit price, cost 50 (real) | yes → **null** | no → **50** | old destroyed the value; new retains it |
| null unit price, cost 50 (decimal) | yes → **null** | no → **50** | same |
| null price and null cost | no → null | no → null | unchanged |
| sub-threshold +0.00005 | yes → rewritten | no → retained | the fix |
| boundary +0.0001 (literal is 8.00009999… in IEEE 754) | yes → rewritten | no → retained | documented in the harness; 1e-4-scale delta, immaterial |
| above boundary +0.00011 | yes → recomputed | yes → recomputed | unchanged |
| material +0.01 | yes → recomputed | yes → recomputed | unchanged |

**Regression coverage updated** (`Tests/assets/ContractedCostTolerance.kql`, now 24 fixtures): null-cost, null-price, and null-both rows for both typings, with the two intentional divergences (`sub-tolerance` suppression and null-price retention) recorded as `expectedDivergent`. The harness compares guard conditions through `coalesce(..., false)`, which models `iff()`'s null-condition behavior exactly — necessary because `null != null` evaluates to `null` in ADX, not `false`, while `null != 75` is `true`; the naive boolean comparison can't represent that tri-state. `HubsContractedCostGuard.Tests.ps1` (8 tests) pins the null-safe expression in both scripts, rejects reintroduction of the exact inequality, and requires the null fixtures to stay present.

On the "reproducible behavioral coverage — Incomplete" line in your matrix: that was written before 1ea52e35 landed — the executable harness plus the unit test wiring are in the PR now, and the null cases from this review round are part of them.

RolandKrummenacher:
#needs-review

MSBrett (approved):
## Follow-up review: approved

The prior null-handling regression is resolved at `IngestionSetup_v1_2.kql:592` and `IngestionSetup_v1_0.kql:489`. The updated guard preserves null-cost backfill, prevents a null unit price from overwriting an existing cost, suppresses sub-tolerance noise, and retains material corrections.

### Independent verification

- Targeted Pester: **8 passed, 0 failed**.
- CI Pester executed `HubsContractedCostGuard.Tests.ps1`; the unit suite completed with **2,042 passed, 0 failed**.
- Executed the exact committed `ContractedCostTolerance.kql` against the public ADX `Samples` cluster: **0 returned failure rows**, `HasErrors = false`.
- Executed separate value assertions for 16 real/decimal cases covering null cost, null price, both null, float noise, sub-tolerance, decimal exact boundary, above-threshold/material corrections, negative values, and provider/quantity/block-size gates: **0 failed cases**.
- `bicep build src/templates/finops-hub/main.bicep --stdout`: exit code **0**; only two unrelated pre-existing warnings.

The Pester layer intentionally pins source text rather than executing KQL, but it is not the sole evidence: the committed behavioral harness was executed independently and matched the production predicate. I found no remaining correctness issue.

### Requirements traceability matrix

| Requirement | Evidence | Status |
| --- | --- | --- |
| Suppress insignificant float-noise recomputes | Production guards plus live execution of committed harness | Pass |
| Preserve material corrections | Above-threshold, positive/negative material fixtures and independent value assertions | Pass |
| Preserve null-cost backfill | Explicit `isempty(ContractedCost)` branch; real and decimal ADX cases | Pass |
| Preserve existing cost when unit price is null | `isnotempty(ContractedUnitPrice)` gate; real and decimal ADX cases | Pass |
| Cover both ingestion versions | Source assertions and real/decimal fixtures for v1.2/v1.0 | Pass |
| Provide meaningful regression coverage | CI source-contract tests plus independently executable 24-row KQL harness | Pass |
| Document the behavior | Updated Unreleased changelog and current `ms.date` | Pass |
| Avoid security/supply-chain regression | No dependency, workflow, permission, or deployment-config changes in the PR diff | Pass |

No confirmed security or supply-chain findings. No further changes requested.

PR #2223, opened 2026-07-31
Add KQL string matching and join/lookup rules to the coding guidelines

RolandKrummenacher:
## 🛠️ Description

The coding guidelines cover general practice, content, and the changelog, but say nothing about KQL — the primary language of the hub. `docs-wiki/Coding-guidelines.md`, `AGENTS.md`, and `CONTRIBUTING.md` don't mention KQL or Kusto at all today. That's part of why the redundant `tolower()` comparisons and word-search `contains` usages accumulated across the ingestion scripts until #2213 / #2220 — and why bare `join`s relying on the `innerunique` default accumulated until #2225. Nothing in the project's written standards contradicted them.

This PR writes those rules down where both humans and AI assistants will find them.

**`docs-wiki/Coding-guidelines.md`** — a new **⚡ KQL** section with two rule sets:

**String matching** (from #2213 / #2220):

- **String comparison**: why `tolower()` in comparison position is always wrong (every KQL string operator is already case-insensitive; the `_cs` forms are the case-sensitive ones), with an avoid/prefer table for the common rewrites — `tolower(x) contains` → `has`, `tolower(x) ==` → `=~`, `tolower(a) != tolower(b)` → `!~`, `indexof(...) >= 0` → `has`, and dynamic operands not needing a `tostring()` wrapper.
- **When `contains` is required**: framed as a semantic distinction (whole term vs. arbitrary substring) rather than a performance one, with the legitimate cases — needles fused inside a larger token (`ConsumedUnit contains 'MB'` also matching `Mbps`), word-stem matching, and fragments that never form a whole term. Also documents the term-index limits for punctuation and sub-3-character needles.
- **Verify before you swap**: per-row cross-tabulation on real data rather than aggregate counts (which can hide offsetting false positives and negatives), plus a pointer to the executable equivalence harness added in #2220.

**Joins and lookups** (from #2225):

- **The `innerunique` trap**: a bare `| join` deduplicates the left side on the join key — the root cause of real silent data loss fixed in #2225 (savings plan recommendations collapsing to one row per subscription). Explicit `kind=` is mandatory.
- **Canonical forms by intent**: `lookup` (+ `summarize take_any(...) by key` dedup — `distinct` is not a per-key dedup) for dimension enrichment, `leftsemi`/`inner` for filtering, `leftanti` for exclusion (never `leftouter` + `isempty()`, which duplicates before it filters), `fullouter` with key coalescing for period comparisons, and shuffle hints for large-to-large joins.
- **ARG is not ADX**: live-probing every flavor gave a matrix that differs from the ARG docs in both directions — supported: `inner`, `innerunique`, `leftouter`, `rightouter`, `fullouter`; rejected: `lookup`, all semi/anti flavors, and `in`/`!in` with a subquery. Since `leftanti` doesn't exist there, the `leftouter` + `isempty()` emulation is documented as the one acceptable exclusion form in ARG (with a key-unique right side).

**`AGENTS.md`** — both rule sets are also stated inline under Coding Standards. That file already listed Bicep, PowerShell, markdown, and commit conventions but not KQL, and it's what both Claude Code (via `CLAUDE.md`) and Copilot (via `.github/copilot-instructions.md`) load. Stating it there means an assistant sees the rule without having to follow the link.

**Merge order:** the join/lookup rules reference `KqlJoinKinds.Tests.ps1`, which lands in #2225 — merge #2225 first so the enforcement pointer resolves (the text marks it as "added in #2225" either way).

No behavior change — documentation only. The guidance is already enforced in CI: string rules by `HubsKqlOperators.Tests.ps1` (from #2220), join rules by `KqlJoinKinds.Tests.ps1` (from #2225). This PR documents the reasoning so contributors know *why* before the test tells them *no*.

> - [x] 🤏 Lint tests — `markdownlint-cli2` with the repo config: 0 issues

### 📦 Deploy to test?

> _Not applicable — documentation only._

> - [x] N/A — `docs-wiki` is contributor documentation, not part of the release changelog. It syncs to the GitHub wiki automatically on merge to `dev`.


RolandKrummenacher:
Good catch on the dynamic-operand row — it was too absolute. Qualified in 150c52de.

Verified the actual behavior on a live cluster first, and the result is a bit different from what the comment assumed: **`tostring()` is not the missing piece.** Raw and `tostring()`-wrapped comparisons return identical results for every payload shape, objects and arrays included:

| `d.v` | `tostring(d.v)` | `d.v =~ 'true'` | `tostring(d.v) =~ 'true'` | `d.v has 'true'` | `tostring(d.v) has 'true'` |
| --- | --- | --- | --- | --- | --- |
| `"true"` (string) | `true` | true | true | true | true |
| `true` (bool) | `true` | true | true | true | true |
| `1` (number) | `1` | false | false | false | false |
| `{"nested":"true"}` (object) | `{"nested":"true"}` | false | false | **true** | **true** |
| `["true"]` (array) | `["true"]` | false | false | **true** | **true** |
| missing key | ``(empty)`` | false | false | false | false |

So the real hazard isn't a missing cast — it's that a non-scalar gets compared against its **JSON serialization**, where `has` matches a value nested anywhere in the text (`{"nested":"true"} has 'true'` → `true`, while `=~` → `false`). Adding `tostring()` produces the same JSON text and the same surprising result, so recommending it would give false confidence.

The table row is now scoped to *scalar* dynamic values, with an IMPORTANT note covering the JSON-serialization behavior, why `tostring()` doesn't help, and what to use instead — extract the member you mean (`Dyn.Field.nested`), or `array_index_of()` / `set_has_element()` for arrays. Every claim in that note was executed against a cluster.

docs-wiki/Coding-guidelines.md
```diff
@@ -45,6 +47,44 @@ We adhere to the [Microsoft style guide](https://docs.microsoft.com/style-guide/
...
+
+Every KQL string operator is already case-insensitive (`has`, `contains`, `startswith`, `endswith`, `=~`, `in~`). The `_cs` suffixed forms (and `==`, `in`) are the case-sensitive ones. Wrapping a column in `tolower()` therefore adds a per-row function call and prevents the engine from using the term index, with no behavioral benefit.
+
+| ❌ Avoid | ✅ Prefer | Why |
+| --- | --- | --- |
+| `tolower(Col) contains 'term'` | `Col has 'term'` | `has` matches whole terms using the term index; `tolower()` disables it |
+| `tolower(Col) == 'value'` | `Col =~ 'value'` | `=~` is the case-insensitive equality operator |
+| `tolower(a) != tolower(b)` | `a !~ b` | One comparison instead of two per-row allocations |
+| `Col contains 'Windows'` | `Col has 'Windows'` | Whole-word needle – see below for when `contains` is right |
+| `indexof(Col, 'x') >= 0` | `Col has 'x'` | `indexof()` computes a position when you only need existence |
```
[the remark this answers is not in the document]
RolandKrummenacher: You're right, and it was worse than flagged — thanks. Fixed in 8633b8df.

Presenting this as an avoid/prefer pair contradicted the section's own point that `has` vs `contains` is a *semantic* difference, not a refactor. Verified on a live cluster, and the row was wrong on a second axis as well: `indexof()` is case-**sensitive**.

| value | `indexof(v,'Windows') >= 0` | `contains` | `contains_cs` | `has` |
| `Windows Server` | true | true | true | true |
| `WindowsServer` | true | true | true | **false** |
| `MyWindowsBuild` | true | true | true | **false** |
| `windows server` | **false** | true | false | false |

So `indexof(Col,'x') >= 0` is exactly `contains_cs` — neither `contains` nor `has`. The row now keeps the one piece of advice that does hold (don't compute a position when you only need a boolean) while naming both differences and telling contributors to pick by intent rather than swapping mechanically.

On the "also appears on line 65" pointer: that's the dynamic-operand row, which I'd already scoped to *scalar* dynamic values in 150c52de after measuring that `tostring()` changes nothing for objects/arrays — the hazard there is comparing against the JSON serialization, not a missing cast. Details in [this comment](https://github.com/microsoft/finops-toolkit/pull/2223#issuecomment-5140468139). Happy to tighten it further if that doesn't fully cover your concern.

For context on scope: `indexof` doesn't appear in any `.kql` file in the repo today, so this is guidance-only — no existing query changes.

docs-wiki/Coding-guidelines.md
```diff
@@ -45,6 +47,47 @@ We adhere to the [Microsoft style guide](https://docs.microsoft.com/style-guide/
...
+
+Hub KQL runs over every ingested row, so string matching choices show up directly in ingestion cost and query latency. Two rules cover most of it: **never wrap a column in `tolower()` to compare it**, and **reach for `has` before `contains`**.
+
+### String comparison
+
+Every KQL string operator is already case-insensitive (`has`, `contains`, `startswith`, `endswith`, `=~`, `in~`). The `_cs` suffixed forms (and `==`, `in`) are the case-sensitive ones. Wrapping a column in `tolower()` therefore adds a per-row function call and prevents the engine from using the term index, with no behavioral benefit.
+
+| ❌ Avoid | ✅ Prefer | Why |
+| --- | --- | --- |
+| `tolower(Col) contains 'term'` | `Col has 'term'` | `has` matches whole terms using the term index; `tolower()` disables it |
```
flanakin: I thought `has` was tokenized and case-sensitive. Is it not?

docs-wiki/Coding-guidelines.md
```diff
@@ -45,6 +47,47 @@ We adhere to the [Microsoft style guide](https://docs.microsoft.com/style-guide/
...
+| `tolower(Col) == 'value'` | `Col =~ 'value'` | `=~` is the case-insensitive equality operator |
```
flanakin: nit: Does `=~` do a tolower on both values? I thought `tolower(Col) == '{lowercase-literal}'` was actually faster. Do we know?

docs-wiki/Coding-guidelines.md
```diff
@@ -45,6 +47,47 @@ We adhere to the [Microsoft style guide](https://docs.microsoft.com/style-guide/
...
+| `tolower(a) != tolower(b)` | `a !~ b` | One comparison instead of two per-row allocations |
+| `Col contains 'Windows'` | `Col has 'Windows'` | Whole-word needle – see below for when `contains` is right |
```
flanakin: As far as I know, this can be a tokenization trap, aside from the case-sensitivity.

docs-wiki/Coding-guidelines.md
```diff
@@ -45,6 +47,47 @@ We adhere to the [Microsoft style guide](https://docs.microsoft.com/style-guide/
...
+| `indexof(Col, 'x') >= 0` | `Col has 'x'` *or* `Col contains 'x'` | Don't compute a position you don't need – but pick the operator by intent, not mechanically: `indexof()` is case-**sensitive** and substring-based, so it is equivalent to `contains_cs`, not to `has` |
+| `tostring(Dyn.Field) =~ 'true'` | `Dyn.Field =~ 'true'` | These operators accept a *scalar* dynamic operand directly – see the note below |
+
+Prefer `in~` over chained `=~` comparisons, and `has_any` / `has_all` over chained `has`.
```
flanakin: nit: Why not work these into the table  like the others for consistency?

docs-wiki/Coding-guidelines.md
```diff
@@ -45,6 +47,47 @@ We adhere to the [Microsoft style guide](https://docs.microsoft.com/style-guide/
...
+
+> [!IMPORTANT]
+> Only compare a dynamic field this way when it holds a **scalar** (string, bool, or number), as `x_SkuDetails.AHB` does. If the field can hold an object or an array, the comparison runs against its JSON serialization, which is rarely what you want: `has` matches a value nested anywhere inside the JSON text, so `{"nested":"true"} has 'true'` is `true` while `=~ 'true'` is `false`. Adding `tostring()` does not change this — it produces the same JSON text and the same result. Extract the value you actually mean instead (for example `Dyn.Field.nested`), or compare with `array_index_of()` / `set_has_element()` for arrays.
+
+### When `contains` is required
+
+ADX tokenizes string columns at ingestion: runs of alphanumeric characters become *terms*, and punctuation (`/`, `.`, `-`, space) are separators. `has` matches whole terms and can use the term index; `contains` scans for an arbitrary substring. That makes them **semantically different**, not just faster and slower.
```
flanakin: nit: I'd prefer to see this baked into the guidance above. I'd rather see 2 lines, one for each with the valid cases for each so all rules are cleanly listed together. Not a blocker, but this merged with what's above would tell the whole story. What's above alone can lead to bugs. This applies to both H3s here.

docs-wiki/Coding-guidelines.md
RolandKrummenacher: Half right — tokenized yes, case-sensitive no. Plain `has` is case-**insensitive**; the case-sensitive variant is `has_cs`. Verified on a live cluster:

| value | `has 'windows'` | `has 'Windows'` | `has_cs 'windows'` | `has_cs 'Windows'` | `contains 'windows'` |
| --- | --- | --- | --- | --- | --- |
| `Windows Server` | true | true | false | true | true |
| `WINDOWS SERVER` | true | true | false | false | true |
| `windows server` | true | true | true | false | true |
| `WindowsServer` | **false** | **false** | false | false | true |
| `MyWindowsBuild` | **false** | **false** | false | false | true |

Rows 1–3 show the case-insensitivity (both needle casings match every haystack casing; only `has_cs` discriminates); rows 4–5 show the tokenization (fused tokens never match `has`, regardless of case). Same split as the rest of the operator family: plain forms are case-insensitive, `_cs` forms are case-sensitive ([string operators reference](https://learn.microsoft.com/kusto/query/datatypes-string-operators)).

The section now states this explicitly up front (4024ee9a): tokenization and case-sensitivity are independent axes, and `has` sits at term-based + case-insensitive.

docs-wiki/Coding-guidelines.md
RolandKrummenacher: Measured it, since we can. Setup: 29.7M-row hub `Costs` table (production data), predicate matching 7.29M rows, query results cache disabled, variants interleaved 4× each, timings from `.show queries`; every run scanned all 37 extents, so the comparisons are apples-to-apples.

| predicate | wall clock (median) | total CPU (median) |
| --- | ---: | ---: |
| `ServiceName =~ 'virtual machines'` | 121 ms | 203 ms |
| `tolower(ServiceName) == 'virtual machines'` | **557 ms** | **1,047 ms** |
| `ServiceName == 'Virtual Machines'` | 119 ms | 219 ms |

So no — `=~` does not lower both sides per row; it compares ordinally case-insensitive without materializing lowered copies. `tolower(Col) ==` was ~4.6× slower wall-clock and ~5× CPU, which is the materialization cost showing up. The [official best practices](https://learn.microsoft.com/kusto/query/best-practices) call out this exact pair verbatim: *use `Col =~ "lowercasestring"`, don't use `tolower(Col) == "lowercasestring"`*.

The part of your recollection that does hold: when the stored casing is known and fixed, case-sensitive `==` is the documented first choice. On this table it measured the same as `=~` within noise, but it can't be slower, so the row (as of 4024ee9a) now names `==` as the preferred form when casing is known.

docs-wiki/Coding-guidelines.md
RolandKrummenacher: Agreed — as written this row contradicted the section's own "semantically different" point by presenting the swap as mechanical (same flaw you and Copilot caught on the `indexof` row). Fixed in 4024ee9a: the row is now `Col has 'Windows'` **only after verifying**, with the why-column naming the trap explicitly (`contains` also matches fused tokens like `WindowsServer`), and it points at the verification procedure which now lives in the same section. Case-sensitivity isn't a difference on this pair — both `has` and `contains` are case-insensitive (measured in the thread above) — so tokenization is the one axis the swap changes, and it's now front and center.

docs-wiki/Coding-guidelines.md
RolandKrummenacher: Done in 4024ee9a — the chained forms are now two table rows (`Col =~ 'a' or Col =~ 'b'` → `Col in~ ('a', 'b')`, and `Col has 'a' or Col has 'b'` → `Col has_any ('a', 'b')`, with `has_all` noted for conjunctions), and the loose sentence is gone.

docs-wiki/Coding-guidelines.md
RolandKrummenacher: Restructured in 4024ee9a — both H3s are gone and everything now lives in one **String matching** section that reads top-down: (1) the two engine facts that drive every rule (case-insensitivity of the plain operators; ingestion-time tokenization), (2) the two lines you asked for — one for `has`, one for `contains`, each with its valid cases — (3) the full avoid/prefer table, and (4) the verification procedure and CI enforcement inline at the end. The `contains Windows` row now carries its own "behavioral change — verify" warning, so the table alone no longer suggests a mechanical swap.

RolandKrummenacher:
Addressed the three suppressed review comments in 3c658f43:

- **`KqlJoinKinds.Tests.ps1` does not exist in this repo** (Coding-guidelines.md:113, AGENTS.md:216) — correct: the test lands in #2225 (`src/powershell/Tests/Lint/KqlJoinKinds.Tests.ps1`, 316 passing tests on that branch). Both references now say "added in #2225", the ARG callout points at "the join lint (below)" instead of a bare filename, and the PR description calls out the merge order (#2225 first).
- **Escaped pipes in the avoid/prefer table** (Coding-guidelines.md:105) — reworded every snippet so no pipe is needed at all (e.g. "`distinct Key, Col1, Col2` as the right side of a `lookup`/`join`" instead of an inline piped query). The table is now copy-paste-safe from raw markdown and reads better rendered, too.

flanakin (approved):

PR #2225, opened 2026-08-02
Fix KQL join pitfalls and adopt lookup for dimension enrichment

RolandKrummenacher:
## 📝 Changes

Follow-up to a repo-wide review of KQL `join` vs `lookup` best practices (~880 join/lookup usages examined across hub database scripts, workbooks, ARG recommendation queries, the query catalog, and docs). This PR fixes the high-severity correctness bugs, adopts `lookup` where it is the documented best practice, and adds a lint rule so bare joins cannot come back. One commit per surface:

1. **Workbooks (SavingsPlan, AHB ×2)** — correctness fixes:
   - SavingsPlan summary/details: the final `resourcecontainers` join had no `kind` and defaulted to `innerunique`, deduplicating results by subscription and **silently dropping all but one savings plan recommendation per subscription**. Now `kind=inner`.
   - AHB "VM Latest Change Last 7 days": joined the `resourcechanges` record id against `resources.id`, which never match, so the tile was always empty. Now joins `properties.targetResourceId` (lowercased both sides) with `kind=inner` so mv-expanded license changes are preserved.
   - Get-SQL-AHB-Disabled/Enabled: bare join on `VMName` dropped SQL VMs with duplicate names (innerunique) and never matched uppercase names (left original case vs right `tolower`). Now joins `properties.virtualMachineResourceId` against the VM resource id.
2. **ARG recommendation queries + finops-alerts logic app** — every bare `| join (` made an explicit `kind=inner`; `Recommendations-Microsoft-SQLVMsWithoutAHB` gets the same `virtualMachineResourceId` join-key fix as above. `docs/deploy/finops-alerts-*.json` are generated and intentionally untouched; the next release build picks up the bicep change.
3. **Hub ingestion scripts** — the v1_0/v1_2 EA transforms enriched cost rows from the open-data dimension tables (`PricingUnits`, `Regions`, `ResourceTypes`, `Services`) with `join kind=leftouter`; converted to `lookup kind=leftouter` (broadcasts the small dimension, no duplicated key columns; no downstream references to the suffixed columns existed). Also guards the `Services` enrichment against row fan-out with `summarize take_any(...) by x_ResourceType` — **this was an active bug**: Services.csv has 30 duplicate resource-type keys (up to ×31 for `microsoft.sql/locations`), so cost rows for those types were being multiplied.
4. **Query catalog** — `tagging-policy-compliance`, `storage-tier-distribution`, `macc-consumption-vs-commitment`: fact-to-small-dimension joins converted to `lookup`.
5. **Docs** — compute.md AHB examples (bare joins + the same VMName case bug, published as copy-paste guidance), commitment-coverage examples now teach `lookup`, invalid `join ... on 1 == 1` percent-of-total examples rewritten with `toscalar()`, `fullouter` examples now coalesce their join keys so baseline-only rows keep their dimension values, and networking.md examples fixed (bare joins + an invalid `resource` table name that made the idle public IP example fail outright).
6. **Lint rule** — new `Tests/Lint/KqlJoinKinds.Tests.ps1` scans every KQL-carrying surface and fails on any bare `| join` without an explicit `kind=`. The 48 remaining pre-existing bare joins (4 workbook files, all with unique left keys today) are baselined per file as a ratchet: counts can only go down (tracked in #2228). The SQL DB optimization runbook and networking.md were brought to zero in this commit.

## ✅ Validation

All changes that can run against live services were validated — three rounds documented in the comments:
- [Round 1](https://github.com/microsoft/finops-toolkit/pull/2225#issuecomment-5160056954): ARG innerunique default proven with a controlled probe; Services fan-out confirmed from open data; ingestion lookup chain executed against a hub ADX cluster; old-vs-new catalog queries equivalent on 1.16M cost rows.
- [Round 2](https://github.com/microsoft/finops-toolkit/pull/2225#issuecomment-5163368167): logic app queries, all edited doc examples (old `on 1 == 1` versions fail live with General_BadRequest; new `toscalar()` versions return correct totals), `fullouter` fixes eliminate the empty-key rows, and all 8 edited transform/HubSetup functions compile-checked against real schemas.
- [Round 3](https://github.com/microsoft/finops-toolkit/pull/2225#issuecomment-5163575777): two production-scale hub deployments (36.6M / 28.3M rows) — the pre-fix Services join is deployed in their live transforms; A/B on populated dimension tables shows 1 row → 5/4/2 rows with the old join vs 1:1 with the new lookup; no existing data corruption found (both ingest via a FOCUS path predating the enrichment). Lint suite: 161/161 passing.

## ⚠️ Known follow-ups (tracked as issues)

- #2226 — Optimization Engine join-correctness findings (anti-join emulation, missing dimension dedup, replication fan-out). Its two bare joins are fixed here so the lint starts clean.
- #2227 — SavingsPlan workbook queries have 4 joins, above ARG''s documented 3-join guidance (pre-existing; ARG accepted them in testing).
- #2228 — 48 baselined bare joins + ~75 explicit `kind=innerunique` tag-filter semi-joins across workbooks (style only; left keys unique today).
- open-data README region refresh query could use `lookup` (maintenance-only query, intentionally left as is; not issue-worthy).


RolandKrummenacher:
## ✅ Validation results (FTK test tenant + ftk-dev ADX cluster)

**ARG semantics probe** — confirmed ARG bare `join` defaults to `innerunique`: a 40-row left side with one distinct key returns **1 row** bare vs **40 rows** with `kind=inner`. The bug class this PR fixes is real.

**Services open-data fan-out (commit 3)** — confirmed against `src/open-data/Services.csv`: **30 resource types have duplicate rows**, e.g. `microsoft.apimanagement/service` ×4 and `microsoft.sql/locations` ×31. Before this PR, the v1_0/v1_2 EA ingestion transforms multiplied every cost row for those resource types by that factor. The `take_any` dedup eliminates this.

**Hub ingestion lookup chain (commit 3)** — the exact edited fragment (PricingUnits → Regions → ResourceTypes → Services ×2) executed against the `ftk-dev` Ingestion database: compiles, returns the expected schema, no duplicated key columns, non-key conflicts suffixed identically to `join` (so downstream projections are unaffected).

**Catalog queries (commit 4)** — old vs new run back-to-back on `ftk-dev` Hub DB (1,162,170 `Costs()` rows):
- `macc-consumption-vs-commitment`: byte-identical
- `storage-tier-distribution`: identical rows (order differs; query has no `order by`)
- `tagging-policy-compliance`: identical to 9 decimal places (float summation order)

**Recommendation queries + AHB/SavingsPlan workbook queries (commits 1-2)** — all executed via the ARG REST API with parameters substituted: parse and run cleanly, incl. the 4-join SavingsPlan queries (ARG accepted 4 joins). Old vs new SavingsPlan counts match in this tenant (1 recommendation per subscription, so `innerunique` happens to coincide — the probe above shows the general case). `resourcechanges.properties.targetResourceId` verified populated and resource-id-shaped on 376/376 rows, confirming the new VM Latest Change join key; the old key (`resourcechanges` record id) can never match a resource id.

**Not validated:** row-level semantics of the SQL VM / public IP / app gateway queries (test tenant has no IaaS resources — all returned 0 rows, syntax-only), and the logic app queries (validated indirectly via identical shapes + `bicep build`).

RolandKrummenacher:
## ✅ Validation round 2 — full coverage of everything testable

Completing the earlier validation, every remaining testable change has now been run against live services (ARG REST API on the FTK test tenant; `ftk-dev` ADX cluster for hub KQL):

**ARG (now all executed):**
- finops-alerts `logicApp.bicep`: both edited queries extracted from the bicep, `variables('resourcesTable')` substituted, executed — PASS.
- `compute.md` Windows-AHB + SQL-AHB examples — PASS.
- `networking.md` backendless app gateway + idle public IP examples — the idle-IP query **failed on first run and exposed a pre-existing bug**: the published example referenced table `resource` instead of `resources` (`DisallowedLogicalTableName`), so it never ran at all. Fixed in a follow-up commit; now PASS.

**Hub ADX cluster (old vs new, `Hub` DB, 1.16M cost rows):**
- `compute.md` commitment-coverage queries: join→lookup versions return identical row counts (3 / 100).
- `on 1 == 1` rewrites (`finops-hub-database-guide.md`, `ftk-database-query.md`): **old versions fail with General_BadRequest** (confirming the invalid predicate), new `toscalar()` versions pass with real data (6 / 44 rows).
- `fullouter` key-coalesce fixes: same row counts old vs new, but old had **1** (cost-spike) and **3** (service-deep-dive) rows with an empty key column; new versions have **0** — exactly the fix intent.

**Transform functions (compile-checked against real schemas):**
- All 6 edited ingestion transforms (`ActualCosts/AmortizedCosts/Costs_transform_v1_0/v1_2`) extracted from the scripts and executed with `| take 0` against the `Ingestion` DB (raw tables present) — all compile.
- Both edited `HubSetup_v1_2` functions (`CommitmentDiscountUsage_v1_2`, `Costs_v1_2`) compile against the `Hub` DB.

**Mirrors:** the three edited queries in `Compute/AHB.workbook` verified byte-identical to the live-tested `AHB/AHB.workbook` copies.

Remaining untestable in this environment: row-level semantics of the SQL VM / public IP / app gateway ARG queries (test tenant has no IaaS — all validated as executing with 0 rows).

RolandKrummenacher:
## ✅ Validation round 3 — production-scale hub deployments

Additional read-only validation against **two production FinOps hub deployments** (36.6M and 28.3M cost rows; identities withheld). Key findings:

**The Services fan-out fix (commit 3) is directly relevant to deployed hubs:**
- Both hubs have the `Services` open-data table populated with the **same 30 duplicate resource-type keys** as `src/open-data/Services.csv`.
- Both hubs have the pre-fix `join kind=leftouter (Services | ... project ...)` **deployed in their live `ActualCosts/AmortizedCosts_transform_v1_0/v1_2` functions**.
- Live A/B on one hub's populated dimension tables: the old join turns 1 input row into **5 rows** (`microsoft.security/pricings`), **4 rows** (`microsoft.web/sites`), **2 rows** (`microsoft.storage/storageaccounts`); the new lookup+`take_any` chain returns exactly 1:1 with correct enrichment values.
- On one hub, **66% of the last 30 days' cost rows** (585,710 of 880,341) carry a duplicate-key resource type — the blast radius if the legacy EA path were used.
- Two keys (`microsoft.security/pricings` ×5, `microsoft.web/sites` ×2) have genuinely different value tuples, so they would fan out **even through the newer `lookup (Services | distinct ...)` variant** — validating that this PR also converts those `distinct`-based lookups to `take_any ... by x_ResourceType`.

**No existing data corruption found in either hub:** both currently ingest via the FOCUS path, and their deployed `Costs_transform_v1_2` predates the Services enrichment entirely (probe: 0 charges with service-variant duplicates in stored data on both hubs). The buggy join is deployed but not currently exercised — this PR defuses it before it is.

**Equivalence re-confirmed on production data:** `storage-tier-distribution` old (join) vs new (lookup) on real multi-tier storage data — identical rows, 0 value mismatches.

src/powershell/Tests/Lint/KqlJoinKinds.Tests.ps1
```diff
@@ -0,0 +1,90 @@
...
+            @{ Path = 'src/powershell/Tests/assets'; Filter = '*.kql'; Recurse = $false }
+            @{ Path = 'src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Recommendations/queries'; Filter = '*.json'; Recurse = $false }
+            @{ Path = 'src/templates/finops-hub'; Filter = 'dashboard.json'; Recurse = $false }
+            @{ Path = 'src/templates/finops-alerts/modules'; Filter = 'logicApp.bicep'; Recurse = $false }
+            @{ Path = 'src/workbooks'; Filter = '*.workbook'; Recurse = $true }
+            @{ Path = 'src/workbooks'; Filter = 'workbook.json'; Recurse = $true }
+            @{ Path = 'src/optimization-engine/runbooks'; Filter = '*.ps1'; Recurse = $true }
+            @{ Path = 'src/optimization-engine/views'; Filter = '*.json'; Recurse = $true }
+            @{ Path = 'docs-mslearn/best-practices'; Filter = '*.md'; Recurse = $false }
+        )
```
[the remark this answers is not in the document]
RolandKrummenacher: Good catch — fixed in 40b82621: the docs scan target is now all of `docs-mslearn` recursively (122 markdown files). The sweep surfaced exactly the two bare joins you referenced in `customize-workbooks.md`; both are now `join kind=inner` (same left-unique `ResourceContainers` → `resources` pattern as elsewhere in this PR), so no baseline entry was needed. Lint suite: 276/276 passing.

src/powershell/Tests/Lint/KqlJoinKinds.Tests.ps1
```diff
@@ -0,0 +1,114 @@
...
+            })
+
+        # Surfaces whose KQL runs on Azure Resource Graph. Workbook files may also contain the
+        # occasional Log Analytics query (queryType 0); if one legitimately needs lookup or a
+        # semi/anti join, add a per-file allowlist analogous to the bare-join baseline.
+        $argFiles = @($scanFiles | Where-Object {
+                $_.RelPath -like 'src/workbooks/*' -or
+                $_.RelPath -like 'src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Recommendations/queries/*' -or
+                $_.RelPath -eq 'src/templates/finops-alerts/modules/logicApp.bicep'
+            })
```
flanakin, line 63: 🤖 [AI][Claude Code] ⚠️ **Should fix**

`$argFiles` only matches `src/workbooks/*`, the `Recommendations/queries/*` JSON files, and `logicApp.bicep`. But `docs-mslearn/best-practices/compute.md` and `networking.md` (scanned above for the bare-join check via the `docs-mslearn` scan target) also contain live Azure Resource Graph query examples — e.g. the `resourcecontainers | ... | join kind=inner (resources ...)` snippets this PR just edited. Those files are checked for bare joins but not for the ARG-rejected-operator rule (`lookup`, semi/anti joins), so a future edit that introduces `lookup` or `leftanti` into one of those doc examples would pass lint silently even though it would fail live against ARG.

Consider adding a condition for `docs-mslearn` ARG examples, or at minimum scoping it to the specific files/sections known to contain ARG queries (this may need a lighter-weight signal than "any file under docs-mslearn" since most docs-mslearn KQL is ADX/Log Analytics, not ARG).

src/workbooks/optimization/SavingsPlan/SavingsPlan.workbook
```diff
@@ -105,7 +105,7 @@
...
...es.extendedProperties.term),\r\nsavings=todouble(properties.extendedProperties.annualSavingsAmount),\r\nsavingsAmount = todouble(properties.extendedProperties.savingsAmount),\r\nRecommendation=tostring(properties.shortDescription.solution),  \r\ncurrency = tostring(properties.extendedProperties.savingsCurrency),\r\ndisplayQty = tostring(properties.extendedProperties.displayQty),\r\ndisplaySKU = tostring(properties.extendedProperties.displaySKU),\r\nregion = tostring(properties.extendedProperties.region),\r\nresources=tostring(properties.resourceMetadata.resourceId),  \r\nsubscription = tostring(properties.extendedProperties.subId),\r\ncommitment = tostring(properties.extendedProperties.commitment)\r\n| where term == \"{term}\" and lookbackPeriod == \"{LookBackPeriod}\"\r\n| summarize Subscriptions=dcount(resources),  \r\n  bin (sum(savings), 0.01) by subscription ,commitment ,lookbackPeriod,scope,term ,currency\r\n| order by sum_savings desc\r\n| join kind=inner (\r\nresourcecontainers\r\n| where type == 'microsoft.resources/subscriptions'\r\n| extend subscription = subscriptionId\r\n| project name,subscription\r\n) on subscription\r\n| project-away subscription1,subscription\r\n",
```
flanakin: 🤖 [AI][Claude Code] 💡 **Suggestion**

The final join (`| join kind=inner (resourcecontainers | ... ) on subscription`) attaches the subscription display name to already-summarized recommendation rows. `kind=inner` is safe today since the left side is unique per subscription after `summarize ... by subscription`, so this isn't the `innerunique` bug — but if a subscription is ever missing from `resourcecontainers` (deleted, out of scope, access change), `inner` silently drops that recommendation row from the workbook rather than just showing a blank name. `leftouter` would preserve the recommendation either way. Same pattern in the "Savings plan details" query below (line 147). Not blocking — just flagging since this PR is specifically about join-kind correctness in this file.

flanakin (commented):

**Summary:** The join/lookup correctness fixes are well-validated (live ARG probes, production-scale A/B on real hub data, before/after row-count checks) and match the described bugs. Found one lint-coverage gap and one minor join-kind suggestion — neither blocks merge.

### ⚠️ Should fix (1)

1. The new `KqlJoinKinds.Tests.ps1` lint doesn't check `docs-mslearn/*` ARG examples for operators ARG rejects.

### 💡 Suggestions (1)

1. SavingsPlan workbook's final subscription-name join could use `leftouter` instead of `inner` to guarantee recommendation rows are never dropped.

src/powershell/Tests/Lint/KqlJoinKinds.Tests.ps1
RolandKrummenacher: Good catch — and your instinct that "any file under docs-mslearn" is too coarse turned out to be load-bearing: `docs-mslearn/best-practices/compute.md` carries **both** engines. Its ARG inventory queries sit a few hundred lines away from hub cost queries that use `lookup kind=leftouter` against `Costs`/`Prices` — which this PR itself introduced as the recommended pattern. A file-level classification would have failed on those two blocks immediately.

So docs are now classified **per fenced code block** (`ea2b7ae5`): a block whose first non-comment line names an ARG table (`resources`, `resourcecontainers`, `advisorresources`, `resourcechanges`, …) is checked for `lookup` and the semi/anti flavors; every other block is left alone. A KQL query names its source table first, so the opening line is a reliable engine signal and needs no per-file allowlist.

Verified both directions, since a lint rule that never fires is worse than none:
- **Negative:** injecting `| join kind=leftanti (resources) on id` into a real ARG block in `networking.md` fails exactly one test — `docs ARG examples: docs-mslearn/best-practices/networking.md`.
- **Positive:** the two legitimate `lookup` usages in `compute.md`'s hub blocks stay green.

Full suite: 438/438 passing, 122 of them the new docs checks.

src/workbooks/optimization/SavingsPlan/SavingsPlan.workbook
RolandKrummenacher: Agreed — changed both to `kind=leftouter` in `ea2b7ae5` (summary query and the details query below it).

Your reasoning is the deciding one: this join only attaches a display name, so it should never be able to remove a recommendation. `inner` made row survival depend on the subscription being present in `resourcecontainers`, which fails exactly when the user most needs to see the recommendation — a subscription that was deleted, moved out of scope, or lost access. `project-away subscription1, subscription` still works unchanged; an unmatched row now simply renders with an empty name.

Fits the PR's theme, so it belongs here rather than in a follow-up.

flanakin (approved):

PR #2231, opened 2026-08-03
Deploy hub subnets as child resources so upgrades don't delete customer subnets

### LINKED ISSUE #2156: Bug: Hub upgrade with private networking deletes custom VNet subnets and breaks existing private network configurations
## 🐛 Problem
Upgrading FinOps hub with private networking enabled (enablePublicAccess: false) is destructive to the existing network configuration. The template does not account for an already-deployed private hub and attempts to re-create or reconcile VNet and subnet resources as if performing a fresh deployment. This breaks existing private networking in two ways:

Issue 1 — Subnet deletion: The template defines the 3 hub subnets inline inside the Microsoft.Network/virtualNetworks resource (subnets: [...]). ARM treats this inline list as the complete desired state and attempts to delete any subnet not in the list. This breaks user-added subnets (e.g. a Power BI Gateway subnet with a Microsoft.PowerPlatform/vnetaccesslinks service association link), causing the deployment to fail with InUseSubnetCannotBeDeleted.

Issue 2 — No upgrade-safe path for existing private networking: Even when the subnet error is avoided, the upgrade redeploys all private networking infrastructure (NSG, VNet, DNS zones, private endpoints) without any mechanism to detect or preserve existing configurations, putting connectivity for dependent services (Power BI Gateway, ADX client connections, peered spoke VNets) at risk during deployment.

## 👣 Repro steps
1. Deploy FinOps hub 0.12 (or any prior version) with enablePublicAccess: false
2. After initial deployment, add a custom subnet to the hub VNet — e.g. a Power BI Gateway subnet (pbi-network-data-gw-subnet) delegated to Microsoft.PowerPlatform/vnetaccesslinks for connecting Power BI reports to the ADX cluster
3. Upgrade to FinOps hub v14 by redeploying the template against the same resource group with identical hubName, enablePublicAccess: false, and virtualNetworkAddressPrefix parameters
4. Observe deployment failure at Microsoft.FinOpsHubs.Core_Infrastructure:
```
{
  "code": "InUseSubnetCannotBeDeleted",
  "message": "Subnet pbi-network-data-gw-subnet is in use by ...subnets/pbi-network-data-gw-subnet/serviceAssociationLinks/PowerPlatformSAL and cannot be deleted."
}
```
5. Even if the custom subnet is temporarily removed, re-running the upgrade redeploys the NSG, VNet, DNS zones, and private endpoints — disrupting active private endpoint connections during the deployment window.

## 🤔 Expected
Hub upgrades should be fully non-destructive to existing private networking:

Subnet preservation: The template should only manage the 3 hub-owned subnets (private-endpoint-subnet, script-subnet, dataExplorer-subnet) and leave all other subnets in the VNet untouched across upgrades — regardless of what those subnets are used for or what service association links they carry.
Idempotent private networking: Redeploying the template against an existing private hub should result in zero effective changes to the VNet, NSG rules, DNS zones, and private endpoints if the configuration has not changed — no recreation, no connectivity disruption.
Upgrade-aware deployment: Ideally, the template (or upgrade documentation) should explicitly call out the safe upgrade path for private hub deployments, including any parameters that must match the existing deployment exactly (e.g. virtualNetworkAddressPrefix).

## 📷 Screenshots

## 🔧 Environment
TODO: Complete the following (remove any that are not applicable):
- **FinOps hub version:** 12.0 to 14.0 upgrade
- **Power BI report type:**  ADX (Azure Data Explorer)
- **Private networking::** Enabled (enablePublicAccess: false)
- **Custom subnet::** pbi-network-data-gw-subnet delegated to Microsoft.PowerPlatform/vnetaccesslinks — added post-deployment for Power BI Gateway VNet integration with ADX

## ℹ️ Additional context
Root cause — subnet deletion:

In modules/Microsoft.FinOpsHubs/Core/infrastructure.bicep, the VNet is declared with an inline subnets: array:

```
resource vNet 'Microsoft.Network/virtualNetworks@2023-11-01' = if (hub.options.privateRouting) {
  properties: {
    addressSpace: { addressPrefixes: [hub.options.networkAddressPrefix] }
    subnets: subnets  // ARM deletes any subnet not declared here
  }
}
```

When ARM PUTs a VNet with an inline subnets array, it reconciles the complete subnet list and removes any subnet not in the template.

Proposed fix — standalone subnet child resources:

Remove subnets: from the VNet resource and declare each hub subnet as an independent Microsoft.Network/virtualNetworks/subnets resource. ARM only manages explicitly declared child resources, leaving all others untouched:

```
resource vNet 'Microsoft.Network/virtualNetworks@2023-11-01' = if (hub.options.privateRouting) {
  properties: {
    addressSpace: { addressPrefixes: [hub.options.networkAddressPrefix] }
    // No subnets property — managed as standalone child resources below
  }
}

resource finopsHubSubnet 'Microsoft.Network/virtualNetworks/subnets@2023-11-01' = if (hub.options.privateRouting) {
  parent: vNet
  name: 'private-endpoint-subnet'
  ...
}
// Repeat for script-subnet and dataExplorer-subnet, chained with dependsOn
// to avoid "concurrent subnet modification" ARM errors
```

Broader upgrade hardening needed:

Document the required parameters for upgrade vs. fresh deployment (especially virtualNetworkAddressPrefix must match exactly)
Consider adding upgrade documentation/script that validates existing VNet configuration before deploying
Consider detecting existing private networking resources and skipping recreation of already-healthy DNS zones and private endpoints

## 🙋‍♀️ Ask for the community
We could use your help:

2. Leave comments to help us solidify the vision — especially if you have other private networking resources (Bastion, peered spokes, custom route tables) attached to the FinOps hub VNet that were impacted.
3. If you have applied the standalone-subnet workaround in your environment, share your experience in the comments.

MSBrett:
Fixes #2156.

## Problem

The hub virtual network declared its three subnets as an inline `subnets:` array. That makes every ARM PUT authoritative for the **entire** subnet collection, so upgrading a hub either:

- silently deleted any subnet a customer had added to the hub VNet (deployment still reported `Succeeded`), or
- failed the whole deployment with `InUseSubnetCannotBeDeleted` when one of those subnets was in use.

## Fix

Subnets are now declared as standalone child resources. This is the pattern Azure documents for exactly this reason:

> Subnets should always be managed as child resources, and the **subnets** property should never be defined within the virtual network resource.
> -- [Bicep: virtual networks scenario guidance](https://learn.microsoft.com/azure/azure-resource-manager/bicep/scenarios-virtual-networks)

Note for reviewers: a different, older Learn page ([deployment modes](https://learn.microsoft.com/azure/azure-resource-manager/templates/deployment-modes)) still shows the inline pattern. The scenario guidance above is the current one, and AVM's own `avm/res/network/virtual-network` also omits `subnets`.

## Changes

- Convert the three hub subnets from an inline array to deployed child resources with identical properties.
- Serialize subnet deployment via `dependsOn` to avoid `AnotherOperationInProgress`.
- Retain the explicit `vNet` `dependsOn` for the NAT gateway. The NAT gateway is referenced by `resourceId()` string, so there is no implicit Bicep edge and ordering depends on it.
- Correct the script private endpoint to depend on the subnet it is actually created in.
- Refresh stale API versions (Network `2023-11-01` to `2025-07-01`). Both API change logs were reviewed: the delta is purely additive, no property removed and no default changed.
- Add a compiled-ARM regression suite (16 tests).
- Run the hub Bicep on pull requests so a bicep-only change actually exercises the suite.

## Validation

**Static**

- `main.bicep` compiles clean.
- 16/16 new tests pass.
- Compiled-ARM diff: 0 resources added or removed. Only apiVersion and subnet-shape changes.
- Mutation tested: reverting the Bicep fails 10/16 tests; inverting the NAT gateway ternary fails the NAT assertions. The tests actually detect the regressions they claim to.

**Live, against a real subscription**

Both modes were deployed end to end using the real compiled templates, unmodified.

`natGateway: false`

| Step | Result |
| --- | --- |
| Pre-fix deploy, then add a customer subnet, then redeploy | Customer subnet **silently deleted**, deployment reported `Succeeded` |
| Pre-fix redeploy with the customer subnet in use by a private endpoint | Failed with **`InUseSubnetCannotBeDeleted`** |
| Fixed template, identical starting state | `Succeeded`, all subnets and the customer private endpoint intact |
| Fixed template applied twice | Idempotent |

`natGateway: true`

| Step | Result |
| --- | --- |
| Fixed deploy | `Succeeded`. NAT gateway attached to the script and Data Explorer subnets, correctly **not** attached to the private endpoint subnet |
| Pre-fix with an in-use customer subnet | **`InUseSubnetCannotBeDeleted`** |
| Fixed, identical state | `Succeeded`, customer subnet and private endpoint preserved, NAT attachments preserved |
| Toggle true to false, and false to true | `Succeeded` both ways, `defaultOutboundAccess` flips correctly |

All test resource groups were deleted afterwards.

## Notes

- `defaultOutboundAccess` is set explicitly on every subnet. Network API versions released after 2026-03-31 default it to `false` for new VNets, so setting it explicitly insulates the hub from that change. The newest available version today is `2026-01-01`, before that cutoff.
- The `vNetSubnets` output is retained, so this is not a breaking change for consumers.
- Not covered here: recovery for a customer already stuck mid-failure with partially applied ARM state.

### DIFF
--- a/src/powershell/Tests/Unit/HubsPrivateNetworking.Tests.ps1
+++ b/src/powershell/Tests/Unit/HubsPrivateNetworking.Tests.ps1
@@ -1,0 +1,193 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+<#
+    Regression coverage for FinOps hub private networking (#2156).
+
+    The hub VNet must never declare its subnets through the virtual network's `subnets` property.
+    A PUT that carries that property is authoritative for the whole subnet collection, so ARM
+    deletes every subnet the template doesn't list -- which destroyed customer-added subnets on
+    upgrade and failed with InUseSubnetCannotBeDeleted when those subnets had service association
+    links. The hub now declares only the three subnets it owns, as child resources, leaving any
+    other subnet in the VNet untouched.
+
+    Assertions run against the COMPILED ARM JSON, not the Bicep source text, so a reformat or
+    rewrite that preserves behavior passes and one that changes behavior fails.
+
+    API versions are asserted as a MINIMUM, never as an exact value. Omitting `subnets` from a
+    VNet PUT only preserves existing subnets from api-version 2023-09-01 onward, so that floor is
+    the real requirement. Pinning an exact version here would block routine version upgrades.
+
+    Limitation: no static test can prove Azure's runtime behavior. Validating that an upgrade
+    preserves a customer-added subnet requires a live deployment against a real subscription.
+#>
+
+Describe 'HubsPrivateNetworking' {
+
+    BeforeDiscovery {
+        $script:bicepAvailable = [bool](Get-Command 'bicep' -ErrorAction SilentlyContinue)
+    }
+
+    BeforeAll {
+        $repoRoot = (Resolve-Path "$PSScriptRoot/../../../..").Path
+        $bicepFile = Join-Path $repoRoot 'src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Core/infrastructure.bicep'
+
+        # Behavior floor: VNet PUT without the subnets property preserves existing subnets.
+        # https://learn.microsoft.com/azure/azure-resource-manager/bicep/scenarios-virtual-networks
+        $script:minApiVersion = [datetime]::ParseExact('2023-09-01', 'yyyy-MM-dd', $null)
+
+        $script:finopsHubSubnetName = 'private-endpoint-subnet'
+        $script:scriptSubnetName = 'script-subnet'
+        $script:dataExplorerSubnetName = 'dataExplorer-subnet'
+
+        # The NAT gateway is applied through a conditional spread, which compiles to
+        # if(<enabled>, createObject('natGateway', ...), createObject()). Asserting only that
+        # "natGateways" appears somewhere would also pass if the branches were swapped, so match
+        # the natGateway object in the position the true branch occupies.
+        $script:natGatewayTrueBranch = "if\(parameters\('hub'\)\.options\.natGateway,\s*createObject\('natGateway'"
+
+        if (Get-Command 'bicep' -ErrorAction SilentlyContinue)
+        {
+            $outFile = Join-Path ([System.IO.Path]::GetTempPath()) "ftk-infra-$([guid]::NewGuid()).json"
+            $buildOutput = bicep build $bicepFile --outfile $outFile 2>&1
+            if ($LASTEXITCODE -ne 0)
+            {
+                throw "Bicep compilation failed: $($buildOutput | Out-String)"
+            }
+
+            $script:template = Get-Content -Path $outFile -Raw | ConvertFrom-Json
+            Remove-Item $outFile -Force -ErrorAction SilentlyContinue
+
+            $script:vNet = $script:template.resources.'vNet'
+            $script:subnets = @(
+                $script:template.resources.PSObject.Properties `
+                | Where-Object { $_.Value.type -eq 'Microsoft.Network/virtualNetworks/subnets' } `
+                | ForEach-Object { $_.Value }
+            )
+        }
+
+        function Get-SubnetByName([string] $Name)
+        {
+            # The compiled name is an ARM expression referencing the variable that holds the literal.
+            $variableName = switch ($Name)
+            {
+                $script:finopsHubSubnetName { 'finopsHubSubnetName' }
+                $script:scriptSubnetName { 'scriptSubnetName' }
+                $script:dataExplorerSubnetName { 'dataExplorerSubnetName' }
+            }
+            return $script:subnets | Where-Object { $_.name -match [regex]::Escape($variableName) }
+        }
+
+        # Subnets whose properties are built with a conditional spread compile to a single
+        # shallowMerge() expression string instead of an inspectable object, so those are matched
+        # as text. The literal below is what the compiler emits for the property value.
+        function Get-SubnetPropertyText($Subnet)
+        {
+            return ($Subnet.properties | ConvertTo-Json -Depth 20 -Compress)
+        }
+    }
+
+    Context 'Virtual network' -Skip:(-not $bicepAvailable) {
+
+        It 'Should not declare subnets on the virtual network (#2156)' {
+            $propertyNames = $script:vNet.properties.PSObject.Properties.Name
+            $propertyNames | Should -Not -Contain 'subnets' -Because 'a VNet PUT carrying the subnets property is authoritative for the whole collection and deletes customer-added subnets on upgrade'
+        }
+
+        It 'Should not set subnets to an empty or null value' {
+            # Azure treats "subnets": [] and "subnets": null as "delete every subnet".
+            $json = $script:vNet.properties | ConvertTo-Json -Depth 20 -Compress
+            $json | Should -Not -Match '"subnets"\s*:\s*(\[\]|null)'
+        }
+
+        It 'Should still declare the address space' {
+            $script:vNet.properties.addressSpace | Should -Not -BeNullOrEmpty
+        }
+
+        It 'Should use an api-version that preserves subnets when the property is omitted' {
+            $actual = [datetime]::ParseExact($script:vNet.apiVersion, 'yyyy-MM-dd', $null)
+            $actual | Should -BeGreaterOrEqual $script:minApiVersion -Because "omitting subnets only preserves them from 2023-09-01 onward (found $($script:vNet.apiVersion))"
+        }
+    }
+
+    Context 'Hub-owned subnets' -Skip:(-not $bicepAvailable) {
+
+        It 'Should declare exactly three subnets as child resources' {
+            $script:subnets.Count | Should -Be 3
+        }
+
+        It 'Should deploy every subnet rather than only reference it' {
+            foreach ($subnet in $script:subnets)
+            {
+                [bool]$subnet.existing | Should -BeFalse -Because "subnet '$($subnet.name)' must be deployed so the hub owns its configuration"
+            }
+        }
+
+        It 'Should use an api-version at or above the floor for every subnet' {
+            foreach ($subnet in $script:subnets)
+            {
+                $actual = [datetime]::ParseExact($subnet.apiVersion, 'yyyy-MM-dd', $null)
+                $actual | Should -BeGreaterOrEqual $script:minApiVersion
+            }
+        }
+
+        It 'Should attach the network security group to every subnet' {
+            foreach ($subnet in $script:subnets)
+            {
+                # Match the property key, not the resource type name in resourceId().
+                Get-SubnetPropertyText $subnet | Should -Match 'networkSecurityGroup[''"]' -Because "subnet '$($subnet.name)' must stay behind the hub NSG"
+            }
+        }
+
+        It 'Should serialize subnet deployment to avoid AnotherOperationInProgress' {
+            # Concurrent writes to subnets of the same VNet fail, so each subnet after the first
+            # must wait on the previous one.
+            $scriptSubnet = Get-SubnetByName $script:scriptSubnetName
+            $dataExplorerSubnet = Get-SubnetByName $script:dataExplorerSubnetName
+
+            $scriptSubnet.dependsOn | Should -Contain 'vNet::finopsHubSubnet'
+            $dataExplorerSubnet.dependsOn | Should -Contain 'vNet::scriptSubnet'
+        }
+
+        It 'Should give the private endpoint subnet a storage service endpoint' {
+            $subnet = Get-SubnetByName $script:finopsHubSubnetName
+            $subnet.properties.serviceEndpoints.service | Should -Contain 'Microsoft.Storage'
+        }
+
+        It 'Should delegate the script subnet to Azure Container Instances' {
+            $subnet = Get-SubnetByName $script:scriptSubnetName
+            Get-SubnetPropertyText $subnet | Should -Match 'Microsoft.ContainerInstance/containerGroups'
+        }
+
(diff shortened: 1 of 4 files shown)


src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Core/infrastructure.bicep
```diff
@@ -346,7 +347,7 @@ resource tablePrivateDnsZone 'Microsoft.Network/privateDnsZones@2024-06-01' = if
 // Script storage
 //------------------------------------------------------------------------------
 
-resource scriptStorageAccount 'Microsoft.Storage/storageAccounts@2022-09-01' = if (hub.options.privateRouting) {
+resource scriptStorageAccount 'Microsoft.Storage/storageAccounts@2026-04-01' = if (hub.options.privateRouting) {
```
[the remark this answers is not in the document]
MSBrett: Accepted, and thanks for catching it. That bump was unrelated scope creep on my part: this PR is a pointed subnet fix, and nothing in the `scriptStorageAccount` body needed a newer API. I checked the whole template set and it made line 350 the only reference on `2026-04-01` out of 8 storage references, with the rest on `2022-09-01` and one outlier on `2023-04-01`. The sharper problem is that `fx/hub-deploymentScript.bicep:76` reads back this same storage account as an `existing` resource on `2022-09-01`, so the drift was within a single account. Reverted to `2022-09-01` in 93da7059.

I do want these on current API versions, but as a deliberate repo-wide change rather than a one-line bump smuggled into a subnet fix, so I have tracked that separately.

To be explicit about what I did not revert: the `Microsoft.Network` bump from `2023-11-01` to `2025-07-01` stays. Those are the resources this PR restructures, and the child-resource subnet pattern depends on PUT semantics from `2023-09-01` onward, which the new tests assert as a floor rather than an exact match. Build is clean and all 16 tests still pass after the revert.

MSBrett:
Closing this PR. It failed a scope audit.

The subnet fix itself is sound and was proven against a live subscription, but this PR also contained changes that are not that fix:

- `Microsoft.Network/networkSecurityGroups` apiVersion bumped from `2023-11-01` to `2025-07-01`. An NSG is neither a subnet nor a virtual network.
- `Microsoft.Network/virtualNetworks` apiVersion bumped from `2023-11-01` to `2025-07-01`. I previously defended this as load-bearing. That was wrong: `2023-11-01` already clears the `2023-09-01` floor the fix depends on, so the bump was never required.
- `.github/workflows/dev.yml` CI trigger paths.

A storage apiVersion bump in the same vein was already caught in review and reverted. Rolling back rather than unpicking the rest in place. I will resubmit with only the subnet restructure and its regression tests.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Core/infrastructure.bicep
```diff
@@ -95,7 +34,7 @@ var subnets = !hub.options.privateRouting ? [] : [
 // Network
 //------------------------------------------------------------------------------
 
-resource nsg 'Microsoft.Network/networkSecurityGroups@2023-11-01' = if (hub.options.privateRouting) {
+resource nsg 'Microsoft.Network/networkSecurityGroups@2025-07-01' = if (hub.options.privateRouting) {
```
RolandKrummenacher, line 37: **API version `2025-07-01` loses compile-time validation.** Current Bicep (0.40.2) has no type catalog for this version, so every bumped resource now emits `BCP081` and property names are no longer validated at build time.

The fix itself only requires ≥ `2023-09-01` (the floor your own tests assert), and the `defaultOutboundAccess` insulation comes from setting the property explicitly, not from the API version. Unless there's a specific reason to be on the newest version, a version Bicep has types for (e.g. `2024-05-01`) would keep authoring safety with no behavioral difference. Not blocking — but curious if the newest version was a deliberate choice.

src/powershell/Tests/Unit/HubsPrivateNetworking.Tests.ps1
```diff
@@ -0,0 +1,193 @@
...
+    the real requirement. Pinning an exact version here would block routine version upgrades.
+
+    Limitation: no static test can prove Azure's runtime behavior. Validating that an upgrade
+    preserves a customer-added subnet requires a live deployment against a real subscription.
+#>
+
+Describe 'HubsPrivateNetworking' {
+
+    BeforeDiscovery {
+        $script:bicepAvailable = [bool](Get-Command 'bicep' -ErrorAction SilentlyContinue)
```
RolandKrummenacher, line 28: **Minor: silent skip could make the CI gate vacuous.** `windows-latest` currently ships Bicep, but if a future runner image drops it, all 16 tests skip silently while the workflow reports green — and the new `*.bicep` path trigger in `dev.yml` exists precisely to run this suite.

Consider failing instead of skipping when running in CI, e.g.:

```powershell
if ($env:CI -and -not $script:bicepAvailable) { throw 'Bicep CLI is required in CI' }
```

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Core/infrastructure.bicep
```diff
@@ -429,7 +430,7 @@ output vNetId string = !hub.options.privateRouting ? '' : vNet.id
 #disable-next-line BCP318 // Null safety warning for conditional resource access
 output vNetAddressSpace array = !hub.options.privateRouting ? [] : vNet.properties.addressSpace.addressPrefixes
 
-@description('Virtual network subnets.')
+@description('Virtual network subnets. Includes any subnets added by the customer, not only the subnets the hub manages.')
 #disable-next-line BCP318 // Null safety warning for conditional resource access
 output vNetSubnets array = !hub.options.privateRouting ? [] : vNet.properties.subnets
```
RolandKrummenacher, line 435: **Worth a sentence in the PR description:** this output's semantics change slightly — it now comes from a runtime GET, so it includes customer-added subnets and its ordering is no longer template-deterministic (previously it was the declaration order of the inline array). Nothing in this repo indexes into it positionally (I checked), and the updated `@description` covers the content change, but external consumers relying on position or count could notice. Not blocking.

src/powershell/Tests/Unit/HubsPrivateNetworking.Tests.ps1
```diff
@@ -0,0 +1,193 @@
...
+
+        $script:finopsHubSubnetName = 'private-endpoint-subnet'
+        $script:scriptSubnetName = 'script-subnet'
+        $script:dataExplorerSubnetName = 'dataExplorer-subnet'
+
+        # The NAT gateway is applied through a conditional spread, which compiles to
+        # if(<enabled>, createObject('natGateway', ...), createObject()). Asserting only that
+        # "natGateways" appears somewhere would also pass if the branches were swapped, so match
+        # the natGateway object in the position the true branch occupies.
+        $script:natGatewayTrueBranch = "if\(parameters\('hub'\)\.options\.natGateway,\s*createObject\('natGateway'"
```
RolandKrummenacher, line 47: Nit, no action needed: this regex couples the test to the exact expression the current compiler emits for the conditional spread. A future Bicep codegen change would break it — but loudly (failure, not a false pass), and the comment above already documents why positional matching is needed. Fine as is.

src/powershell/Tests/Unit/HubsPrivateNetworking.Tests.ps1
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    Context 'Virtual network' -Skip:(-not $bicepAvailable) {
+
+        It 'Should not declare subnets on the virtual network (#2156)' {
+            $propertyNames = $script:vNet.properties.PSObject.Properties.Name
+            $propertyNames | Should -Not -Contain 'subnets' -Because 'a VNet PUT carrying the subnets property is authoritative for the whole collection and deletes customer-added subnets on upgrade'
+        }
+
+        It 'Should not set subnets to an empty or null value' {
```
RolandKrummenacher, line 97: Nit: this can't fail while the previous test ('Should not declare subnets') passes, so it's technically redundant. Harmless belt-and-braces given that `[]` and `null` both mean "delete all subnets" — fine to keep.

RolandKrummenacher (approved):
The fix is correct and follows the documented Azure guidance for exactly this failure mode. I verified the riskiest claims independently rather than taking the PR description on faith:

- **Condition inheritance**: compiled the PR head with Bicep 0.40.2 — all three subnet child resources carry `condition: [parameters('hub').options.privateRouting]`, so `privateRouting: false` deployments are unaffected by the move to nested resources.
- **Dependency graph** (compiled ARM): nsg → vNet (→ natGateway) → finopsHubSubnet → scriptSubnet → dataExplorerSubnet. The script storage account depends on `scriptSubnet` (its VNet ACL rule needs the service endpoint), and the `scriptEndpoint` dependency change to `finopsHubSubnet` is a genuine fix — the endpoint lands in `private-endpoint-subnet`.
- **Cross-module ordering**: subnets now deploy *after* the vNet PUT, and subnet IDs are precomputed strings (no implicit edges), so I checked every downstream consumer. All safe: `hub-app.bicep` private endpoints run via `appRegistration` which has an explicit `dependsOn: [infrastructure]`, and Analytics consumes `core.outputs.metadata`, giving it an implicit dependency on the whole Core module.
- **Tests**: 16/16 pass against the PR head locally. Asserting against compiled ARM with an API-version floor (≥ 2023-09-01) instead of an exact pin is the right design.

Inline comments are non-blocking: the API-version question is the only one I'd like a response on before merge; the rest are minor or nits.

Beyond the code, the validation rigor here (live deploys of both failure modes and both NAT modes, mutation-testing the test suite, additive-only API delta review) is exemplary.

PR #2167, opened 2026-06-03
feat(plugins): add FinOps Toolkit agent plugins

MSBrett:
## Why

The FinOps Toolkit needed one plugin implementation that stays consistent across Claude Code, GitHub Copilot CLI, Visual Studio Code agent plugins, and the Azure SRE Agent packaging model. The previous split between `claude-plugin` and `copilot-plugin` created drift risk and duplicated maintenance.

This PR consolidates to a single source (`src/templates/agent-plugin`) and removes deprecated surface area so we can ship one canonical FinOps plugin definition.

### Consolidated plugin template

- Replaced `src/templates/claude-plugin` and `src/templates/copilot-plugin` with `src/templates/agent-plugin`.
- Added a root `plugin.json` for Copilot CLI and kept `.claude-plugin/plugin.json` for Claude.
- Updated repo discovery and marketplaces to point to `src/templates/agent-plugin`:
  - `.plugin`
  - `.github/plugin/marketplace.json`
  - `.claude-plugin/marketplace.json`

### Skill and command surface cleanup

- Deprecated and removed the `azure-cost-management` skill from the plugin payload.
- Kept a single shared `finops-toolkit` skill.
- Removed `/ftk/cost-optimization` from the command set.
- Kept 4 commands in scope:
  - `/ftk/hubs-connect`
  - `/ftk/hubs-healthCheck`
  - `/ftk/mom-report`
  - `/ftk/ytd-report`
- Scrubbed `mom-report` and `ytd-report` instructions to remove references to the removed skill.

### Build and validation wiring

- Updated `Build-AgentPlugin.ps1` for the consolidated layout.
- Kept query/docs injection behavior for packaged output:
  - `src/queries` -> `skills/finops-toolkit/references/queries`
  - `docs-mslearn` -> `skills/finops-toolkit/references/docs-mslearn`
- Updated `AgentPlugins.Tests.ps1` to validate the single consolidated plugin shape.

### Spec conformance RTMs

Validated against:

1. Claude marketplace spec: https://code.claude.com/docs/en/plugin-marketplaces
2. Copilot plugin creation spec: https://docs.github.com/en/copilot/how-tos/copilot-cli/customize-copilot/plugins-creating
3. Copilot marketplace spec: https://docs.github.com/en/copilot/how-tos/copilot-cli/customize-copilot/plugins-marketplace
4. VS Code agent plugins spec: https://code.visualstudio.com/docs/agent-customization/agent-plugins

[...]

PR #2234, opened 2026-08-04
Surface KQL language rules to the query-writing agent plugins

RolandKrummenacher:
## 🛠️ Description

Follow-up to #2220 / #2223 / #2225. The KQL rules added to the coding guidelines in #2223 live in `docs-wiki/Coding-guidelines.md` and `AGENTS.md` — contributors and repo coding agents read those, but the **shipped plugin agents never load them**. The `ftk-database-query` agent generates ad-hoc KQL against `Costs()`/`Prices()`/etc. at runtime and would reproduce exactly the pitfall classes fixed repo-wide (bare `join` defaulting to `innerunique` and silently dropping rows, `tolower()` in comparisons, `contains` where `has` is meant, `fullouter` without key coalescing). #2225 fixes the *example queries* in the skill references, but examples don't immunize newly generated queries — the rules themselves need to be in the context the agent loads.

This PR puts a distilled, runtime-relevant subset of the rules where query-writing agents already look:

- **`src/queries/finops-hub-database-guide.md`** — new **KQL language rules** subsection under *Query best practices*: case-insensitive operators (never `tolower()` in comparison position), `has` vs `contains` by matching intent, explicit `join kind=` always, `lookup` for small-dimension enrichment with a `take_any`-deduped right side, `leftanti` for exclusions, key coalescing after `fullouter`, `toscalar()` for grand totals, plus a short Azure Resource Graph caveat (no `lookup`, no semi/anti flavors, 3-join limit). This file is what the `finops-toolkit` skill's task routing loads before writing any KQL, so the rules ride along with the schema.
- **`src/templates/claude-plugin/agents/ftk-database-query.md`** — new operational guideline pointing at that section, with the three highest-impact rules inlined.

### Placement rationale

- The skills reference `src/queries` via links from both the current `agent-skills` structure **and** the restructured `agent-plugin` layout in #2167, so the rules survive the plugin move without duplication. cc @MSBrett — when #2167 lands, the same one-line pointer belongs in `ftk-database-query.agent.md`.
- Stacked logically after #2225: the guide's percent-of-total example still shows the old `join on 1 == 1` pattern on `dev`; #2225 replaces it with `toscalar()`. Until then the new rules section and that example briefly disagree — intentional, to avoid editing the same lines in two open PRs.
- No repo changelog entry: this changes agent guidance docs only; the changelog dev section is already being reshaped by #2225 and would conflict.

## 📷 Screenshots

N/A — documentation only.

## 🙋‍♀️ Ask for the community

Review whether the distilled rule set is the right size — it's loaded into agent context on every query-writing task, so it's deliberately terse compared to the full guidelines in #2223.

### DIFF
--- a/src/queries/finops-hub-database-guide.md
+++ b/src/queries/finops-hub-database-guide.md
@@ -119,9 +119,6 @@ Apply these rules to every query you write or modify. They mirror the project co
 - For exclusions ("rows with no match"), use `join kind=leftanti` — not `kind=leftouter` + `where isempty(...)`, which inflates counts when the right side has duplicate keys.
 - For two-period comparisons, `join kind=fullouter` is correct, but coalesce the key columns afterwards (`| extend Key = coalesce(Key, Key1) | project-away Key1`) or right-only rows render with empty keys.
 - For grand totals and percent-of-total, use `let Total = toscalar(...)` — never a cross join (`on 1 == 1` is not valid KQL and fails at runtime).
-
-> **Note:** One legacy example later in this guide still uses the `join ... on 1 == 1` pattern. It is replaced with `toscalar()` in [#2225](https://github.com/microsoft/finops-toolkit/pull/2225) — do not copy it.
-
 **Azure Resource Graph is different.** If you are writing ARG queries (resource inventory via `az graph query` — not the hub database), the bare-join `innerunique` trap is the same, but ARG supports *no* `lookup` and no semi/anti join flavors, and allows at most 3 joins per query. Exclusions in ARG must use the `join kind=leftouter` + `where isempty(...)` emulation with a key-unique right side.
 
 ---
--- a/src/queries/finops-hub-database-guide.md
+++ b/src/queries/finops-hub-database-guide.md
@@ -105,13 +105,13 @@ The FinOps hubs database is designed to support advanced cost and usage analytic
 
 Apply these rules to every query you write or modify. They mirror the project coding guidelines and come from real correctness bugs fixed across the toolkit ([#2189](https://github.com/microsoft/finops-toolkit/pull/2189), [#2220](https://github.com/microsoft/finops-toolkit/pull/2220), [#2225](https://github.com/microsoft/finops-toolkit/pull/2225)).
 
-**String matching**
+#### String matching
 
-- Never wrap a column in `tolower()`/`toupper()` to compare it. Every plain KQL string operator is already case-insensitive (`=~`, `has`, `contains`, `startswith`, `in~`); the `_cs` variants and `==`/`in` are the case-sensitive forms. Write `Col =~ 'value'`, not `tolower(Col) == 'value'`.
+- Never wrap a column in `tolower()`/`toupper()` to compare it. KQL's string *matching* operators are case-insensitive by default (`=~`, `has`, `contains`, `startswith`, `in~`); case-sensitive matching is what you opt into via the `_cs` variants and the equality operators `==`/`in`. Write `Col =~ 'value'`, not `tolower(Col) == 'value'`.
 - Prefer `has` when the needle is a whole term or a separator-bounded phrase (`ResourceId has '/microsoft.capacity/reservationorders/'`) — it uses the term index. Use `contains` only when the needle can be fused inside a larger token (`ConsumedUnit contains 'MB'` also matches `Mbps`).
 - Collapse operator chains: `Col in~ ('a', 'b')` instead of repeated `=~` with `or`; `has_any`/`has_all` instead of `has` chains.
 
-**Joins and lookups**
+#### Joins and lookups
 
 - Never write a bare `| join` — always state `kind=` explicitly. The default flavor is `innerunique`, which deduplicates the *left* side on the join key and silently drops rows.
 - To enrich cost rows from a small reference table, use `lookup kind=leftouter` instead of `join kind=leftouter`. It broadcasts the small side and emits no duplicated key columns.
@@ -119,6 +119,8 @@ Apply these rules to every query you write or modify. They mirror the project co
 - For exclusions ("rows with no match"), use `join kind=leftanti` — not `kind=leftouter` + `where isempty(...)`, which inflates counts when the right side has duplicate keys.
 - For two-period comparisons, `join kind=fullouter` is correct, but coalesce the key columns afterwards (`| extend Key = coalesce(Key, Key1) | project-away Key1`) or right-only rows render with empty keys.
 - For grand totals and percent-of-total, use `let Total = toscalar(...)` — never a cross join (`on 1 == 1` is not valid KQL and fails at runtime).
+
+> **Note:** One legacy example later in this guide still uses the `join ... on 1 == 1` pattern. It is replaced with `toscalar()` in [#2225](https://github.com/microsoft/finops-toolkit/pull/2225) — do not copy it.
 
 **Azure Resource Graph is different.** If you are writing ARG queries (resource inventory via `az graph query` — not the hub database), the bare-join `innerunique` trap is the same, but ARG supports *no* `lookup` and no semi/anti join flavors, and allows at most 3 joins per query. Exclusions in ARG must use the `join kind=leftouter` + `where isempty(...)` emulation with a key-unique right side.
 
@@ -716,7 +718,7 @@ The following table lists the columns produced in the `All available columns` qu
 | 2025-05-16 | 1.0     | FinOps Toolkit Team | Initial documentation                                                                                                                                                                                                                                                                                                                                                                              |
 | 2025-05-16 | 1.1     | FinOps Toolkit Team | Expanded schema, glossary, references                                                                                                                                                                                                                                                                                                                                                              |
 | 2026-05-28 | 1.2     | Sprint 3000 UAT     | Live-Hub schema audit: `Costs()`, `Prices()`, `Recommendations()` numeric columns retyped from `decimal` to `real` to match deployed Hub schema (cause of SEM0019 errors). `Recommendations()` table expanded from 12 to 20 columns to add the 8 columns present in the live schema. `x_RecommendationDate` documented as commonly-null in live Hubs (root cause of T-3000.13). |
-| 2026-08-04 | 1.3     | FinOps Toolkit Team | Added KQL language rules (case-insensitive operators, explicit join kinds, `lookup` for dimension enrichment) distilled from the project coding guidelines so query-writing agents load them alongside the schema. |
+| 2026-08-04 | 1.3     | FinOps Toolkit Team | Added KQL language rules (case-insensitive operators, explicit join kinds, `lookup` for dimension enrichment) distilled from the project coding guidelines so query-writing agents load them alongside the schema.                                                                                                                                                                                 |
 
 ---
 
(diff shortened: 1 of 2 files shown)


src/queries/finops-hub-database-guide.md
```diff
@@ -100,6 +101,27 @@ The FinOps hubs database is designed to support advanced cost and usage analytic
 - **Leverage Enrichment Columns:**
   Columns prefixed with `x_` provide additional context and enrichment for FinOps analysis.
 
+### KQL language rules
+
+Apply these rules to every query you write or modify. They mirror the project coding guidelines and come from real correctness bugs fixed across the toolkit ([#2189](https://github.com/microsoft/finops-toolkit/pull/2189), [#2220](https://github.com/microsoft/finops-toolkit/pull/2220), [#2225](https://github.com/microsoft/finops-toolkit/pull/2225)).
+
+**String matching**
+
+- Never wrap a column in `tolower()`/`toupper()` to compare it. Every plain KQL string operator is already case-insensitive (`=~`, `has`, `contains`, `startswith`, `in~`); the `_cs` variants and `==`/`in` are the case-sensitive forms. Write `Col =~ 'value'`, not `tolower(Col) == 'value'`.
```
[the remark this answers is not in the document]
RolandKrummenacher: Good catch — the opening claim overreached. Reworded in 636db16f: the rule now scopes case-insensitivity to the string *matching* operators (`=~`, `has`, `contains`, `startswith`, `in~`) and names the `_cs` variants **and the equality operators `==`/`in`** as the case-sensitive forms you opt into.

src/queries/finops-hub-database-guide.md
```diff
@@ -100,6 +101,27 @@ The FinOps hubs database is designed to support advanced cost and usage analytic
...
+
+**Joins and lookups**
+
+- Never write a bare `| join` — always state `kind=` explicitly. The default flavor is `innerunique`, which deduplicates the *left* side on the join key and silently drops rows.
+- To enrich cost rows from a small reference table, use `lookup kind=leftouter` instead of `join kind=leftouter`. It broadcasts the small side and emits no duplicated key columns.
+- Neither `join` nor `lookup` deduplicates the right side — a reference table with more than one row per key multiplies your fact rows. Guarantee one row per key with `summarize take_any(Col1), take_any(Col2) by Key` (`distinct` over multiple columns does *not* guarantee this).
+- For exclusions ("rows with no match"), use `join kind=leftanti` — not `kind=leftouter` + `where isempty(...)`, which inflates counts when the right side has duplicate keys.
+- For two-period comparisons, `join kind=fullouter` is correct, but coalesce the key columns afterwards (`| extend Key = coalesce(Key, Key1) | project-away Key1`) or right-only rows render with empty keys.
+- For grand totals and percent-of-total, use `let Total = toscalar(...)` — never a cross join (`on 1 == 1` is not valid KQL and fails at runtime).
```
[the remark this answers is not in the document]
RolandKrummenacher: Added in 636db16f — a note right under the `toscalar()` rule now flags the legacy `join ... on 1 == 1` example below as not-to-copy and points at #2225, which replaces it. The note gets removed together with the example when #2225 merges.

src/templates/claude-plugin/agents/ftk-database-query.md
```diff
@@ -129,12 +129,13 @@ The plugin provides an `azure-mcp-server` with the Kusto namespace for executing
...
-4. **Filter early**: Always scope queries to relevant time periods using `ChargePeriodStart` before aggregation.
-5. **Prefer EffectiveCost**: Use `EffectiveCost` (after discounts) as the default cost metric unless the user specifically asks for `BilledCost` (billed), `ContractedCost` (negotiated), or `ListCost` (retail).
-6. **Handle tags carefully**: Tags is a dynamic column. Extract values with `tostring(Tags['key-name'])`.
-7. **Format results**: Present query output in markdown tables with clear column headers. Include the source query and any parameter values used.
-8. **Explain the query**: When constructing KQL, explain what data you're accessing, which table function, and why.
+3. **Follow the KQL language rules**: Apply the "KQL language rules" section of `skills/finops-toolkit/references/queries/finops-hub-database-guide.md` to every query you write. In particular: never write a bare `| join` (the `innerunique` default silently drops rows — always state `kind=`), prefer `lookup` for enriching from small reference tables, and never use `tolower()` in comparisons (KQL string operators are case-insensitive — use `=~`, `has`, `in~`).
```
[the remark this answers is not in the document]
RolandKrummenacher: Leaving this as a plain path + section name intentionally: this text is agent instructions, not rendered docs — the agent opens the local file, where a markdown anchor doesn't aid navigation. The stable identifier is the section name, which the guideline already states verbatim ("the 'KQL language rules' section of ...").

src/queries/finops-hub-database-guide.md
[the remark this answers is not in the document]
RolandKrummenacher: Second thought after discussion: dropped the note again (ad35f43f). The rule two lines up already states `on 1 == 1` is invalid — that's the actual guardrail — and the note would be stale the moment #2225 replaces the example it points at. Transient cross-PR breadcrumbs in shipped docs are noise.

flanakin (approved):

src/queries/finops-hub-database-guide.md
```diff
@@ -100,6 +101,26 @@ The FinOps hubs database is designed to support advanced cost and usage analytic
...
   Columns prefixed with `x_` provide additional context and enrichment for FinOps analysis.
 
+### KQL language rules
+
+Apply these rules to every query you write or modify. They mirror the project coding guidelines and come from real correctness bugs fixed across the toolkit ([#2189](https://github.com/microsoft/finops-toolkit/pull/2189), [#2220](https://github.com/microsoft/finops-toolkit/pull/2220), [#2225](https://github.com/microsoft/finops-toolkit/pull/2225)).
+
+#### String matching
+
+- Never wrap a column in `tolower()`/`toupper()` to compare it. KQL's string *matching* operators are case-insensitive by default (`=~`, `has`, `contains`, `startswith`, `in~`); case-sensitive matching is what you opt into via the `_cs` variants and the equality operators `==`/`in`. Write `Col =~ 'value'`, not `tolower(Col) == 'value'`.
+- Prefer `has` when the needle is a whole term or a separator-bounded phrase (`ResourceId has '/microsoft.capacity/reservationorders/'`) — it uses the term index. Use `contains` only when the needle can be fused inside a larger token (`ConsumedUnit contains 'MB'` also matches `Mbps`).
```
flanakin: 🤖 [AI][claude] 🚫 **Blocker**

I want to flag this rather than assert it outright, since I can't run KQL against a live cluster to confirm — but per the Microsoft docs on [string operators](https://learn.microsoft.com/kusto/query/datatypes-string-operators), `has` matches whole *terms*, and a term is defined as "a maximal sequence of alphanumeric characters." Non-alphanumeric characters like `/` and `.` are term boundaries, not part of a term. That means `'/microsoft.capacity/reservationorders/'` isn't a single term — it breaks into `microsoft`, `capacity`, `reservationorders` — so `ResourceId has '/microsoft.capacity/reservationorders/'` won't reliably match the way this rule describes (the docs' own example, `"KustoExplorerQueryRun" has "Explorer"` → `false`, illustrates the same pitfall for substrings spanning term boundaries). If this loads into agent context to guide future query generation, worth confirming against a live cluster and either swapping in a true whole-term example or switching this example to `contains`.

**Suggestion:**

```suggestion
- Prefer `has` when the needle is a whole term (`ResourceType has 'reservations'`) — it uses the term index and is faster than a full scan. `has` only matches whole alphanumeric terms, so it can't match multi-segment strings containing separators like `/` or `.` (e.g. `ResourceId has '/microsoft.capacity/reservationorders/'` will not reliably match — use `contains` for that). Use `contains` when the needle can be fused inside a larger token or spans separators (`ConsumedUnit contains 'MB'` also matches `Mbps`; `ResourceId contains '/microsoft.capacity/reservationorders/'` matches the full path).
```

src/queries/finops-hub-database-guide.md
```diff
@@ -100,6 +101,26 @@ The FinOps hubs database is designed to support advanced cost and usage analytic
...
+- For exclusions ("rows with no match"), use `join kind=leftanti` — not `kind=leftouter` + `where isempty(...)`, which inflates counts when the right side has duplicate keys.
+- For two-period comparisons, `join kind=fullouter` is correct, but coalesce the key columns afterwards (`| extend Key = coalesce(Key, Key1) | project-away Key1`) or right-only rows render with empty keys.
+- For grand totals and percent-of-total, compute the total once with `let Total = toscalar(...)` instead of joining an aggregate subquery onto every row.
+**Azure Resource Graph is different.** If you are writing ARG queries (resource inventory via `az graph query` — not the hub database), the bare-join `innerunique` trap is the same, but ARG supports *no* `lookup` and no semi/anti join flavors, and allows at most 3 joins per query. Exclusions in ARG must use the `join kind=leftouter` + `where isempty(...)` emulation with a key-unique right side.
```
flanakin: 🤖 [AI][claude] ⚠️ **Should fix**

Missing blank line before this paragraph — on GitHub's renderer it merges into the preceding bullet list instead of standing apart as its own paragraph.

**Suggestion:**

```suggestion
- For grand totals and percent-of-total, compute the total once with `let Total = toscalar(...)` instead of joining an aggregate subquery onto every row.

**Azure Resource Graph is different.** If you are writing ARG queries (resource inventory via `az graph query` — not the hub database), the bare-join `innerunique` trap is the same, but ARG supports *no* `lookup` and no semi/anti join flavors, and allows at most 3 joins per query. Exclusions in ARG must use the `join kind=leftouter` + `where isempty(...)` emulation with a key-unique right side.
```

flanakin: 🤖 [AI][claude] 💡 **Suggestion**

Per Microsoft's ARG docs, the limit is 3 `join`/`union` operations combined (not 3 joins alone), with an exception allowing 3 cross-table joins between `Resources`/`ResourceContainers`. Recommend rewording to "allows at most 3 `join`/`union` operations combined per query" for accuracy.

flanakin (commented):
🤖 [AI][claude] **PR Review**

**Summary:** Good idea and mostly solid, technically accurate rules (verified against Microsoft's KQL docs — join/lookup semantics, case sensitivity, and dedup guidance all check out). One example in the new content contradicts how `has` actually works and should be fixed before this ships into agent context; one small formatting issue and one imprecise ARG claim are worth a look too.

### 🚫 Blockers (1)

1. The `has` example in "String matching" uses a multi-segment path as the needle, but `has` only matches whole terms (alphanumeric runs) — this example won't work as described.

### ⚠️ Should fix (1)

1. Missing blank line before the "Azure Resource Graph is different" paragraph causes it to merge into the preceding bullet list on GitHub's renderer.

### 💡 Suggestions (1)

1. The ARG "at most 3 joins per query" claim is a bit imprecise per Microsoft docs.

src/queries/finops-hub-database-guide.md
RolandKrummenacher: Verified against a live ADX cluster with a synthetic `datatable` (so independent of any table contents) — the flagged example **does match**:

```kusto
datatable(s:string) [ '/providers/microsoft.capacity/reservationorders/abc-123' ]
| extend a = s has '/microsoft.capacity/reservationorders/',  // true
         b = s has 'microsoft.capacity',                      // true
         c = s has 'capacity/reservationorders',              // true
         d = s has 'soft.capacity',                           // false — edge lands mid-term
         e = s has 'microsoft-capacity',                      // false — separator must match exactly
         f = s has 'reservation'                              // false — partial term
```

So the "whole term" rule in the docs describes single-term needles. A needle containing separators performs a case-insensitive **exact-substring match whose edges must fall on term boundaries** — which is exactly what "separator-bounded phrase" in the rule refers to. (The docs' `"KustoExplorerQueryRun" has "Explorer"` → `false` example is the *d* case above: an edge landing mid-term.)

Switching the example to `contains` would therefore weaken the guidance — `has` is both correct and index-accelerated here. I did sharpen the rule wording in de3dfa18 to state the anchored-substring semantics explicitly, so the next reader doesn't hit the same doubt.

src/queries/finops-hub-database-guide.md
RolandKrummenacher: Fixed in de3dfa18 — blank line added so the paragraph renders separately from the bullet list.

RolandKrummenacher: Reworded in de3dfa18: "documents a limit of 3 `join`/`union` operations combined per query". One nuance from live ARG probing (2026-08-03): a 4-join query was accepted, so this is a documented rather than strictly enforced limit — the new wording says so and still advises not to rely on more.

PR #2247, opened 2026-08-12
Fix PathNotFound error in msexports ETL ingestion on first run

### LINKED ISSUE #2088: [ADF] msexports_ETL_ingestion fails with PathNotFound for reservation recommendation exports
## Bug description

The `msexports_ETL_ingestion` pipeline fails with a `PathNotFound` error when processing reservation recommendation exports. The pipeline tries to read Parquet files from a path like:

```
ingestion/Recommendations/<year>/<month>/providers/microsoft.billing/billingaccounts/<id>/billingprofiles/<id>-reservationrecommendations-daily
```

The path does not exist in the storage account, causing the `Get Existing Parquet Files` activity to fail with:

```
ADLS Gen2 operation failed for: Operation returned an invalid status code 'NotFound'.
ErrorCode: 'PathNotFound'. Message: 'The specified path does not exist.'
```

This cascades into `msexports_ExecuteETL` and subsequently `ingestion_ExecuteETL` / `ingestion_ETL_dataExplorer` failures.

## Steps to reproduce

1. Deploy a FinOps hub with Fabric and recommendations enabled
2. Have a Cost Management export configured for reservation recommendations
3. Wait for the daily trigger to run the `msexports_ExecuteETL` pipeline

## Expected behavior

The pipeline should handle the case where the ingestion path does not yet exist (e.g., first run or no data from the export) gracefully, instead of failing with `PathNotFound`.

## Environment

- FinOps toolkit version: 12 (deployed), 14.0-dev (source)
- Analytics backend: Microsoft Fabric

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

flanakin:
## 🛠️ Description

The "Get Existing Parquet Files" `GetMetadata` activity fails with `PathNotFound` when the destination folder doesn't exist yet (e.g. first ingestion for a scope/dataset/month combination). Reservation recommendation exports hit this most often because their destination path adds an extra `exportName` segment, but any export type can hit it on a first run.

Fixed in both places this pattern exists: `msexports_ETL_ingestion` (`Microsoft.CostManagement/Exports/app.bicep`) and the analytics ingestion ETL pipeline (`Microsoft.FinOpsHubs/Analytics/app.bicep`). The two call sites aren't failing for the same reason, though: Exports writes to a *destination* folder that legitimately may not exist yet on first run, so this is a genuine fix there. Analytics reads a *source* folder that a manifest file just landed in, so that folder exists by construction whenever the pipeline runs — the change there is diagnostic, replacing an opaque `PathNotFound` ADLS error with the pipeline's own `IngestionFilesNotFound` error code if the folder is ever missing for some other reason (e.g. a malformed `folderPath`). It does not make a missing ingestion folder a no-op.

Fix: add `'exists'` to the `GetMetadata` activity's `fieldList`. Per [ADF's GetMetadata docs](https://learn.microsoft.com/azure/data-factory/control-flow-get-metadata-activity#supported-capabilities), specifying `exists` makes the activity return `exists: false` instead of throwing when the path is missing — "If `exists` isn't specified in the field list, the Get Metadata activity fails if the object isn't found." The downstream `Filter` activity's `items` expression now checks `.output.exists` before reading `.output.childItems`, so a missing folder resolves to an empty list instead of propagating the failure.

This is the same idiom already used elsewhere in this codebase for "path may not exist yet" scenarios: `Check Schema` in `Exports/app.bicep`, and both `GetMetadata` activities in `IngestionQueries/app.bicep` (one of which — "Get Existing Parquet Files" — already combines `exists` + `childItems` exactly like this fix does).

### Root cause verification (re: issue #2088 comment)

A prior comment on the issue did solid root-cause analysis but its line numbers had drifted and its proposed fix options were speculative (custom error handling / parent-folder existence checks / wrapping in an If Condition). Re-verified against the current code:

- `Microsoft.CostManagement/Exports/app.bicep`: `GetMetadata` activity is at lines 1090-1122, `Filter Out Current Exports` at 1123-1147 (comment's estimate of ~1090-1122 / ~1123-1147 was accurate).
- `Microsoft.FinOpsHubs/Analytics/app.bicep`: identical vulnerable pattern confirmed at lines 1668-1707 (comment said ~1667-1706, essentially correct).
- The actual cleanest fix is ADF's documented, built-in behavior for this exact case (`exists` in `fieldList`), not a custom If Condition or Web-activity workaround — and this idiom was already present elsewhere in the repo, so this change makes the two vulnerable call sites consistent with the rest of the codebase rather than introducing a new pattern.

Fixes #2088

## 📷 Screenshots

Not applicable — backend ETL pipeline fix, no UI changes.

> - [x] 🤏 Lint tests
> - [ ] 💪 Unit tests
> - [ ] 🙌 Integration tests

- `bicep build src/templates/finops-hub/modules/Microsoft.CostManagement/Exports/app.bicep --stdout` — builds cleanly
- `bicep build src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/app.bicep --stdout` — builds cleanly
- Verified compiled ARM JSON includes `"exists"` in both `fieldList` arrays
- `pwsh -Command "./src/scripts/Test-PowerShell.ps1 -Lint"` — 3418/3418 passed
- Manually verified against `pr-2247-manual`: `msexports_ETL_ingestion` ran against a scope/month with no prior parquet files. `Get Existing Parquet Files` succeeded with `exists: false` (previously would have thrown `PathNotFound`), and `Filter Out Current Exports` succeeded with an empty item list. Confirms the fix resolves the first-run failure end-to-end, not just at the Bicep/ARM level.

### 📦 Deploy to test?

> - [x] Hubs + ADX (managed)
> - [x] Hubs (manual)
> - [ ] Hubs (no data)
> - [ ] Workbooks
> - [ ] Alerts

> - [x] ➡️ Will add log in a future PR (feature branch PRs only)

### DIFF
--- a/src/templates/finops-hub/modules/Microsoft.CostManagement/Exports/app.bicep
+++ b/src/templates/finops-hub/modules/Microsoft.CostManagement/Exports/app.bicep
@@ -1129,7 +1129,7 @@ resource dataFactory 'Microsoft.DataFactory/factories@2018-06-01' existing = {
             {
               activity: 'Get Existing Parquet Files'
               dependencyConditions: [
-                'Completed'
+                'Succeeded'
               ]
             }
           ]
--- a/src/templates/finops-hub/modules/Microsoft.CostManagement/Exports/app.bicep
+++ b/src/templates/finops-hub/modules/Microsoft.CostManagement/Exports/app.bicep
@@ -1109,6 +1109,7 @@ resource dataFactory 'Microsoft.DataFactory/factories@2018-06-01' existing = {
               }
             }
             fieldList: [
+              'exists'
               'childItems'
             ]
             storeSettings: {
@@ -1135,7 +1136,7 @@ resource dataFactory 'Microsoft.DataFactory/factories@2018-06-01' existing = {
           userProperties: []
           typeProperties: {
             items: {
-              value: '@if(contains(activity(\'Get Existing Parquet Files\').output, \'childItems\'), activity(\'Get Existing Parquet Files\').output.childItems, json(\'[]\'))'
+              value: '@if(and(activity(\'Get Existing Parquet Files\').output.exists, contains(activity(\'Get Existing Parquet Files\').output, \'childItems\')), activity(\'Get Existing Parquet Files\').output.childItems, json(\'[]\'))'
               type: 'Expression'
             }
             condition: {
--- a/src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/app.bicep
+++ b/src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/app.bicep
@@ -1694,6 +1694,7 @@ resource pipeline_ExecuteIngestionETL 'Microsoft.DataFactory/factories/pipelines
             }
           }
           fieldList: [
+            'exists'
             'childItems'
           ]
           storeSettings: {
@@ -1720,7 +1721,7 @@ resource pipeline_ExecuteIngestionETL 'Microsoft.DataFactory/factories/pipelines
         userProperties: []
         typeProperties: {
           items: {
-            value: '@if(contains(activity(\'Get Existing Parquet Files\').output, \'childItems\'), activity(\'Get Existing Parquet Files\').output.childItems, json(\'[]\'))'
+            value: '@if(and(activity(\'Get Existing Parquet Files\').output.exists, contains(activity(\'Get Existing Parquet Files\').output, \'childItems\')), activity(\'Get Existing Parquet Files\').output.childItems, json(\'[]\'))'
             type: 'Expression'
           }
           condition: {


src/templates/finops-hub/modules/Microsoft.CostManagement/Exports/app.bicep
```diff
@@ -1135,7 +1136,7 @@ resource dataFactory 'Microsoft.DataFactory/factories@2018-06-01' existing = {
           userProperties: []
           typeProperties: {
             items: {
-              value: '@if(contains(activity(\'Get Existing Parquet Files\').output, \'childItems\'), activity(\'Get Existing Parquet Files\').output.childItems, json(\'[]\'))'
+              value: '@if(and(activity(\'Get Existing Parquet Files\').output.exists, contains(activity(\'Get Existing Parquet Files\').output, \'childItems\')), activity(\'Get Existing Parquet Files\').output.childItems, json(\'[]\'))'
```
RolandKrummenacher, line 1139: **Unguarded `.output.exists` breaks the `Completed` dependency tolerance.**

`Filter Out Current Exports` depends on `Get Existing Parquet Files` with `dependencyConditions: ['Completed']` (not `Succeeded`) - it is deliberately designed to run even when the GetMetadata activity *fails*. That is exactly why the original expression used only `contains(output, 'childItems')`: a failed activity returns an output object without that property, so the expression degraded to `json('[]')` and the pipeline continued. (Both the `Completed` condition and the `contains` guard were introduced together in #1512.)

Adding `exists` to the fieldList stops the `PathNotFound` failure, but the activity can still fail for other reasons - transient ADLS Gen2 5xx/throttling, an RBAC error on the ingestion container, or the 12h `timeout`. In those cases `output` has no `exists` property either, so `and(activity('Get Existing Parquet Files').output.exists, ...)` now fails evaluation ("the expression cannot be evaluated because property 'exists' doesn't exist"), the Filter activity fails, `For Each Old File` (Succeeded dependency) is skipped, and `msexports_ETL_ingestion` fails - where it previously degraded gracefully to an empty file list.

The `output.exists` term is also redundant: when `exists` is false, GetMetadata returns no `childItems`, so the pre-existing `contains(output, 'childItems')` check alone already yields `json('[]')`. The simplest fix is to keep only the `fieldList` addition and leave this expression unchanged (or, if you want the explicit check, guard it with `contains(activity('Get Existing Parquet Files').output, 'exists')` first).

Note the precedent cited in the PR description (`IngestionQueries/app.bicep:363`) is safe only because that Filter uses a `Succeeded` dependency, so it never evaluates against a failed-activity output.

```suggestion
              value: '@if(contains(activity(\'Get Existing Parquet Files\').output, \'childItems\'), activity(\'Get Existing Parquet Files\').output.childItems, json(\'[]\'))'
```

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/app.bicep
```diff
@@ -1694,6 +1694,7 @@ resource pipeline_ExecuteIngestionETL 'Microsoft.DataFactory/factories/pipelines
             }
           }
           fieldList: [
+            'exists'
```
RolandKrummenacher, line 1697: **This changes which error the user gets, but the pipeline still fails on a missing folder.**

When `ingestion_ExecuteETL` runs for a folderPath whose ingestion folder does not exist yet, `Get Existing Parquet Files` now succeeds with `exists: false`, `Filter Out Folders` yields an empty array, and then `If No Files` (below, `@equals(length(activity('Filter Out Folders').output.Value), 0)`) fires the `Files Not Found` Fail activity with errorCode `IngestionFilesNotFound`.

So the cascade into `ingestion_ExecuteETL` described in #2088 is not actually resolved here - `PathNotFound` is just replaced by `IngestionFilesNotFound`, whose message ("Please confirm the folder path is the full path, including the \"ingestion\" container and not starting with or ending with a slash") points the operator at a path-formatting problem that is not the real cause.

If a first-run / no-data folder should be a no-op, `If No Files` needs to distinguish "folder absent" (`exists == false`) from "folder present but empty / path malformed". If failing loudly is still the intent here, it would be worth saying so in the PR description, since the current text implies this hunk fixes the failure.

src/templates/finops-hub/modules/Microsoft.CostManagement/Exports/app.bicep
RolandKrummenacher: Agreed — this is the better direction, and my suggestion was wrong about what the `Completed` condition was buying us. Resolving.

I traced the dependency graph in `msexports_ETL_ingestion` to check what actually changes on a genuine GetMetadata failure, since my objection assumed the old behaviour was graceful degradation:

- `Convert to Parquet` depends on `Set Destination Path` / `Load Schema Mappings` / `Set Additional Columns` — **not** on `Filter Out Current Exports`, so the conversion path is independent of GetMetadata.
- `For Each Old File` depends on both `Convert to Parquet` (Succeeded) and `Filter Out Current Exports` (Succeeded).
- `Read Hub Config` / `If Not Retaining Exports` / `Delete Source File` hang off `Convert to Parquet`.

So on a GetMetadata failure, before and after this commit:

| | `Completed` (before) | `Succeeded` (now) |
|---|---|---|
| New parquet written | yes | yes |
| Superseded files deleted | **no** (empty list) | **no** (activity skipped) |
| Source export cleaned up | yes | yes |
| Pipeline outcome | **Succeeded** | **Failed** |

The data outcome is identical — the only thing that changes is whether anyone finds out. `Completed` wasn't degrading gracefully, it was reporting green on a run that had just left duplicate data in the ingestion container, which is the exact failure this activity's own description ("necessary to avoid data duplication in reports") exists to prevent. That's a better argument than the one I made, so withdrawing the suggestion.

It also makes `.output.exists` safe here: the Filter now only evaluates against a successful activity, and `exists` is guaranteed present in the result once it's in the `fieldList`. The expression stays redundant in the strict sense — `exists: false` implies no `childItems`, so the `contains` term alone would do — but it now matches `IngestionQueries/app.bicep:357-363` and the Analytics twin character for character, and I'd rather have three identical call sites than three subtly different ones.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/app.bicep
RolandKrummenacher: Partly retracting this.

I checked how the pipeline is invoked: `trigger_IngestionManifestAdded` fires on a blob-created event for `manifest.json` in the ingestion container (`Analytics/app.bicep:682-697`) and passes `@triggerBody().folderPath`. The folder therefore exists by construction whenever this pipeline runs, so there is no first-run "folder not created yet" case on this path the way there is in `msexports_ETL_ingestion`.

What the hunk does change is which failure surfaces when the folder genuinely is missing — removed between trigger and run, or a `folderPath` that makes the `join(skip(split(...), 1), '/')` derivation at line 1662 point somewhere wrong:

- **before:** `Get Existing Parquet Files` fails with `PathNotFound`, `Filter Out Folders` is skipped, the pipeline fails at activity level
- **now:** GetMetadata succeeds with `exists: false`, `Filter Out Folders` yields empty, `If No Files` fires the `Files Not Found` Fail activity with `errorCode: IngestionFilesNotFound`

I'll take that as an improvement rather than a regression — a defined `errorCode` is easier to alert on than a raw ADLS error, and the message ("confirm the folder path is the full path, including the `ingestion` container...") happens to describe the most likely remaining cause given that derivation. No objection to keeping it.

The one thing I'd still ask for is in the description rather than the code. It says the fix is applied "in both places this pattern exists", which reads as though both were failing the same way. They aren't: Exports writes to a *destination* folder that legitimately may not exist yet, Analytics reads a *source* folder a manifest just landed in. A sentence noting the Analytics change is diagnostic rather than a fix would stop the next person concluding that a missing ingestion folder is now a no-op there. Non-blocking.

RolandKrummenacher (approved):
Re-reviewed at 73a00761. Both of my threads are settled — the dependency-condition change is a better fix than what I suggested, and the Analytics hunk holds up once you account for how the pipeline is triggered. Details in the threads.

Net change against `dev` is +5/-3 across two files, and the three `Get Existing Parquet Files` call sites (`Exports`, `Analytics`, `IngestionQueries`) now use an identical `Succeeded` + `exists` + `contains` shape, which is worth more than the individual hunks.

One non-blocking ask left in the second thread: the description implies both call sites were failing the same way, and they weren't — Exports genuinely hits a missing destination folder on first run, Analytics doesn't. A sentence there would save the next reader the trace. Approving regardless.

MSBrett (approved):

MSBrett (approved):

src/templates/finops-hub/modules/Microsoft.CostManagement/Exports/app.bicep
flanakin, line 1: Holding for validation

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/app.bicep
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Added a sentence to the description distinguishing the two call sites — Exports is a genuine fix (missing destination folder), Analytics is diagnostic only (folder exists by construction; this just swaps `PathNotFound` for the pipeline's own `IngestionFilesNotFound` error code).

PR #2103, opened 2026-04-21
Add billing currency filter to Data Explorer dashboard (#2093)

RolandKrummenacher:
## Summary

Adds a `Currency` parameter to the FinOps hub Data Explorer dashboard so cost totals and savings KPIs can be scoped to a single billing currency. Implements Phase 1 items 1 + 2 of [#2093](https://github.com/microsoft/finops-toolkit/issues/2093); remaining items (Phase 1 item 3, savings-plan FX distortion, Phase 2 USD normalization) continue to be tracked there.

## Why

The dashboard aggregates `BilledCost` / `EffectiveCost` / `ContractedCost` / `ListCost` across rows without any `BillingCurrency` awareness. For any tenant with more than one billing currency (EA→MCA transitions, M&A, CSP partners with customers in multiple currencies, multi-cloud via FOCUS), the resulting totals and savings rates are mathematically invalid — currencies get summed as if they were comparable numbers.

The `CostsPlus` base query feeds ~74 of the 86 dashboard queries, so adding the filter there propagates the fix everywhere downstream with minimal blast radius.

### 1. New `selectedBillingCurrency` parameter (displayed as **Currency**)

- Scalar, string, sourced from `distinct BillingCurrency` over the hub's configured retention window.
- Wired into the `CostsPlus` base query as `| where isempty(selectedBillingCurrency) or BillingCurrency == selectedBillingCurrency`, immediately after the time-range filter. Inherited by all ~74 downstream tile queries through `CostsPlus` and its derived base variables (`CostsThisMonth`, `CostsLastMonth`, `CostsByMonth`, `CostsByDay`, `CostsByDayAHB`).
- Same filter also applied to the one tile that bypassed `CostsPlus` — the *Summary of list, contracted, and effective cost alignment* data-quality table on the Data ingestion page (queries `Costs()` directly).

### 2. Conditional ⚠️ All option — implicit mixed-currency warning

Rather than adding a separate warning tile (no clean place on the dense Summary page), the warning is embedded into the parameter dropdown itself:

- **Single-currency tenants**: dropdown shows only their currency. No "All" option, no noise.
- **Multi-currency tenants**: dropdown opens with **⚠️ All** as the default. The warning is visible the instant the dashboard loads, and picking any specific currency removes it automatically.

This is done via a synthetic row unioned into the parameter query only when `dcount(BillingCurrency) > 1`, with an empty `BillingCurrency` value that the filter treats as "no filter applied".

## Scope / what this PR does NOT do

[...]

Issue #2093, opened 2026-04-15
Data Explorer dashboard: multi-currency support — savings and cost KPIs silently produce wrong results


MSBrett:
## Summary

The Data Explorer dashboard (`dashboard.json`) aggregates cost columns across billing currencies without normalization or filtering. For multi-currency environments, **all cost totals and savings KPIs are mathematically invalid**.

The data layer already carries `BillingCurrency`, `PricingCurrency`, exchange rates, and USD-normalized columns — but the dashboard ignores them entirely.

## Severity

**High** for affected customers; **low visibility** because the majority of single-currency EA/MCA customers never notice.

### Who is affected
| Scenario | Risk |
|----------|------|
| Single EA enrollment (one currency) | ✅ Safe |
| Single MCA billing account | ✅ Safe |
| Multiple billing accounts (EA→MCA transition, M&A) | ⚠️ Wrong totals |
| MCA with multiple billing profiles in different currencies | ⚠️ Wrong totals |
| CSP/MPA partners with customers in multiple currencies | 🔴 **Orders of magnitude wrong** |
| Multi-cloud (AWS USD + Azure EUR via FOCUS) | ⚠️ Wrong totals |

### Worst-case example (CSP partner)
- US customer: `List=$1M`, `Effective=$500K` → 50% savings
- Japan customer: `List=¥100M`, `Effective=¥100M` → 0% savings (~$667K at ¥150/USD)
- **Dashboard ESR**: `500K / (1M + 100M)` ≈ **0.5%**
- **True USD-normalized ESR**: `500K / (1M + 667K)` ≈ **30%**

## Root cause

The `CostsPlus` base query (which feeds `CostsByMonth`, `CostsByDay`, and 74 of 86 dashboard queries) calls `Costs_v1_2` without any currency filter. All downstream queries then do:

```kql
| summarize
    ListCost = sum(ListCost),
    ContractedCost = sum(ContractedCost),
    EffectiveCost = sum(EffectiveCost)
// ...then compute savings:
| extend TotalSavings = ListCost - EffectiveCost
| extend EffectiveSavingsRate = TotalSavings / ListCost
```

These sums aggregate across all `BillingCurrency` values without grouping or filtering.

### Specific issues found

1. **~66+ of 86 queries** aggregate cost columns without currency awareness
[...]

PR #2248, opened 2026-08-12
Fix Data Explorer dashboard savings formulas and FX fallback

flanakin:
## Summary

Addresses the remaining Phase 1.3 items from [#2093](https://github.com/microsoft/finops-toolkit/issues/2093) that were left open after PR #2103 shipped the currency-filtering fix:

1. **Savings formulas** — replaced inline `sum(ListCost) - sum(EffectiveCost)` style calculations with `sum()` of the canonical, row-level clamped columns (`x_TotalSavings`, `x_CommitmentDiscountSavings`, `x_NegotiatedDiscountSavings`) across 21 tiles in `dashboard.json`.
2. ~~**Savings-plan utilization FX distortion** — removed the silent 1:1 exchange-rate fallback in the data layer.~~ **Reverted** — see note below.

### Savings formulas (dashboard.json)

`Costs`/`CostsPlus`/`CostsByMonth`/`CostsByDay` already carry `x_TotalSavings`, `x_CommitmentDiscountSavings`, and `x_NegotiatedDiscountSavings` per row, computed with built-in clamping (e.g. `x_NegotiatedDiscountSavings = iff(ListCost < ContractedCost, decimal(0), ListCost - ContractedCost)`). Several dashboard tiles instead summed the raw cost columns first and then subtracted (`sum(ListCost) - sum(EffectiveCost)`), which skips that clamping and can diverge from the canonical values (e.g. when a row's contracted cost temporarily exceeds list cost due to data timing/rounding).

Fixed by summing the canonical `x_*Savings` columns directly wherever a query aggregates over `Costs*` and then derives savings. Two tiles (`Savings breakdown by month`, `Effective cost breakdown by month`) were already locally re-deriving the same `iff()` clamp formula redundantly — simplified those to just sum the canonical columns that already exist on the source table, removing the redundant `extend`.

**Left unchanged:** the *Summary of list, contracted, and effective cost alignment* data-quality diagnostic table (Data ingestion page). That tile intentionally computes raw, unclamped deltas (`ListCost - ContractedCost`, etc.) to detect data-quality anomalies like `ListCost too low` or `ContractedCost should be 0` — using clamped values there would defeat its purpose.

**Status:** 15 of the 21 tiles were code-reviewed and verified clean as originally shipped — no further changes. The other 6 tiles required a follow-up fix (below) after review feedback.

### Follow-up: reconciling on-screen equations on 6 tiles

@rkrummenacher and @MSBrett independently flagged that 6 tiles display an equation or decomposition the user can visually verify on screen (for example, "List − Contracted = Negotiated savings", or Total savings ÷ List cost = Effective Savings Rate). On those tiles, summing the clamped per-row savings columns for the *result* side while leaving the *operand* side (`ListCost`/`ContractedCost`/`EffectiveCost`) as raw aggregates broke that on-screen reconciliation — the two sides of the displayed equation no longer matched, because clamping doesn't distribute over sums (`sum(clamp(a-b)) + sum(clamp(b-c))` isn't always `sum(clamp(a-c))`).

Rather than reverting these 6 tiles to the original raw-subtraction formulas (which would silently reintroduce the original clamping bug for just these tiles), the fix derives the displayed `ListCost`/`ContractedCost` operands from the same clamped, row-level basis as the results:

- `EffectiveCost` (raw sum) is the only unclamped anchor — it's never negative-clamped.
- `ContractedCost' = EffectiveCost + sum(x_CommitmentDiscountSavings)`
- `ListCost' = ContractedCost' + sum(x_NegotiatedDiscountSavings)`
- `TotalSavings' = sum(x_NegotiatedDiscountSavings) + sum(x_CommitmentDiscountSavings)` (the sum of the two already-clamped components, not the independently clamped `x_TotalSavings`)

This makes both sides of each displayed equation reconcile exactly, using clamped math throughout — it does not go back to the raw-subtraction bug the original PR fixed.

Affected tiles:
- Savings breakdown, monthly and daily (full List → Contracted → Effective waterfall)
- Total/Negotiated/Commitment savings breakdown bar
- Cost + savings running total stacked area chart (documented the derivation in a comment; its own stacking already reconciled internally, no formula change needed)
- Effective Savings Rate, monthly and daily

**This changes the Effective Savings Rate (ESR) definition** on the two ESR tiles, since ESR's `TotalSavings` and `ListCost` inputs are now derived the same way. A release note has been added under FinOps hubs v15 in `docs-mslearn/toolkit/changelog.md`.

### Savings-plan utilization FX distortion (data layer) — reverted

This PR originally also changed `CommitmentDiscountQuantity` derivation (`HubSetup_v1_2.kql`, `IngestionSetup_v1_2.kql`) to require `isnotempty(x_BillingExchangeRate)` before dividing by it, falling through to `real(null)` otherwise, to avoid silently assuming a 1:1 rate when the exchange rate was missing.

**That change has been reverted** after a closer look at the actual execution path found the original justification didn't hold:

- The edited branch (`CommitmentDiscountCategory == 'Spend', EffectiveCost / x_BillingExchangeRate`) sits inside a `case()` that short-circuits earlier — at `isnotempty(CommitmentDiscountQuantity)` — for rows that already have a quantity, and the EA/legacy raw-source ingestion paths cited as the motivating scenario also null out `CommitmentDiscountQuantity` upstream, so they short-circuit at that first branch and never reach the edited line at all.
- `x_BillingExchangeRate` is an "Azure 1.0-preview(v1)+" FOCUS column that Cost Management commonly leaves null for single-currency billing accounts (no exchange rate needed when `PricingCurrency == BillingCurrency`) — which is the common case, not the exception. For those accounts, assuming a rate of 1.0 is correct, not a guess.
- Requiring `isnotempty(x_BillingExchangeRate)` therefore made single-currency accounts (the majority) silently drop rows from `CommitmentDiscountQuantity`-based aggregates — `sum()` ignores nulls, so utilization percentages lose rows from their denominator without any visible signal. That trades a bounded, visible inaccuracy (a wrong-but-present number under the old code) for a silent aggregate hole (a confidently wrong percentage) under the new code — a regression for the common case, not an improvement.
- There's also a secondary effect: `CommitmentDiscountUnit` derives from `isempty(CommitmentDiscountQuantity)`, so nulling the quantity also blanks the unit column for the same rows.

Restored the original `coalesce(x_BillingExchangeRate, real(1))` fallback in both files. Only these two lines (plus the comment that justified the now-reverted change) were touched — the savings-formula fix in `dashboard.json` is untouched and remains in scope.

## Out of scope

Per the issue, this PR does **not** address:
- `BillingCurrency` visibility in detail tables (only the Purchases table currently shows it).
- Phase 2 USD-normalization toggle (blocked on `x_*InUsd` column coverage validation across ingestion paths).

## Verification

- `dashboard.json` validated as well-formed JSON before and after edits.
- `src/powershell/Tests/Unit/HubsKqlOperators.Tests.ps1` and `src/powershell/Tests/Lint/KqlJoinKinds.Tests.ps1` pass (578/578).
- Manually verified the derived-operand fix against the counterexample raised in review — two rows `(List, Contracted, Effective) = (100,120,80)` and `(100,80,60)`: clamped sums give `NegotiatedDiscountSavings=20`, `CommitmentDiscountSavings=60`; derived `EffectiveCost=140` → `ContractedCost'=200` → `ListCost'=220`; `List' - Contracted' = 20` ✓ matches Negotiated; `Contracted' - Effective = 60` ✓ matches Commitment; `TotalSavings' = 80` and `ESR = 80/220 ≈ 36.4%`, both self-consistent with the displayed operands.


src/templates/finops-hub/dashboard.json
```diff
@@ -3071,25 +3071,25 @@
...
...:\"➖\" },',\n        '{ \"order\":23, \"type\":\"Effective\", \"label\":\"After commitment discounts\", \"value\":\"', numberstring(round(EffectiveCost, 2)), '\" },',\n        '{ \"order\":24, \"type\":\"\", \"label\":\"\", \"value\":\"🟰\" },',\n        '{ \"order\":25, \"type\":\"PartialSavings\", \"label\":\"Commitment savings\", \"value\":\"', numberstring(round(CommitmentDiscountSavings, 2)), '\" },',\n        //\n        '{ \"order\":31, \"type\":\"List\", \"label\":\"Cost without discounts\", \"value\":\"', numberstring(round(ListCost, 2)), '\" },',\n        '{ \"order\":32, \"type\":\"\", \"label\":\"\", \"value\":\"➖\" },',\n        '{ \"order\":33, \"type\":\"Effective\", \"label\":\"After commitment discounts\", \"value\":\"', numberstring(round(EffectiveCost, 2)), '\" },',\n        '{ \"order\":34, \"type\":\"\", \"label\":\"\", \"value\":\"🟰\" },',\n        '{ \"order\":35, \"type\":\"TotalSavings\", \"label\":\"Total savings\", \"value\":\"', numberstring(round(TotalSavings, 2)), '\" }',\n    ']'))\n    | mv-expand json\n    | order by toint(json.order) asc\n    | project Label = tostring(json.label), Value = tostring(json.value), Type = tostring(json.type)\n);\ndata",
```
RolandKrummenacher, line 3086: **Blocking — the tile now renders arithmetic that does not add up.**

This tile (`e17346d1`) and its CostsByDay twin (`f1ef29df`) literally display three equations:

```
Cost without discounts  -  After negotiated discounts  =  Negotiated savings
After negotiated disc.  -  After commitment discounts  =  Commitment savings
Cost without discounts  -  After commitment discounts  =  Total savings
```

The change leaves the operands as raw `sum(ListCost)` / `sum(ContractedCost)` / `sum(EffectiveCost)` but swaps the results to clamped `sum(x_*Savings)`. Those disagree.

Measured on a production hub, last 3 months, applying this tile's own `x_AmortizationClass != 'Principal'` filter:

| Row | vs. the subtraction shown directly above it |
|---|---|
| Negotiated savings | **+25.9%** |
| Commitment savings | **+53.8%** |
| Total savings | **+40.2%** |

So the tile renders something like `2,283,331 - 2,056,289 = 285,782`. That is not a rounding artifact — a user will read the two numbers above the result and see it is wrong.

Either keep the raw subtraction here (this tile's whole purpose is showing the derivation), or also derive the displayed `ListCost`/`ContractedCost`/`EffectiveCost` from the clamped relationship so the equation closes.

src/templates/finops-hub/dashboard.json
```diff
@@ -3071,25 +3071,25 @@
...
```
RolandKrummenacher, line 3092: Same defect as `e17346d1` above — this is the CostsByDay twin of the savings-breakdown tile, with the same three rendered equations whose operands stayed raw while the results became clamped. Whatever fix lands on the monthly tile needs to land here too.

src/templates/finops-hub/dashboard.json
```diff
@@ -3071,25 +3071,25 @@
...
+      "text": "let data = materialize(\n    CostsByMonth | extend Period = 'Last n months'\n    | union (CostsByDay | extend Period = 'Last n days')\n    | summarize\n        TotalSavings = round(sum(x_TotalSavings), 2),\n        NegotiatedDiscountSavings = round(sum(x_NegotiatedDiscountSavings), 2),\n        CommitmentDiscountSavings = round(sum(x_CommitmentDiscountSavings), 2)\n        by\n        Period\n    | project Period, json = todynamic(strcat('[{ \"Label\":\"Total\", \"Value\":', TotalSavings, ' }, { \"Label\":\"Negotiated\", \"Value\":', NegotiatedDiscountSavings, ' }, { \"Label\":\"Commitment\", \"Value\":', CommitmentDiscountSavings, ' }]'))\n    | mv-expand json\n    | project Label = tostring(json.Label), Value = tolong(json.Value), Period\n);\ndata\n",
```
RolandKrummenacher, line 3074: **Blocking — the parts no longer sum to the whole.**

This tile shows Total / Negotiated / Commitment side by side. Because each `x_*Savings` column clamps independently at row level, `x_NegotiatedDiscountSavings + x_CommitmentDiscountSavings != x_TotalSavings`.

On a production hub across the full dataset, negotiated + commitment came out **~61% larger** than total. Previously the three were consistent by construction, since all came from the same aggregates.

A user looking at this tile will see two components that visibly exceed the total they are supposed to decompose.

src/templates/finops-hub/dashboard.json
```diff
@@ -2873,31 +2873,31 @@
...
+      "text": "let monthname = dynamic(['', 'January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December']);\nCostsPlus\n| where startofmonth(ChargePeriodStart) >= startofmonth(now(), -1)\n| summarize \n    EffectiveCost = sum(EffectiveCost),\n    CommitmentDiscountSavings = sum(x_CommitmentDiscountSavings),\n    NegotiatedDiscountSavings = sum(x_NegotiatedDiscountSavings)\n    by\n    ChargePeriodStart,\n    Month = strcat(format_datetime(ChargePeriodStart, 'MM '), monthname[monthofyear(ChargePeriodStart)])\n| order by ChargePeriodStart asc\n| extend EffectiveCostRunningTotal = row_cumsum(EffectiveCost, prev(Month) != Month)\n| extend CommitmentDiscountSavingsRunningTotal = row_cumsum(CommitmentDiscountSavings, prev(Month) != Month)\n| extend NegotiatedDiscountSavingsRunningTotal = row_cumsum(NegotiatedDiscountSavings, prev(Month) != Month)\n| project ChargePeriodStart, CommitmentDiscountSavingsRunningTotal, NegotiatedDiscountSavingsRunningTotal, EffectiveCostRunningTotal, Month\n| render areachart  ",
```
RolandKrummenacher, line 2876: Related to the parts-vs-whole problem: this area chart stacks `EffectiveCost` + `CommitmentDiscountSavings` + `NegotiatedDiscountSavings`. Previously that stack summed to exactly `ListCost` by construction, so the chart height was meaningful. With independently clamped columns it no longer does, and the stacked total will overshoot list cost by a tenant-dependent amount.

src/templates/finops-hub/dashboard.json
```diff
@@ -3251,19 +3251,19 @@
...
+      "text": "let data = materialize(\n    CostsByMonth\n    //\n    // Don't double-count commitment discount purchases\n    | where x_AmortizationClass != 'Principal'\n    //\n    | summarize \n        ListCost = sum(ListCost),\n        TotalSavings = sum(x_TotalSavings)\n    | extend EffectiveSavingsRate = TotalSavings / ListCost\n    | project json = todynamic(strcat('[',\n        '{ \"order\":11, \"type\":\"TotalSavings\", \"label\":\"Total savings\", \"value\":\"', numberstring(round(TotalSavings, 2)), '\" },',\n        '{ \"order\":12, \"type\":\"\", \"label\":\"\", \"value\":\"➗\" },',\n        '{ \"order\":13, \"type\":\"List\", \"label\":\"Cost without discounts\", \"value\":\"', numberstring(round(ListCost, 2)), '\" },',\n        '{ \"order\":14, \"type\":\"\", \"label\":\"\", \"value\":\"🟰\" },',\n        '{ \"order\":15, \"type\":\"EffectiveSavingsRate\", \"label\":\"Effective savings rate\", \"value\":\"', percentstring(EffectiveSavingsRate), '\" }',\n    ']'))\n    | mv-expand json\n    | order by toint(json.order) asc\n    | project Label = tostring(json.label), Value = tostring(json.value), Type = tostring(json.type)\n);\ndata",
```
RolandKrummenacher, line 3254: **Effective savings rate is silently inflated here.**

`EffectiveSavingsRate = sum(x_TotalSavings) / sum(ListCost)` now mixes a clamped numerator with a raw denominator, so the ratio is no longer internally consistent.

On a production hub, last 3 months, the reported ESR rose by roughly **40% relative** to the pre-change value. Broken out by month the instability is worse — one month moved **+230%** because a single large negative-savings correction got clamped to zero instead of netting out, while adjacent months moved 11-17%.

ESR is a headline FinOps KPI. A step change of that size with no release note will read as a regression to customers, and the month-to-month series becomes non-comparable across the upgrade boundary.

src/templates/finops-hub/dashboard.json
```diff
@@ -3251,19 +3251,19 @@
...
+      "text": "let data = materialize(\n    CostsByDay\n    //\n    // Don't double-count commitment discount purchases\n    | where x_AmortizationClass != 'Principal'\n    //\n    | summarize \n        ListCost = sum(ListCost),\n        TotalSavings = sum(x_TotalSavings)\n    | extend EffectiveSavingsRate = TotalSavings / ListCost\n    | project json = todynamic(strcat('[',\n        '{ \"order\":11, \"type\":\"TotalSavings\", \"label\":\"Total savings\", \"value\":\"', numberstring(round(TotalSavings, 2)), '\" },',\n        '{ \"order\":12, \"type\":\"\", \"label\":\"\", \"value\":\"➗\" },',\n        '{ \"order\":13, \"type\":\"List\", \"label\":\"Cost without discounts\", \"value\":\"', numberstring(round(ListCost, 2)), '\" },',\n        '{ \"order\":14, \"type\":\"\", \"label\":\"\", \"value\":\"🟰\" },',\n        '{ \"order\":15, \"type\":\"EffectiveSavingsRate\", \"label\":\"Effective savings rate\", \"value\":\"', percentstring(EffectiveSavingsRate), '\" }',\n    ']'))\n    | mv-expand json\n    | order by toint(json.order) asc\n    | project Label = tostring(json.label), Value = tostring(json.value), Type = tostring(json.type)\n);\ndata",
```
RolandKrummenacher, line 3266: Same ESR issue as `e1bc2d51` — clamped numerator over raw denominator, on the CostsByDay variant.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts/HubSetup_v1_2.kql
```diff
@@ -114,7 +114,10 @@ Costs_v1_2()
         )
         | extend CommitmentDiscountQuantity = case(
             isempty(CommitmentDiscountStatus), real(null),
-            CommitmentDiscountCategory == 'Spend', EffectiveCost / coalesce(x_BillingExchangeRate, real(1)),
+            // Don't assume a 1:1 exchange rate when it's missing -- that silently distorts
+            // the derived quantity (and any utilization built on it) by the true FX factor.
+            // Leave it unset like the other not-derivable cases below instead.
+            CommitmentDiscountCategory == 'Spend' and isnotempty(x_BillingExchangeRate), EffectiveCost / x_BillingExchangeRate,
```
RolandKrummenacher: **This likely regresses non-Microsoft FOCUS ingestion.**

`x_BillingExchangeRate` is a Microsoft-specific extension column — it does not exist in the FOCUS 1.0 spec. For an AWS or GCP FOCUS export, `CommitmentDiscountCategory == 'Spend'` is exactly what AWS Savings Plans land on, and those rows will have no exchange rate while `BillingCurrency == PricingCurrency`. In that case the old 1:1 assumption was **correct**, and the new guard nulls out `CommitmentDiscountQuantity` for every one of them. `CommitmentDiscountUnit` then follows it to `''` via the `isempty(CommitmentDiscountQuantity)` branch just below.

Suggested narrower guard that keeps the real fix while not punishing same-currency sources:

```kql
CommitmentDiscountCategory == 'Spend' and isnotempty(x_BillingExchangeRate), EffectiveCost / x_BillingExchangeRate,
CommitmentDiscountCategory == 'Spend' and (isempty(PricingCurrency) or PricingCurrency == BillingCurrency), EffectiveCost,
```

`PricingCurrency` is explicitly set to `''` when it cannot be determined (`IngestionSetup_v1_2.kql:1180` and `:1401`), so the guard is safe.

Worth noting: I could not find a single row the strict version actually fixes. Across three production Azure hubs, every spend-commitment row had the exchange rate populated (and equal to 1) — zero missing. So as written this half of the PR changes nothing observable on Azure while creating a real risk on the non-Azure path.

Separately, the description justifies not adding a flag by pointing at `x_SourceValues`'s `checkReal('CommitmentDiscountQuantity', ...)`. That exists only at `IngestionSetup_v1_2.kql:886`. This function (`Costs_v1_2()`) applies the same `case()` to unioned `Costs_final_v1_0` rows with no `old_` capture and no `checkReal` — it just passes `x_SourceValues` through. So legacy v1.0 rows lose the quantity with no audit trail.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts/IngestionSetup_v1_2.kql
```diff
@@ -529,7 +529,10 @@ Costs_transform_v1_2()
         isnotempty(CommitmentDiscountQuantity), CommitmentDiscountQuantity,
         // FOCUS 1.0-preview, 1.0
         isempty(CommitmentDiscountStatus), real(null),
-        CommitmentDiscountCategory == 'Spend', EffectiveCost / coalesce(x_BillingExchangeRate, real(1)),
+        // Don't assume a 1:1 exchange rate when it's missing -- that silently distorts
+        // the derived quantity (and any utilization built on it) by the true FX factor.
+        // Leave it unset like the other not-derivable cases below instead.
+        CommitmentDiscountCategory == 'Spend' and isnotempty(x_BillingExchangeRate), EffectiveCost / x_BillingExchangeRate,
```
RolandKrummenacher: Same concern as the `HubSetup_v1_2.kql` copy — see that comment for the detail and the suggested currency-equality guard. This is the path that does have the `old_CommitmentDiscountQuantity` / `checkReal` audit trail, so the change is at least observable here.

RolandKrummenacher (changes requested):
Reviewed the refactor and validated the assumptions by running the before/after formulas against three production hubs I have read access to (read-only queries, no identifying details below). The mechanical KQL refactor is clean, but the behavior change is materially larger than the description suggests, and it breaks the tiles that render savings as a visible equation.

## Summary

The core issue is that `x_TotalSavings`, `x_CommitmentDiscountSavings`, and `x_NegotiatedDiscountSavings` are clamped **per row**. Summing them is therefore not equivalent to subtracting the aggregates — it is a different metric, one that structurally cannot net out credits, refunds, and corrections. On one hub the last-3-months total savings comes out **~40% higher** after this change; on another the divergence is ~0.04%. So the blast radius is entirely tenant-dependent, which makes shipping it silently riskier, not safer.

That may well be the metric we want. But it is a product decision that belongs in #2093 with a release note, not a side effect of a "use the canonical column" refactor.

Details in the inline comments. Ordered by severity:

1. **Savings-breakdown tiles now display arithmetic that does not add up** — operands stayed raw, results became clamped.
2. **Parts no longer sum to the whole** — negotiated + commitment overstates total by ~61% on one hub.
3. **Effective savings rate silently inflated** — clamped numerator over raw denominator.
4. **The FX guard likely regresses non-Microsoft FOCUS ingestion** — and is a no-op on every Azure hub I checked.
5. The auditability claim in the description only holds for one of the two code paths.

## What is right

- The refactor itself is correct KQL. The union/pivot tiles properly re-sum the already-renamed columns in the second `summarize`.
- Dropping the redundant local `extend` re-derivations in `3b3f0a58` and `0341b3e4` is a genuine simplification — the latter's were entirely dead code.
- Removing now-unused `ListCost`/`ContractedCost` from `summarize` lists (e.g. `2d7b6447`) is a real efficiency win on wide scans.
- Leaving the *Summary of list, contracted, and effective cost alignment* diagnostic on raw unclamped deltas is the right call, and the reasoning in the description is sound.

## Two things not introduced here, but now load-bearing

**The two clamp formulas disagree.** `HubSetup_v1_2.kql:169-171` uses `iff(ContractedCost < EffectiveCost, real(0), ...)`; `IngestionSetup_v1_2.kql:706-708` uses `iff(isempty(ListCost) or ListCost == 0 or ListCost - EffectiveCost < 0.0001, real(0), ...)`. Pre-existing, but this PR makes 21 tiles depend on those columns, so a hub with mixed v1.0/v1.2 data will now get two different clamping semantics inside a single displayed number. Worth aligning while we are here.

**Dashboard/hub version coupling.** I verified via `getschema` that a hub running a current released version does not yet expose the `x_*Savings` columns on `Costs_v1_2()` — they come from this branch. Fine for a normal template deploy, but importing the new `dashboard.json` against a not-yet-upgraded hub now produces hard query failures on all 21 tiles rather than degraded output. Probably worth a release-note line.

## Recommendation

Blocking on 1-3: either keep raw subtraction in the tiles that render savings as a visible equation, or derive the displayed operands from the clamped relationship so the math closes. 4 should be narrowed to a currency-equality guard before merge. 5 is a follow-up.

Happy to take the tile fixes if that is useful.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts/HubSetup_v1_2.kql
flanakin: I'm good with this addition, but it leaves us open for any cases where PricingCurrency isn't being set and we don't have an exchange rate. Is it better to leave CommitmentDiscountQuantity empty in those cases or assume a 1:1 exchange rate and populate it with the EffectiveCost? 🤔 If it's empty and FinOps practitioners need it, they'll just use EffectiveCost anyway, so I'm thinking it's best to fall back to EffectiveCost. And if we do that, this is a complete noop. It's arguably better logic, but no actual change.

RolandKrummenacher: > And if we do that, this is a complete noop.

That's true only if the `EffectiveCost` fallback also applies when the currencies are *known to differ* — in that case the whole expression collapses back to `coalesce(x_BillingExchangeRate, real(1))` and we've written the original logic in more words. But that case is the one the PR was actually fixing, so it doesn't have to be included in the fallback.

Splitting it three ways keeps your fallback and is not a no-op:

```kql
CommitmentDiscountCategory == 'Spend' and isnotempty(x_BillingExchangeRate), EffectiveCost / x_BillingExchangeRate,
CommitmentDiscountCategory == 'Spend' and (isempty(PricingCurrency) or PricingCurrency == BillingCurrency), EffectiveCost,
// Spend, no exchange rate, and a known-DIFFERENT pricing currency: we cannot convert,
// so leave it empty rather than publish a number we know to be wrong.
```

Which gives:

| Case | Rate | Currencies | Result |
|---|---|---|---|
| 1 | present | any | `EffectiveCost / x_BillingExchangeRate` |
| 2 | absent | equal (AWS/GCP Savings Plans) | `EffectiveCost` — 1:1 is genuinely correct here |
| 3 | absent | **different** | empty — the actual fix |
| 4 | absent | `PricingCurrency` unset | `EffectiveCost` — your case |

Case 4 is where I'd land on your side: `PricingCurrency` is required in FOCUS 1.0, so an empty one means a malformed export, and falling back there matches the previous behaviour and costs nothing. Case 3 is the only behaviour change, and it's the one worth having — a fabricated 1:1 quantity is worse than an empty one, because an empty column reads as "unavailable" while a wrong number reads as fact.

> If it's empty and FinOps practitioners need it, they'll just use EffectiveCost anyway

Agreed for case 4. In case 3 that's exactly what we don't want them to do — `EffectiveCost` is in billing currency there, and the commitment is denominated in pricing currency, so silently substituting one for the other is the bug rather than the workaround.

Happy to measure how many rows actually land in each bucket on a real hub if that would help settle it — my guess is case 3 is rare on Microsoft data and case 2 is the whole of any AWS/GCP export.

RolandKrummenacher: Fine with the revert, and it's the right call for this PR even though it drops the one case I argued for.

Your reasoning holds for Microsoft data: a null `x_BillingExchangeRate` there means single-currency billing, so the currencies-differ case doesn't arise, and the EA/legacy rows null `CommitmentDiscountQuantity` upstream and short-circuit at the first `case()` branch before reaching this one. I'd only qualify it against the four cases from the thread above:

| Case | Rate | Currencies | After this commit |
|---|---|---|---|
| 1 | present | any | `EffectiveCost / x_BillingExchangeRate` |
| 2 | absent | equal | `EffectiveCost` — correct |
| 3 | absent | **different** | `EffectiveCost` — still fabricated |
| 4 | absent | unset | `EffectiveCost` — correct |

Case 3 is only reachable on a third-party FOCUS 1.0 export: `x_BillingExchangeRate` is a Microsoft extension column, so it's absent from every AWS/GCP row, and a Savings Plan billed in a currency other than its pricing currency lands there. That's narrow, unmeasured, and not something a dashboard fix should be carrying.

Net effect is that this half of the PR is now a no-op against `dev`, which matches what we both measured — so the PR is a pure `dashboard.json` change and easier to reason about. Good outcome.

If you want case 3 tracked, I'm happy to open a separate issue against the FOCUS 1.0 derivation rather than reopen it here — say the word and I'll write it up with the currency-split logic.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Analytics/scripts/IngestionSetup_v1_2.kql
RolandKrummenacher: Same as the [HubSetup thread](https://github.com/microsoft/finops-toolkit/pull/2248#discussion_r3773072245) — reverted here too, no objection. This was the copy with the `old_CommitmentDiscountQuantity` / `checkReal` audit trail, so if the currency-split ever comes back it should land here first where it's observable.

RolandKrummenacher (changes requested):
Re-reviewed at 0e05f98b. The KQL revert resolves the two threads it touches — replied there, no objection. The PR is now a pure `dashboard.json` change, +21/-21.

The six dashboard threads are untouched by this commit and all still apply, so keeping changes requested. Consolidating them here, since they're one root cause rather than six independent problems.

**Root cause:** the change swaps raw subtractions for `sum(x_*Savings)` on the *result* side of each tile, while leaving `sum(ListCost)` / `sum(ContractedCost)` / `sum(EffectiveCost)` raw on the *operand* side. The `x_*Savings` columns clamp negatives at row level, so the two sides no longer reconcile.

| Tile | Symptom, measured on a production hub |
|---|---|
| Savings breakdown, monthly (`e17346d1`) + daily (`f1ef29df`) | the three equations the tile literally displays don't hold: **+25.9%**, **+53.8%**, **+40.2%** |
| Total / Negotiated / Commitment (line 3074) | components exceed the total they decompose by **~61%** |
| Stacked savings area chart (line 2876) | stack no longer sums to `ListCost`, overshoots by a tenant-dependent amount |
| Effective Savings Rate (`e1bc2d51` + daily twin) | clamped numerator over raw denominator: ESR **+40%** relative, one month **+230%** |

One decision settles all four: either keep the raw subtractions on these tiles, or derive the displayed cost operands from the clamped columns so the arithmetic closes. I don't have a strong preference as long as each tile is internally consistent — a user reading two numbers and their stated difference shouldn't see a third number that isn't the difference.

ESR is the one I'd treat as separate from the arithmetic question: it's a headline KPI, and a step change of that size across an upgrade boundary makes the month-over-month series non-comparable. If the clamped definition is the one we want, that needs a release note rather than just a formula change.

Happy to re-measure whichever direction you pick.

MSBrett (changes requested):
Requesting changes. I re-reviewed the current head (`0e05f98b`) from the base diff, traced the savings fields to their source definitions, checked all 21 changed queries, and independently reproduced the arithmetic failures below.

The replacement of aggregate subtraction with sums of row-level clamped savings is not semantics-preserving for visuals that still rely on exact accounting relationships:

1. **The monthly and daily savings-summary tiles display false equations** (`dashboard.json:3086` and `:3092`). They still show raw aggregate values for List, Contracted, and Effective cost, followed by subtraction and equals signs, but the displayed results now sum independently clamped row values. Whenever a row is clamped, the displayed operands no longer produce the displayed result.

2. **The Total / Negotiated / Commitment summary is no longer a valid decomposition** (`dashboard.json:3074`). Independently clamped negotiated and commitment savings are not additive and can sum to more than total savings.

3. **The stacked Cost + savings chart no longer reconciles to list cost** (`dashboard.json:2876`). It stacks raw effective cost with both independently clamped savings categories, so the stack can overstate the implied cost without discounts.

A minimal counterexample confirms all three failures: for rows `(List, Contracted, Effective) = (100,120,80)` and `(100,80,60)`, the raw negotiated difference is `0` while summed clamped negotiated savings is `20`; negotiated plus commitment savings is `80` while total savings is `60`; and the stacked value is `220` against list cost of `200`.

The clamped Effective Savings Rate formula is not inherently invalid—it matches the repository's `savings-summary-report.kql` precedent—but it is a material KPI-definition change rather than a mechanical refactor. If retained for ESR and standalone savings visuals, the changed semantics and historical discontinuity should be explicitly approved and documented in release notes.

Please either keep aggregate subtraction for visuals that present reconciled equations/components, or redesign those visuals and labels so the independently clamped metrics are not presented as arithmetic identities. Also update the PR title/body now that the FX change has been reverted and remove the claim that the savings portion was verified clean while these blocking threads remain unresolved.

src/templates/finops-hub/dashboard.json
flanakin: Fixed by deriving the displayed operands from the same clamped, row-level basis as the results, rather than reverting to raw subtraction (which would have silently reintroduced the original clamping bug this PR set out to fix for this tile).

`EffectiveCost` (raw sum) is the only value that's never negative-clamped, so it anchors the chain:
- `ContractedCost' = EffectiveCost + sum(x_CommitmentDiscountSavings)`
- `ListCost' = ContractedCost' + sum(x_NegotiatedDiscountSavings)`
- `TotalSavings' = sum(x_NegotiatedDiscountSavings) + sum(x_CommitmentDiscountSavings)` (sum of the two clamped components, not the independently-clamped `x_TotalSavings`)

All three displayed equations now hold exactly: `List' - Contracted' = Negotiated`, `Contracted' - Effective = Commitment`, `List' - Effective = TotalSavings'`. Verified against your counterexample rows `(100,120,80)` and `(100,80,60)`: `Negotiated=20`, `Commitment=60`, `Effective=140` → `Contracted'=200` → `List'=220`; `220-200=20` ✓, `200-140=60` ✓, `220-140=80=TotalSavings'` ✓.

The other tiles' clamped-sum fix is unaffected by this change.

src/templates/finops-hub/dashboard.json
flanakin: Fixed by deriving the displayed operands from the same clamped, row-level basis as the results, rather than reverting to raw subtraction (which would have silently reintroduced the original clamping bug this PR set out to fix for this tile).

`EffectiveCost` (raw sum) is the only value that's never negative-clamped, so it anchors the chain:
- `ContractedCost' = EffectiveCost + sum(x_CommitmentDiscountSavings)`
- `ListCost' = ContractedCost' + sum(x_NegotiatedDiscountSavings)`
- `TotalSavings' = sum(x_NegotiatedDiscountSavings) + sum(x_CommitmentDiscountSavings)` (sum of the two clamped components, not the independently-clamped `x_TotalSavings`)

All three displayed equations now hold exactly: `List' - Contracted' = Negotiated`, `Contracted' - Effective = Commitment`, `List' - Effective = TotalSavings'`. Verified against your counterexample rows `(100,120,80)` and `(100,80,60)`: `Negotiated=20`, `Commitment=60`, `Effective=140` → `Contracted'=200` → `List'=220`; `220-200=20` ✓, `200-140=60` ✓, `220-140=80=TotalSavings'` ✓.

The other tiles' clamped-sum fix is unaffected by this change.

src/templates/finops-hub/dashboard.json
flanakin: Fixed. This tile doesn't display List/Contracted/Effective operands, only the three savings bars (Total, Negotiated, Commitment), so the fix here is narrower than the waterfall tiles: `TotalSavings` is now derived as `sum(x_NegotiatedDiscountSavings) + sum(x_CommitmentDiscountSavings)` instead of the independently-clamped `sum(x_TotalSavings)`, so the Total bar always equals Negotiated + Commitment shown alongside it — the parts sum to the whole again, by construction.

Verified against your counterexample rows: `Negotiated=20`, `Commitment=60` → `Total'=80`, which is internally consistent (no longer the independently-clamped `60` that didn't match `20+60=80`).

The other tiles' clamped-sum fix is unaffected by this change.

src/templates/finops-hub/dashboard.json
flanakin: Looked at this one closely: the stack itself (`EffectiveCost` base layer + `sum(x_CommitmentDiscountSavings)` + `sum(x_NegotiatedDiscountSavings)`, both already clamped sums per this PR) reconciles internally by construction — a stacked area chart's top is always the sum of its layers, so the stack top already equals `EffectiveCost + Commitment + Negotiated`, the same derived "cost without discounts" quantity used in the waterfall/ESR tiles after their fix. No formula change was needed here.

What I did change: added a comment in the query clarifying that the stack top is this *derived* figure, not the raw `sum(ListCost)` KPI shown elsewhere on the page — so nobody cross-checking the stack height against a separate raw List cost tile is confused by a legitimate difference between the two (your `220` vs `200` example). That raw/derived distinction is now called out explicitly at the top of the query.

The other tiles' clamped-sum fix is unaffected by this change.

src/templates/finops-hub/dashboard.json
flanakin: Fixed by deriving `TotalSavings` and `ListCost` from the same clamped, row-level basis as the other tiles rather than mixing a clamped numerator with a raw denominator:

- `EffectiveCost` = raw sum (never negative-clamped)
- `TotalSavings' = sum(x_NegotiatedDiscountSavings) + sum(x_CommitmentDiscountSavings)`
- `ListCost' = EffectiveCost + TotalSavings'`
- `EffectiveSavingsRate = TotalSavings' / ListCost'`

This is a genuine ESR definition change, as you and MSBrett both flagged — the numerator and denominator are now internally consistent with each other and with the other savings tiles on the dashboard, but the resulting rate will differ from the pre-PR raw-subtraction ESR whenever any row was clamped, so it isn't just a formula cleanup. Added a release note under FinOps hubs v15 in `docs-mslearn/toolkit/changelog.md` calling out the definition change explicitly, per your ask.

Verified against the counterexample: `Effective=140`, `TotalSavings'=80` → `ListCost'=220` → `ESR = 80/220 ≈ 36.4%`, self-consistent with the displayed operands.

src/templates/finops-hub/dashboard.json
flanakin: Fixed by deriving `TotalSavings` and `ListCost` from the same clamped, row-level basis as the other tiles rather than mixing a clamped numerator with a raw denominator:

- `EffectiveCost` = raw sum (never negative-clamped)
- `TotalSavings' = sum(x_NegotiatedDiscountSavings) + sum(x_CommitmentDiscountSavings)`
- `ListCost' = EffectiveCost + TotalSavings'`
- `EffectiveSavingsRate = TotalSavings' / ListCost'`

This is a genuine ESR definition change, as you and MSBrett both flagged — the numerator and denominator are now internally consistent with each other and with the other savings tiles on the dashboard, but the resulting rate will differ from the pre-PR raw-subtraction ESR whenever any row was clamped, so it isn't just a formula cleanup. Added a release note under FinOps hubs v15 in `docs-mslearn/toolkit/changelog.md` calling out the definition change explicitly, per your ask.

Verified against the counterexample: `Effective=140`, `TotalSavings'=80` → `ListCost'=220` → `ESR = 80/220 ≈ 36.4%`, self-consistent with the displayed operands.

src/templates/finops-hub/dashboard.json
```diff
@@ -3071,25 +3071,25 @@
...
+      "text": "let data = materialize(\n    CostsByMonth | extend Period = 'Last n months'\n    | union (CostsByDay | extend Period = 'Last n days')\n    | summarize\n        NegotiatedDiscountSavings = round(sum(x_NegotiatedDiscountSavings), 2),\n        CommitmentDiscountSavings = round(sum(x_CommitmentDiscountSavings), 2)\n        by\n        Period\n    //\n    // Total savings is derived as the sum of the two clamped components (not the\n    // independently row-clamped x_TotalSavings) so the Total bar always reconciles\n    // with Negotiated + Commitment shown alongside it.\n    | extend TotalSavings = NegotiatedDiscountSavings + CommitmentDiscountSavings\n    | project Period, json = todynamic(strcat('[{ \"Label\":\"Total\", \"Value\":', TotalSavings, ' }, { \"Label\":\"Negotiated\", \"Value\":', NegotiatedDiscountSavings, ' }, { \"Label\":\"Commitment\", \"Value\":', CommitmentDiscountSavings, ' }]'))\n    | mv-expand json\n    | project Label = tostring(json.Label), Value = tolong(json.Value), Period\n);\ndata\n",
```
RolandKrummenacher, line 3074: **Blocking — this tile and `Cost summary` sit side by side and no longer reconcile.**

Layout puts `Cost summary` at `x=6, y=18, w=8, h=5` and this tile at `x=14, y=18, w=8, h=5` — same row, same height, directly adjacent. Both read the identical `CostsByMonth | union CostsByDay` source with the identical `by Period` grouping.

`Cost summary` (query `5ff29428`) is **not** touched by this PR and still shows raw aggregates. On the counterexample rows it displays List 200 / Contracted 200 / Effective 140, from which a reader derives Negotiated 0, Commitment 60, Total 60. This tile now shows Negotiated 20, Commitment 60, Total 80.

Before this PR the two matched by construction — the base version of this query computed `ListCost - ContractedCost` and `ContractedCost - EffectiveCost` from those same raw sums. Moving to clamped columns here while leaving `Cost summary` raw is the same reconciliation break you just fixed on the six other tiles, except the operands live in a *different tile*, so it can't be fixed inside this query alone.

Either `Cost summary` moves to the derived basis (`Contracted' = Effective + comm`, `List' = Contracted' + neg`), or both tiles stay raw. The first is consistent with the direction of the rest of the PR, and it also fixes the stacked-area overshoot from the still-open thread above, since that chart is on the same page and compares against the same list cost.

src/templates/finops-hub/dashboard.json
```diff
@@ -3113,25 +3113,25 @@
...
       "id": "5bbb5369-ac95-45fc-853c-a6a2ce6a9e7b",
       "usedVariables": ["CostsByMonth"]
     },
     {
       "dataSource": { "kind": "inline", "dataSourceId": "23540be2-ffc9-4b61-8c4c-05e493e682a6" },
-      "text": "CostsByMonth\n//\n// Filter out commitment discount purchases\n| where ChargeCategory == 'Usage' or isempty(CommitmentDiscountId)\n//\n| summarize \n    Savings = round(sum(ListCost - EffectiveCost), 2)\n    by\n    Type = case(\n        isnotempty(CommitmentDiscountType), CommitmentDiscountType,\n        'Negotiated'\n    )",
+      "text": "CostsByMonth\n//\n// Filter out commitment discount purchases\n| where ChargeCategory == 'Usage' or isempty(CommitmentDiscountId)\n//\n| summarize \n    Savings = round(sum(x_TotalSavings), 2)\n    by\n    Type = case(\n        isnotempty(CommitmentDiscountType), CommitmentDiscountType,\n        'Negotiated'\n    )",
```
RolandKrummenacher, line 3122: **Blocking — "Total savings" now means two different numbers on the same page, in adjacent tiles.**

This pie is at `x=15, y=17, w=7, h=10`; **Effective Savings Rate (last n months)** (query `e1bc2d51`) is at `x=0, y=17, w=15, h=4` — same row, immediately to its left. The ESR tile renders a labelled operand `Total savings` = `sum(neg) + sum(comm)` = **80**. This tile is titled `Total savings (last n months)` and sums `x_TotalSavings` = **60**.

Worth noting this is a regression the follow-up commit introduced rather than a pre-existing one: the *first* commit moved this pie from `sum(ListCost - EffectiveCost)` to `sum(x_TotalSavings)`, which aligned it with the ESR tile's then-`sum(x_TotalSavings)`. The follow-up then moved the ESR tile to `sum(neg) + sum(comm)` and left the pie behind.

The same applies to the `(last n days)` pair at `y=53` (`9a9ee4b0` beside `cb34f22b`), and to `Monthly savings trend` / `Daily savings trend`, which also sum `x_TotalSavings`.

Pick one definition of total savings and use it everywhere it is displayed under that label. Given the parts-must-sum-to-the-whole argument that drove the `Savings summary` fix, `sum(neg) + sum(comm)` is the one to standardize on — and it has the side benefit that this pie's `by Type` split then actually adds up to the total shown next to it.

src/templates/finops-hub/dashboard.json
```diff
@@ -3251,19 +3251,19 @@
...
+      "text": "CostsByMonth\n| extend x_AmortizationClass = case(\n    ChargeCategory == 'Purchase' and isnotempty(CommitmentDiscountCategory), 'Principal',\n    isnotempty(CommitmentDiscountCategory), 'Amortized Charge',\n    ''\n)\n| summarize\n    ['List cost']      = round(sumif(ListCost, x_AmortizationClass != 'Principal'), 2),\n    ['Effective cost'] = round(sum(EffectiveCost), 2),\n    Savings            = round(sum(x_TotalSavings), 2)\n    by\n    Account = x_BillingProfileId,\n    Month   = substring(startofmonth(ChargePeriodStart), 0, 7)\n| extend ESR = percentstring(Savings/ ['List cost'])\n| order by Month desc",
```
RolandKrummenacher, line 3260: **Blocking — the ESR redefinition was applied to two tiles and not to this one.**

This grid is at `x=0, y=27` on the Rate optimization page, ten rows below **Effective Savings Rate (last n months)** at `y=17`. It still computes:

```kql
['List cost']      = round(sumif(ListCost, x_AmortizationClass != 'Principal'), 2),
['Effective cost'] = round(sum(EffectiveCost), 2),
Savings            = round(sum(x_TotalSavings), 2)
| extend ESR = percentstring(Savings / ['List cost'])
```

That is precisely the clamped-numerator-over-raw-denominator mix I flagged in the ESR thread — it just wasn't caught because this tile computes its ESR inline instead of through the multistat template. On the counterexample rows this row reads List 200, Effective 140, Savings 60, **ESR 30.0%**, while the ESR tile ten rows up now reads **36.4%**. In the base they both read 30.0%.

It also displays `List cost` and `Effective cost` in the same grid row as `Savings`, so it has the on-screen-equation property too: `List − Effective` should equal `Savings`.

Same treatment as the other ESR tiles: derive `Savings = sum(neg) + sum(comm)` and `List cost = Effective cost + Savings`.

Separately, and pre-existing so not blocking: `Savings` sums over all rows while `List cost` uses `sumif(..., x_AmortizationClass != 'Principal')`, so purchase rows are excluded from one side of the ratio but not the other. Worth aligning while you're in here.

RolandKrummenacher (changes requested):
Re-reviewed after the follow-up commit. **The derived-operand fix is correct** — I verified all six tiles from the previous round and resolved five of those threads.

The derivation `Contracted' = Effective + sum(x_CommitmentDiscountSavings)`, `List' = Contracted' + sum(x_NegotiatedDiscountSavings)`, `Total' = sum(neg) + sum(comm)` makes each of those tiles reconcile exactly on screen, and it does it without going back to the raw-subtraction bug the PR set out to fix. Anchoring on `EffectiveCost` as the only never-clamped value is the right choice. I also agree with the call to document the ESR redefinition in the changelog rather than bury it.

**What blocks:** the new invariant — *displayed savings and their displayed operands both come from the clamped basis* — was applied to six tiles but not to the remaining tiles that render the same quantities. Three places now disagree with a tile sitting directly beside them. Two of these date from the first commit and I missed them in my first pass; they are not new breakage from the follow-up.

I verified each against the head of this branch, using the canonical definitions in `HubSetup_v1_2.kql:169-171` and the counterexample from the earlier thread — rows `(List, Contracted, Effective)` = `(100,120,80)` and `(100,80,60)`:

| | value |
|---|---|
| `sum(ListCost)` / `sum(ContractedCost)` / `sum(EffectiveCost)` | 200 / 200 / 140 |
| `sum(x_NegotiatedDiscountSavings)` | 20 |
| `sum(x_CommitmentDiscountSavings)` | 60 |
| `sum(x_TotalSavings)` | 60 |

Details in the three inline comments. All three are consistency defects — no syntax or column-resolution problems, and none of the six reworked tiles is internally inconsistent.

One thread left open: the **Cost + savings running total** stacked area chart. Your reply and the added comment are fair, but the stack top is `Effective + comm + neg` = 220 in the example, while **Cost summary** seven rows below on the same page still shows list cost = 200. That resolves itself once the Summary page pair below is settled, which is why I left it linked rather than resolved.

Everything else I checked is clean: all `x_*Savings` references resolve against `Costs_final_v1_2`; no `visualOptions` (color rules, pivot columns, multistat label/value columns) references a dropped column; the `| as per | union (per | summarize ...)` re-aggregations in the subscription and chargeback tiles correctly re-sum already-aggregated savings columns; the two removed local `extend x_*Savings` blocks were dead or shadowing; both JSON files parse; and the FX-fallback revert is complete — neither KQL file appears in the diff.

PR #2249, opened 2026-08-12
Add opt-in Key Vault RBAC authorization for CAF/landing zone compliance

### LINKED ISSUE #1067: Update Key Vault to Support RBAC Permissions and Delete Protection
## ⚠️ Problem
For organizations that utilize the [Enterprise Scale](https://github.com/Azure/terraform-azurerm-caf-enterprise-scale) in association with the [Cloud Adoption Framework (CAF)](https://github.com/aztfmod/terraform-azurerm-caf), there are various Azure Policies and Initiatives that are deployed. 

One of those Initiatives is the **Enforce recommended guardrails for Azure Key Vault**. This initiative contains several policies, including:

- Azure Key Vault should use RBAC permission model
- Key vaults should have deletion protection enabled

The current deployment of the FinOps Toolkit (Hub architecture), violates those 2 policies and prevents its deployment. We have to add/create an exception in the target Subscription / Resource Group, for the deployment to complete successfully.

## 🛠️ Solution

Update the FinOps Toolkit's Key Vault implementation to support the [RBAC permissions](https://learn.microsoft.com/en-us/azure/key-vault/general/rbac-guide?tabs=azure-cli) model, and also enable [delete protection](https://learn.microsoft.com/en-us/azure/key-vault/general/soft-delete-overview). 

## ℹ️ Additional context
The client I am currently assisting, uses the CAF / Enterprise Scale Terraform modules.

## 🙋‍♀️ Ask for the community
We could use your help:

flanakin:
## Summary

Adds an `enableRbacAuthorization` parameter to the FinOps hub Bicep templates that switches the remote-hub Key Vault from the legacy access-policy model to Azure RBAC. This is required by Cloud Adoption Framework (CAF) / Enterprise-Scale landing zone Azure Policy guardrails (e.g., "Enforce recommended guardrails for Azure Key Vault"), which flag vaults that use access policies instead of RBAC.

Fixes #1067

## What changed

- **`src/templates/finops-hub/modules/fx/hub-app.bicep`** — the Key Vault resource now sets `enableRbacAuthorization: app.hub.options.keyVaultEnableRbacAuthorization` instead of a hardcoded `false`. When RBAC is enabled:
  - `accessPolicies` is forced to an empty array (Azure rejects a non-empty `accessPolicies` array when `enableRbacAuthorization: true`).
  - A new `keyVaultRoleAssignment` resource grants the Data Factory managed identity the built-in **Key Vault Secrets User** role (`4633458b-17de-408a-b874-0445c86b69e6`) on the vault — the RBAC equivalent of the `secrets: ['get']` access policy it previously relied on. This is the *only* identity that reads secrets from this vault (verified below), so it's the only equivalent role assignment needed.
- **`src/templates/finops-hub/modules/fx/hub-types.bicep`** — added `keyVaultEnableRbacAuthorization` to `HubProperties.options` and threaded it through `newHubInternal`/`newHub`.
- **`src/templates/finops-hub/modules/hub.bicep`** and **`main.bicep`** — added an `enableRbacAuthorization bool = false` parameter, following the exact pattern already used for `enablePurgeProtection`.
- **`createUiDefinition.json`** — added a matching "Enable Key Vault RBAC authorization" checkbox next to the existing purge protection checkbox, gated to the remote-hub analytics engine path (the only path that deploys this vault).
- **`docs-mslearn/toolkit/changelog.md`** — added an entry under Unreleased.

Purge protection (`enablePurgeProtection` / `enableSoftDelete`) was **already fully implemented** on this branch (not something I needed to add) — `enableSoftDelete: true` is hardcoded, and `enablePurgeProtection` is already wired end-to-end as an opt-in parameter defaulting to `false`, following the exact convention I followed for RBAC here.

## Investigation: what currently depends on access policies

I traced every consumer of the Key Vault (`Microsoft.KeyVault/vaults` in `src/templates/finops-hub/modules/fx/hub-app.bicep:471`, only deployed when the `RemoteHub` app requests the `'KeyVault'` feature — `modules/Microsoft.FinOpsHubs/RemoteHub/app.bicep:52`):

1. **Data Factory's `AzureKeyVault` linked service** (`hub-app.bicep:201`, `linkedService_keyVault`) — ADF authenticates as its own system-assigned managed identity at runtime to read the `remoteHubStorage` connection secret via `AzureKeyVaultSecret`. This is the **only** data-plane secret reader, and it's exactly what the pre-existing `keyVaultAccessPolicies` var (`secrets: ['get']`) granted. Now granted via RBAC (`Key Vault Secrets User`) when `enableRbacAuthorization: true`.
2. **The `keyVault_secret` module** (`fx/hub-vault.bicep`, invoked from `RemoteHub/app.bicep:59`) writes the storage key as a nested ARM `Microsoft.KeyVault/vaults/secrets` resource. This is a management-plane (ARM) write authorized by the *deploying principal's* Azure RBAC role on the resource group (e.g., Contributor), not by the vault's own access-policy/RBAC-authorization setting — so it is unaffected either way.
3. No other hub app (`Core`, `Exports`, `Analytics`, `AzureResourceGraph`, etc.) references this Key Vault at all.

So the only identity needing a new role assignment is Data Factory's managed identity, which I added.

## PR #1349 (closed, not merged)

Read `gh pr diff 1349` before starting. It only added `enablePurgeProtection` (12 lines across `main.bicep`/`hub.bicep`/`keyVault.bicep`) against the **pre-reorganization** template layout — those exact file paths (`modules/keyVault.bicep`) no longer exist; the Key Vault resource now lives in `modules/fx/hub-app.bicep` as part of the namespace-based `fx`/`Microsoft.FinOpsHubs` restructuring. It didn't touch RBAC at all, and its purge-protection change has since been superseded by a more complete implementation already on `dev`/`v15-prep`. I did not build on it — nothing to build on for RBAC, and the purge-protection piece it targeted was already done more thoroughly elsewhere.

## Design decision: opt-in, not forced-on

Both `enableRbacAuthorization` and `enablePurgeProtection` are **irreversible per Azure's API contract**:
- Once `enablePurgeProtection: true` is set on a vault, it cannot be reverted.
- Switching `enableRbacAuthorization` from `false` to `true` on an *existing* vault immediately stops honoring access policies; any identity not covered by an equivalent RBAC role assignment loses access with no rollback path (short of recreating the vault).

Because the FinOps hub template supports redeploying over existing hub instances (upgrade scenario), forcing either property on unconditionally would silently break already-deployed hubs that don't opt in, with no way back. I followed the existing precedent set by `enablePurgeProtection` (already opt-in, defaulting to `false`, surfaced identically in `main.bicep`/`hub.bicep`/`createUiDefinition.json`) and applied the same pattern to `enableRbacAuthorization`. Organizations that need CAF/Enterprise-Scale compliance can set both parameters to `true` explicitly; organizations upgrading an existing hub are not force-migrated into a breaking, irreversible change they didn't ask for.

I believe this is the safer default, but it does mean the compliance gap in #1067 isn't closed *by default* — only when explicitly enabled. If maintainers prefer defaulting new deployments to `true` while keeping upgrades safe, that would need a way to distinguish first-deploy from redeploy, which Bicep can't do natively (no reliable "does this resource already exist" check without a `existing` lookup that fails hard on first deploy). Flagging this as an open discussion point for reviewers.

## Verification

- `bicep build src/templates/finops-hub/main.bicep --stdout` — clean, no errors (2 pre-existing unrelated warnings in `Recommendations/app.bicep`, confirmed present before this change via `git stash`).
- `bicep build` on `hub.bicep`, `hub-app.bicep`, and `hub-types.bicep` individually — all clean.
- `pwsh -Command "./src/scripts/Test-PowerShell.ps1 -Lint"` — 3418/3418 passed, including `ms.date` frontmatter checks (already current at `08/12/2026`, no update needed).
- Did not run a live `az deployment ... what-if` — no Azure credentials available in this environment. Everything else that could be verified statically was.

## Open questions for reviewers

1. Should new (first-time) deployments default `enableRbacAuthorization`/`enablePurgeProtection` to `true` while upgrades stay opt-in? Flagged above as not straightforwardly expressible in Bicep.
2. Confirm `Key Vault Secrets User` is the right role scope (read-only `get`/`list` on secrets) rather than `Key Vault Secrets Officer` — I matched the existing access policy's `secrets: ['get']` scope exactly, so `Secrets User` (read-only) seemed correct, but worth a second look given ADF also needs to enumerate the linked service's secret at authoring/refresh time.

## Test plan

- [x] Lint tests (PowerShell / bicep build)
- [ ] PS -WhatIf / az validate (no Azure credentials in this environment)
- [ ] Manually deployed + verified
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/docs-mslearn/toolkit/changelog.md
+++ b/docs-mslearn/toolkit/changelog.md
@@ -29,6 +29,7 @@ The following section lists features and enhancements that are currently in deve
 
 - **Added**
   - Added VNet and private network modes, including opt-in NAT Gateway support for private mode; NAT Gateway incurs additional cost when enabled ([#2163](https://github.com/microsoft/finops-toolkit/pull/2163)).
+  - Added an opt-in `enableRbacAuthorization` parameter to switch the remote hub Key Vault from access policies to Azure RBAC, satisfying Cloud Adoption Framework / Enterprise-Scale landing zone guardrails that require RBAC-authorized key vaults; the Data Factory managed identity is granted an equivalent Key Vault Secrets User role assignment so secret access keeps working when enabled ([#1067](https://github.com/microsoft/finops-toolkit/issues/1067)).
 - **Changed**
   - Replaced redundant `tolower()` comparisons in hub KQL with case-insensitive operators (`has`, `=~`, `!~`) so the engine can use the term index instead of scanning every row ([#2213](https://github.com/microsoft/finops-toolkit/issues/2213)).
   - Replaced whole-term `contains` matches with `has` across hub KQL and the query catalog (resource ID paths, licensing phrases, SKU description terms) and added a per-row operator-equivalence regression harness with unit test coverage ([#2220](https://github.com/microsoft/finops-toolkit/pull/2220)).
--- a/docs-mslearn/toolkit/changelog.md
+++ b/docs-mslearn/toolkit/changelog.md
@@ -3,7 +3,7 @@ title: FinOps toolkit changelog
 description: Review the latest features and enhancements in the FinOps toolkit, including updates to FinOps hubs, Power BI reports, and more.
 author: MSBrett
 ms.author: brettwil
-ms.date: 08/12/2026
+ms.date: 08/26/2026
 ms.topic: reference
 ms.service: finops
 ms.subservice: finops-toolkit
@@ -29,7 +29,7 @@ The following section lists features and enhancements that are currently in deve
 
 - **Added**
   - Added VNet and private network modes, including opt-in NAT Gateway support for private mode; NAT Gateway incurs additional cost when enabled ([#2163](https://github.com/microsoft/finops-toolkit/pull/2163)).
-  - Added an opt-in `enableRbacAuthorization` parameter to switch the remote hub Key Vault from access policies to Azure RBAC, satisfying Cloud Adoption Framework / Enterprise-Scale landing zone guardrails that require RBAC-authorized key vaults; the Data Factory managed identity is granted an equivalent Key Vault Secrets User role assignment so secret access keeps working when enabled ([#1067](https://github.com/microsoft/finops-toolkit/issues/1067)).
+  - Added an opt-in `enableKeyVaultRbacAuthorization` parameter to switch the remote hub Key Vault from access policies to Azure RBAC, satisfying Cloud Adoption Framework / Enterprise-Scale landing zone guardrails that require RBAC-authorized key vaults; the Data Factory managed identity is granted an equivalent Key Vault Secrets User role assignment so secret access keeps working when enabled ([#1067](https://github.com/microsoft/finops-toolkit/issues/1067)).
 - **Changed**
   - Replaced redundant `tolower()` comparisons in hub KQL with case-insensitive operators (`has`, `=~`, `!~`) so the engine can use the term index instead of scanning every row ([#2213](https://github.com/microsoft/finops-toolkit/issues/2213)).
   - Replaced whole-term `contains` matches with `has` across hub KQL and the query catalog (resource ID paths, licensing phrases, SKU description terms) and added a per-row operator-equivalence regression harness with unit test coverage ([#2220](https://github.com/microsoft/finops-toolkit/pull/2220)).
--- a/src/templates/finops-hub/createUiDefinition.json
+++ b/src/templates/finops-hub/createUiDefinition.json
@@ -207,7 +207,7 @@
             "visible": "[equals(basics('analyticsBackend').analyticsEngine, 'remote')]"
           },
           {
-            "name": "enableRbacAuthorization",
+            "name": "enableKeyVaultRbacAuthorization",
             "type": "Microsoft.Common.CheckBox",
             "label": "Enable Key Vault RBAC authorization",
             "toolTip": "Enables Azure RBAC instead of access policies to authorize access to the Key Vault used to store the remote hub storage key. Required by some organizations for policy compliance (e.g., Cloud Adoption Framework guardrails). Enable this if you are deploying to a subscription that enforces this requirement.",
@@ -1020,7 +1020,7 @@
       "remoteHubStorageUri": "[if(equals(basics('analyticsBackend').analyticsEngine, 'remote'), basics('analyticsBackend').remoteHubStorageUri, '')]",
       "remoteHubStorageKey": "[if(equals(basics('analyticsBackend').analyticsEngine, 'remote'), basics('analyticsBackend').remoteHubStorageKey, '')]",
       "enablePurgeProtection": "[if(equals(basics('analyticsBackend').analyticsEngine, 'remote'), coalesce(basics('analyticsBackend').enablePurgeProtection, false), false)]",
-      "enableRbacAuthorization": "[if(equals(basics('analyticsBackend').analyticsEngine, 'remote'), coalesce(basics('analyticsBackend').enableRbacAuthorization, false), false)]",
+      "enableKeyVaultRbacAuthorization": "[if(equals(basics('analyticsBackend').analyticsEngine, 'remote'), coalesce(basics('analyticsBackend').enableKeyVaultRbacAuthorization, false), false)]",
       "tagsByResource": "[steps('tags').tagsByResource]"
     }
   }
--- a/src/templates/finops-hub/createUiDefinition.json
+++ b/src/templates/finops-hub/createUiDefinition.json
@@ -203,7 +203,14 @@
             "name": "enablePurgeProtection",
             "type": "Microsoft.Common.CheckBox",
             "label": "Enable Key Vault purge protection",
-            "toolTip": "Enables purge protection for the Key Vault used to store the remote hub storage key. Purge protection prevents permanent deletion of the Key Vault for 90 days after deletion. Note: If the key is lost, you can regenerate it from the remote hub's storage account.",
+            "toolTip": "Enables purge protection for the Key Vault used to store the remote hub storage key. Purge protection prevents permanent deletion of the Key Vault for 90 days after deletion. Note: If the key is lost, you can regenerate it from the remote hub's storage account. This cannot be disabled once enabled.",
+            "visible": "[equals(basics('analyticsBackend').analyticsEngine, 'remote')]"
+          },
+          {
+            "name": "enableRbacAuthorization",
+            "type": "Microsoft.Common.CheckBox",
+            "label": "Enable Key Vault RBAC authorization",
+            "toolTip": "Enables Azure RBAC instead of access policies to authorize access to the Key Vault used to store the remote hub storage key. Required by some organizations for policy compliance (e.g., Cloud Adoption Framework guardrails). Enable this if you are deploying to a subscription that enforces this requirement.",
             "visible": "[equals(basics('analyticsBackend').analyticsEngine, 'remote')]"
           }
         ],
@@ -1013,6 +1020,7 @@
       "remoteHubStorageUri": "[if(equals(basics('analyticsBackend').analyticsEngine, 'remote'), basics('analyticsBackend').remoteHubStorageUri, '')]",
       "remoteHubStorageKey": "[if(equals(basics('analyticsBackend').analyticsEngine, 'remote'), basics('analyticsBackend').remoteHubStorageKey, '')]",
       "enablePurgeProtection": "[if(equals(basics('analyticsBackend').analyticsEngine, 'remote'), coalesce(basics('analyticsBackend').enablePurgeProtection, false), false)]",
+      "enableRbacAuthorization": "[if(equals(basics('analyticsBackend').analyticsEngine, 'remote'), coalesce(basics('analyticsBackend').enableRbacAuthorization, false), false)]",
       "tagsByResource": "[steps('tags').tagsByResource]"
     }
   }
--- a/src/templates/finops-hub/main.bicep
+++ b/src/templates/finops-hub/main.bicep
@@ -26,8 +26,11 @@ param storageSku string = 'Premium_LRS'
 @description('Optional. Enable infrastructure encryption on the storage account. Default = false.')
 param enableInfrastructureEncryption bool = false
 
-@description('Optional. Enable purge protection for the Key Vault. Default: false.')
+@description('Optional. Enable purge protection for the Key Vault. Once enabled on a vault, purge protection cannot be disabled. Default: false.')
 param enablePurgeProtection bool = false
+
+@description('Optional. Enable Azure RBAC for authorizing access to the Key Vault instead of access policies. Required by some organizations for policy compliance (e.g., Cloud Adoption Framework guardrails). Switching an existing vault from access policies to RBAC has migration implications, so this defaults to false for backward compatibility with existing deployments; review before enabling on an upgrade. Default: false.')
+param enableRbacAuthorization bool = false
 
 @description('Optional. Storage account to push data to for ingestion into a remote hub.')
 param remoteHubStorageUri string = ''
@@ -179,6 +182,7 @@ module hub 'modules/hub.bicep' = {
     storageSku: storageSku
     enableInfrastructureEncryption: enableInfrastructureEncryption
     enablePurgeProtection: enablePurgeProtection
+    enableRbacAuthorization: enableRbacAuthorization
     enableManagedExports: enableManagedExports
     enableRecommendations: enableRecommendations
     enableAHBRecommendations: enableAHBRecommendations
--- a/src/templates/finops-hub/main.bicep
+++ b/src/templates/finops-hub/main.bicep
@@ -30,7 +30,7 @@ param enableInfrastructureEncryption bool = false
 param enablePurgeProtection bool = false
 
 @description('Optional. Enable Azure RBAC for authorizing access to the Key Vault instead of access policies. Required by some organizations for policy compliance (e.g., Cloud Adoption Framework guardrails). Switching an existing vault from access policies to RBAC has migration implications, so this defaults to false for backward compatibility with existing deployments; review before enabling on an upgrade. Default: false.')
-param enableRbacAuthorization bool = false
+param enableKeyVaultRbacAuthorization bool = false
 
 @description('Optional. Storage account to push data to for ingestion into a remote hub.')
 param remoteHubStorageUri string = ''
@@ -182,7 +182,7 @@ module hub 'modules/hub.bicep' = {
     storageSku: storageSku
     enableInfrastructureEncryption: enableInfrastructureEncryption
     enablePurgeProtection: enablePurgeProtection
-    enableRbacAuthorization: enableRbacAuthorization
+    enableKeyVaultRbacAuthorization: enableKeyVaultRbacAuthorization
     enableManagedExports: enableManagedExports
     enableRecommendations: enableRecommendations
     enableAHBRecommendations: enableAHBRecommendations
--- a/src/templates/finops-hub/modules/fx/hub-app.bicep
+++ b/src/templates/finops-hub/modules/fx/hub-app.bicep
@@ -83,8 +83,9 @@ var storageInfrastructureEncryptionProperties = !app.hub.options.storageInfrastr
   }
 }
 
-// KeyVault access policies
-var keyVaultAccessPolicies = [
+// KeyVault access policies -- only used when the vault uses the legacy access-policy auth model
+// (RBAC-authorized vaults must have an empty accessPolicies array; Azure rejects a non-empty array otherwise)
+var keyVaultAccessPolicies = app.hub.options.keyVaultEnableRbacAuthorization ? [] : [
   {
     #disable-next-line BCP318 // Null safety warning for conditional resource access // Null safety warning for conditional resource access
     objectId: dataFactory.identity.principalId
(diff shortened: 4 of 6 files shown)


src/templates/finops-hub/modules/hub.bicep
```diff
@@ -195,6 +198,7 @@ var hub = newHub(
   storageSku,
   keyVaultSku,
   enablePurgeProtection,
+  enableRbacAuthorization,
```
flanakin, line 201: Is this name too generic? This isn't clear it's for Key Vault only.

src/templates/finops-hub/createUiDefinition.json
```diff
@@ -203,7 +203,14 @@
...
             "type": "Microsoft.Common.CheckBox",
             "label": "Enable Key Vault purge protection",
-            "toolTip": "Enables purge protection for the Key Vault used to store the remote hub storage key. Purge protection prevents permanent deletion of the Key Vault for 90 days after deletion. Note: If the key is lost, you can regenerate it from the remote hub's storage account.",
+            "toolTip": "Enables purge protection for the Key Vault used to store the remote hub storage key. Purge protection prevents permanent deletion of the Key Vault for 90 days after deletion. Note: If the key is lost, you can regenerate it from the remote hub's storage account. This cannot be disabled once enabled.",
+            "visible": "[equals(basics('analyticsBackend').analyticsEngine, 'remote')]"
+          },
+          {
+            "name": "enableRbacAuthorization",
+            "type": "Microsoft.Common.CheckBox",
+            "label": "Enable Key Vault RBAC authorization",
```
RolandKrummenacher, line 212: What happens on existing deployments? Will it end up with both, rbac and access policies, as we are not removing access policies via bicep I guess.

RolandKrummenacher (commented):
**Review — PR #2249**

Approving the approach. The consumer trace is the part that matters most here and it holds up: the ADF `AzureKeyVault` linked service is the only data-plane secret reader, the `keyVault_secret` module write is management-plane and genuinely unaffected by the vault's authorization model, and no other hub app touches this vault. Granting exactly one `Key Vault Secrets User` assignment is the right scope — not a broader role, not a second assignment for the deploying principal.

Opt-in defaulting to `false` is also right, and for the reason stated: flipping `enableRbacAuthorization` on an existing vault silently stops honoring access policies with no rollback short of recreating the vault, and this template is explicitly redeploy-over-existing. Forcing it on would break deployed hubs at upgrade time. Following the `enablePurgeProtection` precedent keeps that consistent.

Two notes, neither blocking.

### `newHub()` positional-argument fragility

`keyVaultEnableRbacAuthorization` is inserted mid-signature in both `newHubInternal` and `newHub`, and both are called positionally. There's exactly one caller today (`modules/hub.bicep:190`), so this is safe as written — I checked. But the signature is now 15+ positional bools and strings, and the next insertion is one where a mis-ordered pair of same-typed args compiles clean and misconfigures a hub silently. Not this PR's job to fix, but worth an issue: these should take a single object parameter, or at minimum new options should append rather than insert.

### RBAC propagation vs. access-policy immediacy

Access policy grants take effect essentially immediately; RBAC role assignments can take a few minutes to propagate. ADF reads the secret at pipeline runtime rather than at deployment time, so in the normal case the assignment has long since propagated by the time anything needs it. The edge case is a pipeline triggered immediately after a first deployment with the flag on — that could see a transient 403 that looks like a misconfiguration rather than a timing artifact.

Not worth adding a `dependsOn` dance over, but if the deployment surfaces a "hub is ready" message anywhere, a sentence in the docs noting that RBAC-authorized vaults may need a few minutes before the first ingestion run succeeds would save someone a support round-trip.

### Small thing

The `enablePurgeProtection` tooltip gained "This cannot be disabled once enabled" — good addition. The `enableRbacAuthorization` tooltip doesn't carry the equivalent warning, and its irreversibility is arguably the sharper one of the two (purge protection costs you a 90-day wait; a botched RBAC switch costs you vault access). Consider mirroring the wording from the `main.bicep` parameter description, which does say it.

src/templates/finops-hub/modules/hub.bicep
flanakin: 🤖 [AI][Claude] ✅ **Implemented**

Yes, it was too generic. Renamed `enableRbacAuthorization` → `enableKeyVaultRbacAuthorization` in `main.bicep` and `hub.bicep` (parameter declaration + the `newHub()` call site), and the matching `createUiDefinition.json` element name/output binding, for consistency with the already-correctly-scoped `keyVaultEnableRbacAuthorization` used internally in `hub-types.bicep`/`hub-app.bicep`. Left the literal `enableRbacAuthorization: ...` property in `hub-app.bicep` untouched — that one is the actual ARM schema property name on `Microsoft.KeyVault/vaults`, not our parameter, so it has to stay as-is. Didn't touch `enablePurgeProtection`: purge protection is an unambiguous, Key-Vault-only Azure term, so it doesn't have the same generic-RBAC ambiguity.

src/templates/finops-hub/createUiDefinition.json
flanakin: 🤖 [AI][Claude] 💬 **Needs discussion**

Traced this precisely: no, it will not end up with both. `accessPolicies` is set as a direct `properties.accessPolicies` array on the `Microsoft.KeyVault/vaults` resource itself (`hub-app.bicep`), not via the separate additive `Microsoft.KeyVault/vaults/accessPolicies` child resource type. Since the `keyVault` resource is redeclared on every deployment where the `RemoteHub`/`KeyVault` feature is enabled (not skipped on redeploy), Incremental mode's "don't touch what's absent from the template" guarantee doesn't apply here — it only protects *other* resources not mentioned in the template, not properties on a resource that *is* declared. Whatever array the template sends fully replaces the live value.

The `keyVaultAccessPolicies` variable already reflects this: it evaluates to `[]` whenever `keyVaultEnableRbacAuthorization` is `true` (see the comment already there: "RBAC-authorized vaults must have an empty accessPolicies array; Azure rejects a non-empty array otherwise"). So flipping the flag to `true` on a redeploy of an existing hub cleanly **replaces** the old access policy with an empty array — the vault ends up RBAC-only, not dual-auth. The new `keyVaultRoleAssignment` resource (also gated on the same flag) grants the equivalent `Key Vault Secrets User` role to Data Factory's identity in the same deployment, so access isn't dropped in the process — it's a clean cutover in one apply.

No code change needed; the existing-deployment behavior is already correct and non-destructive by construction (Azure itself enforces the empty-array requirement, which is what forces the replace). The real risk here isn't dual-auth, it's the one already called out in the PR description and param docs: if any identity *other than* Data Factory's happened to rely on an access-policy grant on this vault, it would lose access with no automatic RBAC equivalent — but the PR's own investigation confirmed Data Factory's managed identity is the only data-plane secret reader, so that gap doesn't apply today.

src/templates/finops-hub/main.bicep
```diff
@@ -26,9 +26,12 @@ param storageSku string = 'Premium_LRS'
 @description('Optional. Enable infrastructure encryption on the storage account. Default = false.')
 param enableInfrastructureEncryption bool = false
 
-@description('Optional. Enable purge protection for the Key Vault. Default: false.')
+@description('Optional. Enable purge protection for the Key Vault. Once enabled on a vault, purge protection cannot be disabled. Default: false.')
 param enablePurgeProtection bool = false
 
+@description('Optional. Enable Azure RBAC for authorizing access to the Key Vault instead of access policies. Required by some organizations for policy compliance (e.g., Cloud Adoption Framework guardrails). Switching an existing vault from access policies to RBAC has migration implications, so this defaults to false for backward compatibility with existing deployments; review before enabling on an upgrade. Default: false.')
+param enableKeyVaultRbacAuthorization bool = false
```
RolandKrummenacher, line 33: This is where the missing surface shows up. `enablePurgeProtection` (three lines up) is reachable three ways: this template, the portal UI, and `Deploy-FinOpsHub -EnablePurgeProtection`. `enableKeyVaultRbacAuthorization` is reachable only the first two ways.

`src/powershell/Public/Deploy-FinOpsHub.ps1` needs the matching switch — it isn't in this PR's diff, so I can't anchor a comment there directly:

- **line ~34** — `.PARAMETER EnableKeyVaultRbacAuthorization` help entry, mirroring the `EnablePurgeProtection` one.
- **line ~146** — the switch declaration next to `$EnablePurgeProtection`.
- **line ~322** — forwarding. `enablePurgeProtection` sits in the `>= '13.0'` gate; this one wants a new `>= '15.0'` block, matching how `enableNatGateway` is gated:

```powershell
if ($Version -eq 'latest' -or [version]$Version -ge '15.0')
{
    $parameterSplat.TemplateParameterObject.Add('enableKeyVaultRbacAuthorization', $EnableKeyVaultRbacAuthorization.IsPresent)
}
```

Worth doing here rather than as a follow-up: CAF/Enterprise-Scale shops are disproportionately the ones deploying via script rather than the portal blade, so the audience for this feature is the audience most likely to hit the gap.

src/templates/finops-hub/createUiDefinition.json
```diff
@@ -203,7 +203,14 @@
...
+            "toolTip": "Enables purge protection for the Key Vault used to store the remote hub storage key. Purge protection prevents permanent deletion of the Key Vault for 90 days after deletion. Note: If the key is lost, you can regenerate it from the remote hub's storage account. This cannot be disabled once enabled.",
+            "visible": "[equals(basics('analyticsBackend').analyticsEngine, 'remote')]"
+          },
+          {
+            "name": "enableKeyVaultRbacAuthorization",
+            "type": "Microsoft.Common.CheckBox",
+            "label": "Enable Key Vault RBAC authorization",
+            "toolTip": "Enables Azure RBAC instead of access policies to authorize access to the Key Vault used to store the remote hub storage key. Required by some organizations for policy compliance (e.g., Cloud Adoption Framework guardrails). Enable this if you are deploying to a subscription that enforces this requirement.",
```
RolandKrummenacher, line 213: Still missing the irreversibility warning I flagged last round. The `enablePurgeProtection` tooltip directly above gained "This cannot be disabled once enabled" in this same PR — this one didn't, and its failure mode is the sharper of the two. Purge protection costs you a 90-day wait; a botched RBAC switch costs you vault access with no rollback short of recreating the vault.

The `main.bicep` parameter description already has usable wording. Something like appending:

> Switching an existing vault from access policies to RBAC takes effect immediately and cannot be rolled back without recreating the vault.

Portal users clicking this checkbox on a redeploy are exactly the people who need that sentence, and they never see the Bicep description.

src/templates/finops-hub/modules/fx/hub-app.bicep
```diff
@@ -495,6 +500,18 @@ resource keyVault 'Microsoft.KeyVault/vaults@2023-02-01' = if (usesKeyVault) {
   }
 }
 
+// Grant ADF identity RBAC access to read secrets when the vault uses RBAC instead of access policies
+resource keyVaultRoleAssignment 'Microsoft.Authorization/roleAssignments@2022-04-01' = if (usesKeyVault && usesDataFactory && app.hub.options.keyVaultEnableRbacAuthorization) {
```
RolandKrummenacher, line 504: Non-blocking, carried over from last review. Access policy grants take effect essentially immediately; RBAC role assignments can take a few minutes to propagate.

In the normal case this is a non-issue — ADF reads the secret at pipeline runtime, long after the assignment has settled. The edge case is a pipeline triggered right after a first deployment with the flag on: a transient 403 that reads like a misconfiguration rather than a timing artifact.

Not worth a `dependsOn` dance. But if the deployment surfaces a "hub is ready" message anywhere, one sentence noting that RBAC-authorized vaults may need a few minutes before the first ingestion run succeeds would save someone a support round-trip.

The gating itself is right, for the record — `usesKeyVault && usesDataFactory && ...keyVaultEnableRbacAuthorization` matches the vault's own condition, and the `guid(keyVault.id, roleId, dataFactory.id)` name is properly deterministic across redeploys.

src/templates/finops-hub/modules/fx/hub-types.bicep
```diff
@@ -253,6 +257,7 @@ func newHub(
   storageSku string,
   keyVaultSku string,
   keyVaultEnablePurgeProtection bool,
+  keyVaultEnableRbacAuthorization bool,
```
RolandKrummenacher, line 260: Repeating from last review since it's unaddressed, and explicitly **not** asking this PR to fix it — logging it so it doesn't evaporate.

`keyVaultEnableRbacAuthorization` is inserted mid-signature in both `newHubInternal` and `newHub`, and both are called positionally. There is exactly one caller today (`modules/hub.bicep:193`), so this insertion is safe — I checked it. But the signature is now 15+ positional bools and strings, and a mis-ordered pair of same-typed args would compile clean and silently misconfigure a hub.

Worth a standalone issue: these should take a single object parameter, or at minimum new options should append rather than insert. Happy to open it if nobody objects.

docs-mslearn/toolkit/changelog.md
```diff
@@ -29,6 +29,7 @@ The following section lists features and enhancements that are currently in deve
 
 - **Added**
   - Added VNet and private network modes, including opt-in NAT Gateway support for private mode; NAT Gateway incurs additional cost when enabled ([#2163](https://github.com/microsoft/finops-toolkit/pull/2163)).
+  - Added an opt-in `enableKeyVaultRbacAuthorization` parameter to switch the remote hub Key Vault from access policies to Azure RBAC, satisfying Cloud Adoption Framework / Enterprise-Scale landing zone guardrails that require RBAC-authorized key vaults; the Data Factory managed identity is granted an equivalent Key Vault Secrets User role assignment so secret access keeps working when enabled ([#1067](https://github.com/microsoft/finops-toolkit/issues/1067)).
```
RolandKrummenacher, line 32: This entry is the merge conflict against `flanakin/v15-prep`. I ran the merge locally — it's the only one, and it's the routine pattern: this line plus the neighboring Unreleased entries, and `ms.date` at the top of the file.

Per the repo's conflict guidance: `git merge origin/flanakin/v15-prep`, keep entries from **both** sides (they're additive and independent), and set `ms.date` to today's date rather than either side's value.

RolandKrummenacher (commented):
**Re-review — PR #2249**

One new commit since my 2026-08-18 review (`3d9b06ab`, the rename), plus the two replies. Approach and implementation still hold up; one gap worth closing before merge.

### Resolved

**Rename `enableRbacAuthorization` → `enableKeyVaultRbacAuthorization`** — complete. Grepped the branch: the only surviving `enableRbacAuthorization` is the ARM schema property on `Microsoft.KeyVault/vaults` in `hub-app.bicep`, correctly left alone. Parameter declaration, `newHub()` call site, createUiDefinition element name *and* output binding, and the changelog entry all agree.

**Dual-auth question** — the answer is correct, and I verified the mechanism rather than taking it on faith. `accessPolicies` is a direct property on the vault resource, not the additive `Microsoft.KeyVault/vaults/accessPolicies` child type, and the vault is redeclared on every deploy. Incremental mode's "leave absent things alone" guarantee covers resources, not properties of a resource that *is* declared — so the `[]` array genuinely replaces the live policy. Clean cutover, no dual-auth state. Resolving that thread.

### One gap: the PowerShell deployment path

`Deploy-FinOpsHub` exposes purge protection but not this. `src/powershell/Public/Deploy-FinOpsHub.ps1` declares `-EnablePurgeProtection` as a switch (line 146), documents it (line 34), and forwards it under a `>= '13.0'` version gate (line 322). There is no equivalent for `enableKeyVaultRbacAuthorization`, so anyone deploying through the PowerShell module can't enable RBAC at all — portal and raw Bicep only.

The PR follows "the exact pattern already used for `enablePurgeProtection`," and that pattern spans four surfaces: `main.bicep`, `hub.bicep`, `createUiDefinition.json`, **and** the cmdlet. Three of four are done. The fix is small, with a template right beside it:

```powershell
[Parameter()]
[switch]
$EnableKeyVaultRbacAuthorization,
```

plus a `>= '15.0'` gate block (matching how `enableNatGateway` is gated — branch is `15.0.0-dev.0`) and a `.PARAMETER` help entry. I'd treat this as blocking-ish: the module is the scripted-deployment path for exactly the CAF/landing-zone customers this feature targets. Commented inline on the param declaration since the file itself isn't in the diff.

### Still open from last review

Details inline: the tooltip irreversibility warning (actionable), RBAC propagation timing (cosmetic), and `newHub()` positional fragility (separate issue, not this PR's job).

### Housekeeping

PR shows `CONFLICTING` against `flanakin/v15-prep`. I ran the merge locally — the only conflict is `changelog.md`. Details inline.

**Verdict:** sound work and a clean rename. Add the `Deploy-FinOpsHub` switch and the tooltip sentence, merge the base in, and this is ready to go.

src/templates/finops-hub/modules/hub.bicep
RolandKrummenacher: Verified on the branch: the rename is complete and consistent. The only remaining `enableRbacAuthorization` is the ARM schema property on `Microsoft.KeyVault/vaults` in `hub-app.bicep`, which correctly stays as-is. Parameter declaration, `newHub()` call site, createUiDefinition element name and output binding all agree with `main.bicep`. Agreed on leaving `enablePurgeProtection` alone — it is unambiguous. Resolving.

src/templates/finops-hub/createUiDefinition.json
RolandKrummenacher: Confirmed independently — no dual-auth state, and for the reason given. `accessPolicies` is a direct property on the vault resource rather than the additive `Microsoft.KeyVault/vaults/accessPolicies` child type, and the vault is redeclared on every deploy, so Incremental mode does not protect it: the array the template sends fully replaces the live value. Flipping the flag on a redeploy cleanly replaces the old policy with `[]`, and the `Key Vault Secrets User` assignment lands in the same apply, so ADF never loses access mid-cutover.

The residual risk is the one already documented in the param description — any identity outside the template that was granted an access policy on this vault by hand loses access with no RBAC equivalent. That is inherent to the switch, not a defect here, and the consumer trace shows it does not apply to anything the template creates. Question is answered, so resolving; happy to reopen if you want that upgrade caveat called out in the docs rather than only in the parameter description.

PR #2250, opened 2026-08-12
Bump pinned Az PowerShell module versions in CI workflows

### LINKED ISSUE #945: [Hubs] Update Azure PowerShell version for deployment scripts
## ⚠️ Problem
Deployment scripts are returning this warning:
> WARNING: You're using Az version 8.0.0. The latest version of Az is 12.2.0. Upgrade your Az modules using the following commands:
> - Update-Module Az -WhatIf    -- Simulate updating your Az modules.
> - Update-Module Az            -- Update your Az modules.
> 
> There will be breaking changes from 8.0.0 to 12.2.0. Open https://go.microsoft.com/fwlink/?linkid=2241373 and check the details.

## 🛠️ Solution
Update deployment scripts to use the latest Azure PowerShell release.

## 🙋‍♀️ Ask for the community

We could use your help:

flanakin:
## Summary

Issue #945 reports the deployment scripts printing Az PowerShell's built-in outdated-version warning ("You're using Az version 8.0.0. The latest version of Az is 12.2.0..."). That message is emitted by `Az.Accounts` itself at runtime, comparing the locally installed `Az` module against PSGallery — it is not text this repo generates, so there's no code path in `src/scripts/` or `src/powershell/` that "prints" it or that can be patched directly.

What *is* fixable in-repo: several CI workflows pin specific, multi-year-stale `Az.*` module versions via `Install-Module -RequiredVersion` / `AZ_*_VERSION` env vars, which would themselves trigger that same nag during CI runs (and give contributors who copy those pins into local scripts an outdated baseline). This PR bumps every one of those pins to the current PSGallery stable release, consistently across all files that reference them:

| Module | Old | New |
|---|---|---|
| `Az.Accounts` | 2.19.0 | 5.5.2 |
| `Az.Resources` | 6.16.2 | 10.1.0 |
| `Az.Storage` | 6.2.0 | 9.7.2 |
| `Az.OperationalInsights` | 3.2.0 | 3.4.1 |
| `Az.Sql` | 4.14.1 | 7.0.0 |
| `Az.Automation` | 1.10.0 | 1.12.1 |

Files updated:
- `.github/workflows/ftk-pr-deploy.yml` (env vars `AZ_ACCOUNTS_VERSION`, `AZ_RESOURCES_VERSION`, `AZ_STORAGE_VERSION`, consumed by all deploy jobs + cache keys)
- `.github/workflows/ftk-pr-cleanup.yml`
- `.github/workflows/opendata-instance-size-flexibility.yml`
- `.github/workflows/aoe-cd-dev.yml`, `aoe-cd-test.yml`, `aoe-cd-prod.yml`

No other `RequiredVersion`/`MinimumVersion`/`RequiredModules` pins exist elsewhere in the repo (checked `src/scripts/`, `src/powershell/`, all `.psd1` manifests). Docs (`docs/powershell.md`, `docs-mslearn/**`) use unpinned `Install-Module -Name Az.Accounts` / `Az.Resources`, which is already correct and untouched.

References but does not close #945, since the actual user-facing warning is Az PowerShell's own runtime nag based on whatever `Az` version is installed on the machine running the deployment script — not something this repo enforces or can silence in code. Recommend closing #945 once this merges, with a note in the issue that the fix is "keep CI pins current" rather than "suppress the nag."

## Test plan

- [x] `pwsh -Command "./src/scripts/Test-PowerShell.ps1 -Lint"` — 3418/3418 passed
- [x] Diffs reviewed — mechanical version-string substitutions only, no structural YAML changes
- [x] Verified no remaining references to the old version strings across `.github/workflows/`

### DIFF
--- a/.github/workflows/aoe-cd-dev.yml
+++ b/.github/workflows/aoe-cd-dev.yml
@@ -24,12 +24,12 @@ jobs:
         shell: pwsh
         run: |
           Set-PSRepository PSGallery -InstallationPolicy Trusted
-          Install-Module -Name Az.Accounts -RequiredVersion 2.19.0 -Force
-          Install-Module -Name Az.Resources -RequiredVersion 6.16.2 -Force
-          Install-Module -Name Az.Storage -RequiredVersion 6.2.0 -Force
-          Install-Module -Name Az.OperationalInsights -RequiredVersion 3.2.0 -Force
-          Install-Module -Name Az.Sql -RequiredVersion 4.14.1 -Force
-          Install-Module -Name Az.Automation -RequiredVersion 1.10.0 -Force
+          Install-Module -Name Az.Accounts -RequiredVersion 5.5.2 -Force
+          Install-Module -Name Az.Resources -RequiredVersion 10.1.0 -Force
+          Install-Module -Name Az.Storage -RequiredVersion 9.7.2 -Force
+          Install-Module -Name Az.OperationalInsights -RequiredVersion 3.4.1 -Force
+          Install-Module -Name Az.Sql -RequiredVersion 7.0.0 -Force
+          Install-Module -Name Az.Automation -RequiredVersion 1.12.1 -Force
           Install-Module -Name Microsoft.Graph.Authentication,Microsoft.Graph.Identity.DirectoryManagement -Force
       - name: Check out repository code
         uses: actions/checkout@v3
--- a/.github/workflows/aoe-cd-prod.yml
+++ b/.github/workflows/aoe-cd-prod.yml
@@ -24,12 +24,12 @@ jobs:
         shell: pwsh
         run: |
           Set-PSRepository PSGallery -InstallationPolicy Trusted
-          Install-Module -Name Az.Accounts -RequiredVersion 2.19.0 -Force
-          Install-Module -Name Az.Resources -RequiredVersion 6.16.2 -Force
-          Install-Module -Name Az.Storage -RequiredVersion 6.2.0 -Force
-          Install-Module -Name Az.OperationalInsights -RequiredVersion 3.2.0 -Force
-          Install-Module -Name Az.Sql -RequiredVersion 4.14.1 -Force
-          Install-Module -Name Az.Automation -RequiredVersion 1.10.0 -Force
+          Install-Module -Name Az.Accounts -RequiredVersion 5.5.2 -Force
+          Install-Module -Name Az.Resources -RequiredVersion 10.1.0 -Force
+          Install-Module -Name Az.Storage -RequiredVersion 9.7.2 -Force
+          Install-Module -Name Az.OperationalInsights -RequiredVersion 3.4.1 -Force
+          Install-Module -Name Az.Sql -RequiredVersion 7.0.0 -Force
+          Install-Module -Name Az.Automation -RequiredVersion 1.12.1 -Force
           Install-Module -Name Microsoft.Graph.Authentication,Microsoft.Graph.Identity.DirectoryManagement -Force
       - name: Check out repository code
         uses: actions/checkout@v3
--- a/.github/workflows/aoe-cd-test.yml
+++ b/.github/workflows/aoe-cd-test.yml
@@ -24,12 +24,12 @@ jobs:
         shell: pwsh
         run: |
           Set-PSRepository PSGallery -InstallationPolicy Trusted
-          Install-Module -Name Az.Accounts -RequiredVersion 2.19.0 -Force
-          Install-Module -Name Az.Resources -RequiredVersion 6.16.2 -Force
-          Install-Module -Name Az.Storage -RequiredVersion 6.2.0 -Force
-          Install-Module -Name Az.OperationalInsights -RequiredVersion 3.2.0 -Force
-          Install-Module -Name Az.Sql -RequiredVersion 4.14.1 -Force
-          Install-Module -Name Az.Automation -RequiredVersion 1.10.0 -Force
+          Install-Module -Name Az.Accounts -RequiredVersion 5.5.2 -Force
+          Install-Module -Name Az.Resources -RequiredVersion 10.1.0 -Force
+          Install-Module -Name Az.Storage -RequiredVersion 9.7.2 -Force
+          Install-Module -Name Az.OperationalInsights -RequiredVersion 3.4.1 -Force
+          Install-Module -Name Az.Sql -RequiredVersion 7.0.0 -Force
+          Install-Module -Name Az.Automation -RequiredVersion 1.12.1 -Force
           Install-Module -Name Microsoft.Graph.Authentication,Microsoft.Graph.Identity.DirectoryManagement -Force
       - name: Check out repository code
         uses: actions/checkout@v3
--- a/.github/workflows/ftk-pr-deploy.yml
+++ b/.github/workflows/ftk-pr-deploy.yml
@@ -25,9 +25,9 @@ on:
     types: [labeled]
 
 env:
-  AZ_ACCOUNTS_VERSION: '2.19.0'
-  AZ_RESOURCES_VERSION: '6.16.2'
-  AZ_STORAGE_VERSION: '6.2.0'
+  AZ_ACCOUNTS_VERSION: '5.5.2'
+  AZ_RESOURCES_VERSION: '10.1.0'
+  AZ_STORAGE_VERSION: '9.7.2'
 
 concurrency:
   group: ftk-pr-${{ github.event.pull_request.number }}
--- a/.github/workflows/opendata-instance-size-flexibility.yml
+++ b/.github/workflows/opendata-instance-size-flexibility.yml
@@ -31,7 +31,7 @@ jobs:
         shell: pwsh
         run: |
           Set-PSRepository PSGallery -InstallationPolicy Trusted
-          Install-Module -Name Az.Accounts -RequiredVersion 2.19.0 -Force
+          Install-Module -Name Az.Accounts -RequiredVersion 5.5.2 -Force
 
       - name: Login via Az module
         uses: azure/login@264ef4588d3ca547fea5296679ee9f8a4766d4df # hf_447_release
--- a/.github/workflows/ftk-pr-cleanup.yml
+++ b/.github/workflows/ftk-pr-cleanup.yml
@@ -23,8 +23,8 @@ jobs:
         shell: pwsh
         run: |
           Set-PSRepository PSGallery -InstallationPolicy Trusted
-          Install-Module -Name Az.Accounts -RequiredVersion 2.19.0 -Force
-          Install-Module -Name Az.Resources -RequiredVersion 6.16.2 -Force
+          Install-Module -Name Az.Accounts -RequiredVersion 5.5.2 -Force
+          Install-Module -Name Az.Resources -RequiredVersion 10.1.0 -Force
 
       - name: Azure login
         uses: azure/login@v2


.github/workflows/aoe-cd-dev.yml
```diff
@@ -24,12 +24,12 @@ jobs:
...
-          Install-Module -Name Az.Storage -RequiredVersion 6.2.0 -Force
-          Install-Module -Name Az.OperationalInsights -RequiredVersion 3.2.0 -Force
-          Install-Module -Name Az.Sql -RequiredVersion 4.14.1 -Force
-          Install-Module -Name Az.Automation -RequiredVersion 1.10.0 -Force
+          Install-Module -Name Az.Accounts -RequiredVersion 5.5.2 -Force
+          Install-Module -Name Az.Resources -RequiredVersion 10.1.0 -Force
+          Install-Module -Name Az.Storage -RequiredVersion 9.7.2 -Force
+          Install-Module -Name Az.OperationalInsights -RequiredVersion 3.4.1 -Force
+          Install-Module -Name Az.Sql -RequiredVersion 7.0.0 -Force
+          Install-Module -Name Az.Automation -RequiredVersion 1.12.1 -Force
```
flanakin, line 32: Revert these

.github/workflows/aoe-cd-prod.yml
```diff
@@ -24,12 +24,12 @@ jobs:
...
-          Install-Module -Name Az.Storage -RequiredVersion 6.2.0 -Force
-          Install-Module -Name Az.OperationalInsights -RequiredVersion 3.2.0 -Force
-          Install-Module -Name Az.Sql -RequiredVersion 4.14.1 -Force
-          Install-Module -Name Az.Automation -RequiredVersion 1.10.0 -Force
+          Install-Module -Name Az.Accounts -RequiredVersion 5.5.2 -Force
+          Install-Module -Name Az.Resources -RequiredVersion 10.1.0 -Force
+          Install-Module -Name Az.Storage -RequiredVersion 9.7.2 -Force
+          Install-Module -Name Az.OperationalInsights -RequiredVersion 3.4.1 -Force
+          Install-Module -Name Az.Sql -RequiredVersion 7.0.0 -Force
+          Install-Module -Name Az.Automation -RequiredVersion 1.12.1 -Force
```
flanakin, line 32: Revert

.github/workflows/aoe-cd-test.yml
```diff
@@ -24,12 +24,12 @@ jobs:
...
-          Install-Module -Name Az.Storage -RequiredVersion 6.2.0 -Force
-          Install-Module -Name Az.OperationalInsights -RequiredVersion 3.2.0 -Force
-          Install-Module -Name Az.Sql -RequiredVersion 4.14.1 -Force
-          Install-Module -Name Az.Automation -RequiredVersion 1.10.0 -Force
+          Install-Module -Name Az.Accounts -RequiredVersion 5.5.2 -Force
+          Install-Module -Name Az.Resources -RequiredVersion 10.1.0 -Force
+          Install-Module -Name Az.Storage -RequiredVersion 9.7.2 -Force
+          Install-Module -Name Az.OperationalInsights -RequiredVersion 3.4.1 -Force
+          Install-Module -Name Az.Sql -RequiredVersion 7.0.0 -Force
+          Install-Module -Name Az.Automation -RequiredVersion 1.12.1 -Force
```
flanakin, line 32: Revert

.github/workflows/ftk-pr-deploy.yml
```diff
@@ -25,9 +25,9 @@ on:
     types: [labeled]
 
 env:
-  AZ_ACCOUNTS_VERSION: '2.19.0'
-  AZ_RESOURCES_VERSION: '6.16.2'
-  AZ_STORAGE_VERSION: '6.2.0'
+  AZ_ACCOUNTS_VERSION: '5.5.2'
```
RolandKrummenacher, line 28: Worth knowing before deciding on this approach: **the old pins were never actually loaded**, so `Install-Module -RequiredVersion` isn't governing anything here.

In the last `aoe-cd-dev` run the deploy step resolved to `~/.local/share/powershell/Modules/Az.Accounts/5.5.2`, and in the last `opendata` run it resolved to the image's `/usr/share/az_15.6.1/Az.Accounts/5.5.1`. Neither ever loaded the pinned 2.19.0 — `Install-Module -RequiredVersion` installs side-by-side, it doesn't control what gets imported.

Two consequences:

1. Risk here is much lower than 2.19.0 -> 5.5.2 suggests — CI was already effectively running Az.Accounts 5.5.x.
2. The pins aren't buying anything. If the goal is a controlled baseline, either drop them entirely (saves ~30s/job plus the cache step below) or make them binding with an explicit `Import-Module -RequiredVersion` before the scripts run.

I'd lean toward dropping them: the AOE deploy script already depends on unpinned `Az.Monitor` from the runner image for `Get-AzDataCollectionEndpoint`, so the pin set was never complete anyway.

.github/workflows/ftk-pr-deploy.yml
```diff
@@ -25,9 +25,9 @@ on:
...
+  AZ_RESOURCES_VERSION: '10.1.0'
```
RolandKrummenacher, line 29: One behavior change to keep an eye on across this jump: Az.Resources 8.0 release notes say *"[Breaking Change] Updated API version of resource types from latest to DefaultApiVersion for cmdlets '*-AzResource' and 'Invoke-AzResourceAction'."*

We use `Get-AzResource` in 12 places plus `Remove-AzResource` across `src/scripts/` and `src/powershell/`. Given the pins were already being shadowed by the image's Az 15.6.1, this is probably already exercised in practice — but it's the one item in the changelogs that changes behavior rather than failing loudly.

.github/workflows/opendata-instance-size-flexibility.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         shell: pwsh
         run: |
           Set-PSRepository PSGallery -InstallationPolicy Trusted
-          Install-Module -Name Az.Accounts -RequiredVersion 2.19.0 -Force
+          Install-Module -Name Az.Accounts -RequiredVersion 5.5.2 -Force
```
RolandKrummenacher, line 34: Concrete evidence for the point above — the last run of this workflow ([31362431808](https://github.com/microsoft/finops-toolkit/actions/runs/31362431808)) installed `Az.Accounts 2.19.0` and then resolved to:

```
Result: '/usr/share/az_15.6.1/Az.Accounts/5.5.1/Az.Accounts.psd1'
WARNING: You're using Az version 15.6.1. The latest version of Az is 16.2.0.
```

The pin was inert, and the version warning came from the image's `Az` rollup — which this change doesn't affect.

.github/workflows/aoe-cd-dev.yml
```diff
@@ -24,12 +24,12 @@ jobs:
...
-          Install-Module -Name Az.Accounts -RequiredVersion 2.19.0 -Force
-          Install-Module -Name Az.Resources -RequiredVersion 6.16.2 -Force
-          Install-Module -Name Az.Storage -RequiredVersion 6.2.0 -Force
-          Install-Module -Name Az.OperationalInsights -RequiredVersion 3.2.0 -Force
-          Install-Module -Name Az.Sql -RequiredVersion 4.14.1 -Force
-          Install-Module -Name Az.Automation -RequiredVersion 1.10.0 -Force
+          Install-Module -Name Az.Accounts -RequiredVersion 5.5.2 -Force
+          Install-Module -Name Az.Resources -RequiredVersion 10.1.0 -Force
+          Install-Module -Name Az.Storage -RequiredVersion 9.7.2 -Force
+          Install-Module -Name Az.OperationalInsights -RequiredVersion 3.4.1 -Force
```
RolandKrummenacher, line 30: Minor: this now diverges from `src/optimization-engine/azuredeploy-nested.bicep:868`, which still pins `Az.OperationalInsights 3.2.0` for the Automation account. The deploy-time and runtime versions used to match. Not a problem, but flagging in case the match was intentional.

RolandKrummenacher (commented):
Checked this against the actual CI logs. The diff itself is safe — but two claims in the description need correcting, and the change doesn't deliver the benefit it describes.

### The pins aren't the source of the nag

From the last successful `aoe-cd-dev` run ([31610513279](https://github.com/microsoft/finops-toolkit/actions/runs/31610513279)) and the last `opendata` run ([31362431808](https://github.com/microsoft/finops-toolkit/actions/runs/31362431808)):

```
Result: '/usr/share/az_15.6.1/Az.Accounts/5.5.1/Az.Accounts.psd1'
WARNING: You're using Az version 15.6.1. The latest version of Az is 16.2.0.
```

The runners ship the `Az` rollup at 15.6.1, and the warning is emitted by that rollup during `azure/login` — before the "Installing modules" step runs. Bumping individual `Az.*` pins doesn't change the rollup version, so the same warning will still appear after this merges. See the inline notes for the follow-on point that the old pins were never actually loaded.

### "No other pins exist elsewhere in the repo" isn't accurate

The grep looked for `RequiredVersion`/`MinimumVersion` and missed the PSGallery package-URL pins in `src/optimization-engine/azuredeploy-nested.bicep` (L855-904) and `src/optimization-engine/upgrade-manifest.json`, which set the Automation account's module versions:

| Module | Pinned |
|---|---|
| Az.Accounts | 2.12.1 |
| Az.Compute | 5.7.0 |
| Az.Resources | 6.6.0 |
| Az.Storage | 5.5.0 |
| Az.OperationalInsights | 3.2.0 |
| Az.ResourceGraph | 0.13.0 |
| Az.Monitor | 4.4.1 |

Those are the versions that actually execute AOE in production, and they're staler than the CI pins. They shouldn't be bumped naively, though: the runbooks under `src/optimization-engine/runbooks/` call `Get-AzAccessToken` without `-AsSecureString` (e.g. `Ingest-OptimizationCSVExportsToLogAnalytics.ps1:156`), which breaks on Az.Accounts >= 5.0 ("Changed the default output access token of `Get-AzAccessToken` from plain text to `SecureString`"). Leaving them alone is the right call here — just worth correcting the claim and tracking separately.

### Compatibility checks — no blockers

- All six new versions are the current PSGallery stable.
- Inter-module deps resolve: Az.Resources 10.1.0 and Az.Storage 9.7.2 both require Az.Accounts >= 5.5.2.
- Every Az cmdlet used by `Deploy-AzureOptimizationEngine.ps1`, `src/scripts/`, and `src/powershell/` still exists in the new versions — including the `Select-AzSubscription` alias (verified in the 5.5.2 manifest), `Remove-AzDataLakeGen2Item`, and the AD cmdlets (now exported as functions after the Az.Resources 9.0 codegen change).
- `Get-AzAccessToken -AsSecureString` is already used in `src/powershell/Private/Invoke-Rest.ps1:67` and `Deploy-AzureOptimizationEngine.ps1:1276`.
- `Get-AzStorageBlobContent` is used via pipeline (`Add-FinOpsHubScope.ps1:71`), not the `-AbsoluteUri` form restricted in Az.Storage 8.0.
- Az.Sql 7.0 removed `EnableSoftDelete` from `New/Set-AzSqlServer` — unused here.

### Validation gap

All PR Deploy checks are SKIPPED (label-gated) and `Test-PowerShell.ps1 -Lint` doesn't cover workflow YAML, so nothing in this PR was actually exercised. `aoe-cd-dev` runs unattended on push to `dev`, so it ends up as the canary. Worth applying the deploy labels before merging.

flanakin (commented):
These changes may be fine. I'm tempted to not touch AOE (unless Helder confirms these are okay), but this doesn't touch anything to do with the original issue. The original issue was referring to deployment scripts as part of Azure bicep deployment templates. This has nothing to do with CI/CD deployments. I'm not sure we need any of these changes, tho they may be fine, assuming they've been validated.

MSBrett (approved):
The version updates themselves look safe.

However, they **won’t fix the warning described in issue #945** because that warning comes from a different Azure PowerShell installation. Merge only as routine dependency maintenance, and keep #945 open.

MSBrett (approved):

flanakin:
Closing this PR. On reflection, it targeted the wrong surface for #945 — that issue is specifically about FinOps hub deployment scripts, not CI workflow pins or the Optimization Engine. The version bumps here aren't harmful, but they're unvalidated (this repo's PR deploy checks only exercise template-changing PRs, not workflow-file-only changes, so there's no live signal that these six Az module bumps work end-to-end) and out of scope for what #945 actually asked for. Re-opening #945 against the real target: the toolkit's PowerShell module manifest (`.build/BuildHelper/Build-PsModule.ps1`, which currently pins `Az.Accounts 2.17.0` / `Az.Resources 6.5.1` as minimums) — the actual dependency surface for FinOps hub deployment via `New-FinOpsCostExport` and friends.

PR #2251, opened 2026-08-13
Replace serviceFamily sharding with two verified traversals per price type

RolandKrummenacher:
## 🐛 Problem

The weekly **Update Commitment Discount Eligibility** job exceeded its 60-minute timeout on [run 31612943595](https://github.com/microsoft/finops-toolkit/actions/runs/31612943595) and was killed 13 of 29 shards into the Consumption fetch, so it still published nothing. That run was the first to get past the `$orderby` HTTP 400 fixed in #2164 — the fetch itself works now; it is simply too slow to finish. The dataset has therefore not changed since it first shipped in v14.

## 🔧 Solution

Each price type is now walked twice by following the documented `NextPageLink` verbatim, and the two traversals are compared against each other to prove completeness. The `serviceFamily` sharding and the multi-pass union it required are gone.

### Why the sharding went

The sharding existed to work around a period when the Retail Prices API returned an unstable page order. That workaround is no longer earning its cost:

- The instability was last observed in early June 2026. MS support case 2606030050003725 closed without a root cause, and seven full traversals on 2026-06-29 were byte-identical.
- Run 31612943595's own telemetry shows every repeat pass across **all 22 completed shards adding exactly zero meters**, while costing 2-3x the traversal volume. No shard needed a 4th pass and none reported non-convergence, meaning pass 1 was already complete everywhere.
- Building the shard list required a full discovery traversal per price type, which alone consumed **25 of the job's 60 minutes** and produced only 38 family-name strings.

Per-serviceFamily counts are now tallied in the same pass as the fetch — every item already carries `serviceFamily`, so the per-family bookkeeping costs nothing extra.

### Completeness is verified per run, not inferred from history

The first revision of this PR leaned on the two historical guards alone, and [MSBrett was right](https://github.com/microsoft/finops-toolkit/pull/2251#discussion_r3775549468) that they cannot carry that weight: comparing one traversal against last week's counts with a 15% tolerance passes a uniform under-fetch, passes an offsetting drop-and-add, and cannot see a brand-new meter that was missed.

So **each price type is traversed twice and the two meterId sets are compared directly**, which needs no baseline and catches all three cases on the run that suffers them:

- **The published set is the union, not the intersection.** If only one pass saw a meter, the meter is real. Publishing the intersection would let a faulty pass silently *delete* rows — the failure being guarded against.
- **Per-family *sets* are unioned, not their counts.** Taking the max of two counts undercounts a family whose passes each missed a different meter (pass 1 sees `{a,b}`, pass 2 sees `{a,c}`: union 3, max 2).
- **Drift is a tolerance, not exact equality** — `-MaxVerifyDrift` defaults to 0.1%. The catalogue is live and the two passes span ~9 minutes for Reservation and ~32 for Consumption, so a meter can legitimately appear or retire between them. `-MaxVerifyDrift 0` gives strict equality.

The symmetric difference is logged on **every** run, not just failures: drift that climbs week over week while staying under the threshold is the early warning that the June 2026 fault is returning, and that is only visible if the passing value is printed too.

**The historical guards are retained** as a backstop rather than the primary signal, because they catch the one thing set comparison structurally cannot — a *deterministic* omission shared by both passes, which the comparison sees as perfect agreement.

Removes `Get-ServiceFamily`, `Assert-DiscoveryConverged`, `Add-BaselineShard`, and `Invoke-ShardedUnion`.

### Latent bugs fixed along the way

- **`Retry-After` was never honored.** It was read via `$Response.Headers['Retry-After']`, but `HttpResponseHeaders` has no string indexer, so that silently evaluated to `$null` (it does not throw) and every 429 fell through to the exponential backoff. Now reads the typed `.RetryAfter` property, handling both the delta-seconds and HTTP-date forms.
- **`-MaxSeconds` bounded only the header-driven wait**, not the exponential fallback, so attempt 5 waited `2^5 * 10 = 320s` against a ceiling documented as 300s ([review comment](https://github.com/microsoft/finops-toolkit/pull/2251#discussion_r3794614092)). Both paths are now clamped, so no single wait can exceed the ceiling; worst-case backoff for one page goes from 620s to 600s.
- **`$cachedShardCounts['Reservation']` threw `Cannot index into a null array`** when no baseline sidecar exists. #2164 committed the sidecar alongside the CSV, so this is not the path a normal CI run takes — it is reached by a fresh `-OutputPath` or a removed sidecar, which is how I hit it locally. The guard is still correct and worth keeping, but it was not blocking the scheduled run. Pre-existing on `dev` at four call sites.

### Determinism and workflow safety (added during review)

- **`shardcounts.json` → `familycounts.json`**, and `Get-ShardShortfall` → `Get-FamilyShortfall`. The workflow, the open-data CI path filter, and the `Package-Toolkit.ps1` exclusion were all updated to match, so nothing keys off the old name any more.
- **The sidecar is serialized through `ConvertTo-SortedMap`.** Hashtable key order is not guaranteed, and the validation run wrote the families in neither alphabetical nor insertion order. Since the workflow treats any diff in the sidecar as "data changed", that would have manufactured empty update PRs.
- **`workflow_dispatch` gained `ref` and `dry_run` inputs**, so a change to the fetch script can be exercised end-to-end from its PR branch before merging — previously the checkout was pinned to `ref: dev`, meaning a weekly job that takes ~44 minutes to fail could only be tested after merge. `dry_run` defaults to **true**, and a guard step that runs *before* checkout rejects a non-`dev` ref unless `dry_run` is set, so a data branch can never be built from a feature ref and carry its code commits into the data PR.
- `dry_run` is compared as a string in bash rather than in a GitHub `if:` expression: a `type: boolean` input arrives as the literal string `"false"`, which is **truthy** in expression syntax, so `if: ${{ !inputs.dry_run }}` would have inverted the check.

### Two changes outside the fetch logic

Flagging these explicitly, since they widen the PR beyond `Update-CommitmentDiscountEligibility.ps1`:

- **Changelog entry** in `docs-mslearn/toolkit/changelog.md`. The v14 dataset has been stale for every consumer since it shipped, which is user-facing enough to record.
- **Pester 6 is installed in `opendata-ci.yml`.** Not caused by this PR: `Build-OpenData -Test` calls `Test-PowerShell.ps1`, which requires Pester 6 (#2254) because the suite uses `-AllowNullOrEmptyForEach`, and the ubuntu runner image ships Pester 5.9.0 — so Open Data CI fails on any PR touching open data until the floor is installed. Same command `Init-Repo.ps1` runs, and the same floor `dev.yml` already pins through psmodulecache on Windows runners.

## 🧪 Validation

Two independent full runs against the live API produced **identical** results:

| | Items | Pages | Families | Eligible meters |
|---|---|---|---|---|
| Reservation | 146,671 | 147 | 9 | 66,199 |
| Consumption | 689,723 | 690 | 29 | 82,677 |

Both match sharded run 31612943595 **exactly, per service family as well as in aggregate** (Compute 58,723 / Databases 5,536 / Storage 1,673 / Analytics 132 / AI + ML 82 / Data 49 / Developer Tools 2 / Security 1 / Management and Governance 1). The simplification demonstrably loses nothing, and the matching runs are fresh evidence the API is stable.

Resulting CSV verified: 92,550 rows, sorted, lowercased, no duplicates. The per-family guard ran against #2164's committed baseline rather than skipping it, and passed on real data — Compute −4.3% and Databases −5.7%, both inside the 15% tolerance, with Storage and AI + ML unchanged.

End-to-end in CI, from this branch:

- [Run 31710007476](https://github.com/microsoft/finops-toolkit/actions/runs/31710007476) — the two-pass version, `dry_run=true`: **succeeded in 44m00s**, nothing pushed, no `opendata/*` branch left behind. The sharded run it replaces was killed at 60 minutes having completed 13 of 29 shards.
- [Run 31710035770](https://github.com/microsoft/finops-toolkit/actions/runs/31710035770) — the guard, dispatched with this branch and `dry_run` unticked: **failed in 25 seconds** before checkout, fetch, and push, all three reported `skipped`.

Also exercised all ten dispatch input combinations against the guard, including `dev-experiment` and `DEV` (both correctly rejected — no prefix or case-insensitive match) and the literal string `"false"`.

- Unit tests: 33 pass, including regression tests pinning the `Retry-After` indexer bug (asserts 30, not the fallback 20), the `MaxSeconds` clamp on the exponential path, sidecar key-order stability, and each of the three failure modes the verification pass exists to catch.
- PSScriptAnalyzer: clean. `Lint.Tests.ps1` and `MsLearnDocs.Tests.ps1`: 2982 pass.

## 📝 Notes for reviewers

- **Runtime is ~44 minutes.** An earlier comment on this PR reports 20m39s; that figure belongs to the single-traversal revision (`528a5f1f`) and is superseded. Adding the verification pass roughly doubled it, which is the deliberate trade for a completeness claim that holds. `timeout-minutes` is therefore raised from 60 to **90**; the headroom absorbs `Retry-After` backoff on a throttled run.
- No regenerated open data is included here. Validation wrote to a scratch copy; the data refresh belongs in the workflow's own PR.
- The script now declares `#Requires -Version 7.0` — it depends on `ConvertFrom-Json -AsHashtable`, `Export-Csv -UseQuotes`, and reading the error response as an `HttpResponseMessage`. CI is unaffected; the workflow already runs `shell: pwsh`.


src/scripts/Update-CommitmentDiscountEligibility.ps1
```diff
@@ -557,10 +461,10 @@ $rows | Export-Csv -Path $OutputPath -UseQuotes Always -NoTypeInformation -Encod
...
+# Persist this run's per-family counts as the baseline for the next run's guard.
 $newShardCounts = [ordered]@{
-    Reservation = $riResult.ShardCounts
-    Consumption = $spResult.ShardCounts
+    Reservation = $riResult.FamilyCounts
+    Consumption = $spResult.FamilyCounts
 }
 $newShardCounts | ConvertTo-Json -Depth 4 | Set-Content -Path $ShardCountPath -Encoding utf8 -NoNewline
-Write-Host "Wrote per-shard baseline to $ShardCountPath"
+Write-Host "Wrote per-family baseline to $ShardCountPath"
```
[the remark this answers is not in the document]
RolandKrummenacher: Good catch — fixed in 8bb06cd7.

Added a `ConvertTo-SortedMap` helper that sorts by key into an ordered dictionary before `ConvertTo-Json`. This was not theoretical: the run that validated this PR wrote the families as `Storage, Data, Databases, Compute, Developer Tools, Analytics, Security, AI + Machine Learning, Management and Governance` — neither alphabetical nor insertion order. Since the workflow treats any diff in the sidecar as "data changed", that would have produced empty update PRs.

Covered by a new unit test asserting that two maps differing only in insertion order serialize identically.

src/scripts/Update-CommitmentDiscountEligibility.ps1
```diff
@@ -62,19 +65,6 @@ $ErrorActionPreference = 'Stop'
 
 $apiBase = 'https://prices.azure.com/api/retail/prices?api-version=2023-01-01-preview'
```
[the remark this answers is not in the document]
RolandKrummenacher: Agreed — added `#Requires -Version 7.0` in 8bb06cd7, with a comment naming the specific PS7 dependencies (`ConvertFrom-Json -AsHashtable`, `Export-Csv -UseQuotes`, and reading the error response as an `HttpResponseMessage` rather than 5.1's `HttpWebResponse`).

Placed at the top of the file rather than at this line, so it fails fast before anything else runs. Verified that the directive does not break the comment-based help — `Get-Help` still returns the synopsis correctly. CI is unaffected either way since the workflow runs `shell: pwsh`.

RolandKrummenacher:
Added 528a5f1f, which touches the workflow rather than the script — flagging it explicitly since it widens this PR beyond the fetch logic.

Motivation: the checkout step was pinned to `ref: dev`, so a change to the fetch script could not be exercised end-to-end until after it merged. For a job that runs weekly, takes ~15 minutes to fail, and has not completed a successful run in months, that is a poor feedback loop. Two `workflow_dispatch` inputs address it:

- **`ref`** (default `dev`) — the branch to check out and run from. The scheduled run is unaffected: `github.event.inputs` is null for a schedule, so the expression falls back to `dev`.
- **`dry_run`** (default `false`) — runs the fetch and change detection but skips the branch push, so testing leaves no stray `opendata/*` branches. It writes a summary of what would have changed.

One subtlety worth noting for review: `dry_run` is compared as a string in bash rather than tested in a GitHub `if:` expression. A `type: boolean` input arrives as the literal string `"false"`, which is **truthy** in expression syntax, so `if: ${{ !inputs.dry_run }}` would have inverted the check and skipped the push whenever the box was left unticked.

I verified this by extracting the run block and exercising all four paths against a scratch repository — dry run with changes (correctly detects both the modified CSV and a newly created sidecar, pushes nothing), no changes, scheduled run with an empty `DRY_RUN`, and an explicit `DRY_RUN=false`. The last two both correctly take the push path.

Happy to split this into its own PR if you would rather keep this one to the script.

RolandKrummenacher:
**Correction to the PR description.** One claim in it is wrong, and I want it on the record rather than buried in a commit.

The description states that because the job has never completed a successful run, the baseline sidecar has never existed, and that the null-index bug *"would have failed the run even with the timeout raised."* That is incorrect. #2164 committed `src/open-data/CommitmentDiscountEligibility.shardcounts.json` alongside the CSV on 2026-08-12, so it is present on `dev` and on this branch, and `$cachedShardCounts` is not null in a normal CI run.

I hit the null path locally only because my validation harness copied just the CSV to a scratch `-OutputPath` and not the sidecar — an artifact of my own test setup, which I wrote up as a production finding. The guard is still correct and still worth keeping (a fresh `-OutputPath`, or a removed sidecar, reaches it), but it was not blocking the scheduled run. Comment corrected in bcaff154.

**The good news is that this makes the validation stronger than claimed.** Because the sidecar was present, today's dispatch actually exercised the per-family guard against #2164's committed baseline rather than skipping it. It passed, and the comparison is informative:

| Family | Baseline (#2164) | This run | Delta |
|---|---|---|---|
| Compute | 61,368 | 58,723 | −4.3% |
| Databases | 5,873 | 5,536 | −5.7% |
| Storage | 1,673 | 1,673 | 0% |
| AI + Machine Learning | 82 | 82 | 0% |

Both shrinking families are inside the 15% tolerance, so the guard passed on real data rather than trivially.

**End-to-end dispatch result** ([run 31674572978](https://github.com/microsoft/finops-toolkit/actions/runs/31674572978)), run from this branch with `dry_run=true`:

- **Succeeded in 20m39s**, against a 60-minute timeout. The sharded run it replaces was killed at 60 minutes having completed 13 of 29 shards.
- 146,671 Reservation items / 147 pages → 66,199 RI-eligible; 689,723 Consumption items / 690 pages → 82,677 SP-eligible — **identical to both local validation runs**.
- 92,550 meters written; 576 added, 4,359 changed, 308 removed.
- Zero rate-limiting events on the runner.
- Dry run took the correct path (`Dry run: changes detected, not pushing.`) and left no `opendata/*` branch behind.

Note that 20m39s exceeds the ~12-13 minute figure I projected from the old run's discovery timings, so that projection was optimistic. It is still comfortably inside the timeout, but worth knowing the real number is closer to a third of the budget than a fifth.

.github/workflows/opendata-commitment-eligibility.yml
```diff
@@ -2,8 +2,28 @@ name: Update Commitment Discount Eligibility
...
+        type: string
+      dry_run:
+        # Pair this with a non-dev ref when testing: the fetch and the change detection
+        # still run, but no opendata/* branch is pushed, so a test does not leave stray
+        # branches behind. A real run from a non-dev ref would also base its data branch
+        # on that ref, which is usually not what you want.
+        description: 'Fetch and detect changes only; do not push a branch'
+        required: false
+        default: false
+        type: boolean
```
MSBrett, line 30: Can we prevent a non-`dev` ref from taking the push path? `dry_run` defaults to `false`, so the natural branch-testing flow—setting only `ref`—checks out that feature branch, creates the `opendata/*` branch from it, and generates a compare-to-`dev` URL containing every feature commit as well as the data update. The comment says to pair these inputs, but the defaults make the unsafe combination the easiest one to invoke. Please either default `dry_run` to `true` or fail unless `RUN_REF == dev` or `DRY_RUN == true`.

src/scripts/Update-CommitmentDiscountEligibility.ps1
```diff
@@ -498,24 +422,31 @@ if ($cachedTotal -gt 0)
...
+# when no baseline sidecar exists, and PowerShell throws "Cannot index into a null
+# array" rather than yielding $null, so the section lookup MUST be guarded. The sidecar
+# is committed next to the CSV, so a normal CI run does have one; this path is taken
+# when running against a fresh -OutputPath (as a local test run does) or if the sidecar
+# is ever removed.
+$riBaseline = if ($cachedFamilyCounts) { $cachedFamilyCounts['Reservation'] } else { $null }
+$spBaseline = if ($cachedFamilyCounts) { $cachedFamilyCounts['Consumption'] } else { $null }
+
+$familyShortfall = @()
+$familyShortfall += Get-FamilyShortfall -Section 'Reservation' -Current $riResult.FamilyCounts -Baseline $riBaseline -MaxShrinkFraction $MaxShrinkFraction
```
MSBrett, line 609: This guard cannot provide the stated replacement for convergence. It only compares the single traversal with the previous run and permits a 15% drop per family. A paging regression can therefore omit up to that threshold, omit old meters while new meters in the same family offset the count, or omit brand-new meters entirely, and this check still passes. Step 4 then rewrites the CSV from the incomplete sets. Please retain an independent completeness signal (for example, a second traversal whose meter-ID set must match) rather than relying only on historical counts; otherwise the claim that paging instability will “abort loudly instead of publishing partial data” is not true.

MSBrett (commented):
Review findings from the single-traversal change.

.github/workflows/opendata-commitment-eligibility.yml
RolandKrummenacher: Agreed, and fixed in 055cc2cf. You're right that the defaults made the unsafe combination the easiest one to invoke — a comment telling people to pair the inputs isn't a control.

I applied **both** of your suggestions rather than picking one, because they cover different mistakes:

- **`dry_run` now defaults to `true`**, so a real publish is a deliberate choice rather than the path of least resistance.
- **A guard step rejects `ref != dev` unless `dry_run` is true**, which also covers someone who unticks `dry_run` out of habit and then dispatches from a branch.

The guard runs **before** the checkout step, so an invalid combination costs seconds instead of the ~35 minutes the fetch now takes.

The scheduled run is unaffected. A schedule has no `inputs` object at all, so `DRY_RUN` is empty — not the new `"true"` default — and `RUN_REF` falls back to `dev`, which takes the normal push path.

Verified in CI rather than only in a local harness. Dispatching the exact combination you described (this branch, `dry_run` unticked) is [run 31710035770](https://github.com/microsoft/finops-toolkit/actions/runs/31710035770), which failed in 25 seconds:

```
##[error]ref='RolandKrummenacher/simplify-eligibility-traversal' is not 'dev' and dry_run is not true.
A push from a non-dev ref would base the opendata/* branch on that ref, so the data PR would carry
its code commits as well as the data update. Re-run with dry_run enabled to test a branch, or with
ref='dev' to publish data.
```

Checkout, fetch, and push all reported `skipped`, so nothing was fetched and no branch was created.

I also exercised all ten input combinations against the guard as written, including the edge cases the string comparison exists for: `dev-experiment` and `DEV` are both correctly rejected (no prefix or case-insensitive match), and the literal string `"false"` — which is *truthy* in a GitHub `if:` expression, the reason this is compared in bash — does not bypass it.

src/scripts/Update-CommitmentDiscountEligibility.ps1
RolandKrummenacher: You're right, and the docstring's claim was wrong as written. Fixed in 99f409ec.

All three of your failure modes hold against the per-family guard: it compares one traversal against the *previous run's* counts with a 15% tolerance, so a uniform under-fetch below the threshold passes, an offsetting drop-and-add passes because counts are not sets, and a missed brand-new meter has no baseline to be missing from. Step 4 then rewrites the CSV from the incomplete sets. Presenting that as the replacement for convergence overstated it.

**Each price type is now traversed twice and the two meterId sets are compared directly.** That needs no historical baseline, so it catches all three cases on the run that suffers them rather than inferring a problem from how totals moved since last week.

Three design points worth flagging, since they differ slightly from "must match":

- **The published set is the union, not the intersection.** If exactly one pass missed a meter, the meter is real and belongs in the output. Publishing the intersection would let a faulty pass silently *delete* rows — the very failure being guarded against.
- **Per-family *sets* are unioned, not the counts.** Taking the max of two counts undercounts a family whose passes each missed a different meter (pass 1 sees `{a,b}`, pass 2 sees `{a,c}`: union 3, max 2), so `Get-EligibleMeter` now returns `FamilyKeys` and the counts are derived from it.
- **Drift is a tolerance, not exact equality**, defaulting to 0.1%. The catalogue is live, and the two passes span ~9 minutes for Reservation and ~32 for Consumption, so a meter can legitimately appear or retire between them. Requiring byte equality would make a weekly job flaky for a reason that isn't a fault. `-MaxVerifyDrift 0` gives you strict equality if you'd rather have it.

The symmetric difference is now logged on **every** run, not just failures — a drift that climbs week over week while staying under the threshold is the early warning that the June-2026 fault is returning, and that's only visible if the passing value is printed too.

**I kept the historical guards**, re-documented as a backstop rather than the completeness signal, because they catch the one thing set comparison structurally cannot: a *deterministic* omission shared by both passes, which the comparison sees as perfect agreement.

### Validation

End-to-end on this branch, [run 31710007476](https://github.com/microsoft/finops-toolkit/actions/runs/31710007476) (dry run, nothing pushed):

```
Reservation  Pass 1: 146671 items over 147 pages, 66199 eligible meters
Reservation  Pass 2: 146671 items over 147 pages, 66199 eligible meters
Verification: 66199 meters in union, 0 only in pass 1, 0 only in pass 2 (drift 0%)

Consumption  Pass 1: 689723 items over 690 pages, 82677 eligible meters
Consumption  Pass 2: 689723 items over 690 pages, 82677 eligible meters
Verification: 82677 meters in union, 0 only in pass 1, 0 only in pass 2 (drift 0%)
```

Exact set agreement on both, and identical to the totals from the earlier single-pass run 8.5 hours before — so four independent full traversals now agree, which is further evidence the paging instability really is dormant rather than merely unobserved.

Unit tests go 22 → 32, driving the verification through each mode a count comparison can't see: the offsetting drop-and-add that leaves the total unchanged, a meter missed by either pass, the exact-tolerance boundary, per-family set unioning, and the empty-result divide-by-zero path.

### Cost

44 minutes measured, timeout raised 60 → 90, so 46 minutes of headroom. Reservation is verified first deliberately — at ~3-6 minutes per pass against Consumption's ~15-17, an unstable API aborts after ~12 minutes instead of ~44.

That is still well inside what the removed workaround cost: it needed a full discovery traversal per price type (25 of 60 minutes on its own) before repeat-until-stable across 22 shards, and it timed out. So this isn't re-adding what the PR set out to remove — the difference is between an unbounded repeat loop on top of discovery, and exactly two bounded passes. Fair to say the PR previously bought its runtime back by giving up the integrity guarantee; this puts the guarantee back at a cost that fits.

.github/workflows/opendata-commitment-eligibility.yml
```diff
@@ -2,20 +2,62 @@ name: Update Commitment Discount Eligibility
...
 
 jobs:
   update:
     runs-on: ubuntu-latest
-    timeout-minutes: 60
+    # Two full traversals per price type (fetch + verification) measured 44 minutes on
+    # run 31710007476. The headroom absorbs Retry-After backoff on a throttled run.
+    timeout-minutes: 90
     steps:
+      # Fail before the ~44-minute fetch rather than after it, so an invalid input
```
flanakin, line 42: 🤖 [AI][Claude] ⚠️ **Should fix**

This comment says the 90-minute timeout is based on a run that took 44 minutes, but your own PR comment reports the actual end-to-end run against this code finished in 20m39s. That's a big enough gap that it's worth double-checking which commit run 31710007476 actually corresponds to — if it predates the final two-pass approach, updating the comment to cite the 20-minute figure (or re-measuring) would keep this accurate for whoever tunes the timeout next.

flanakin (commented):
🤖 [AI][Claude] **PR Review**

**Summary:** This replaces the old sharded discovery approach for the weekly commitment-discount-eligibility fetch with a two-pass verified traversal, and fixes a real `Retry-After` parsing bug along the way. The core logic is correct, well tested (15/15 unit tests pass locally, no lint regressions), and the `workflow_dispatch` safety guard behaves correctly across all input combinations I simulated. No blockers — just a few things worth tightening before merge.

### ⚠️ Should fix (3)

1. **PR description is out of date.** It says renaming `shardcounts.json` is "a separate change" and projects "~12-13 minutes" runtime, but the final commits did rename the file to `familycounts.json` and the actual measured runtime is ~20 minutes (per your own follow-up comment on this PR). Since GitHub carries this description into the merge commit, worth a quick edit so it matches what actually shipped. The title ("a single traversal") is also a bit of an undersell — the final approach does two passes per price type with drift verification, which is a stronger reliability story than "single traversal" suggests.
2. **No changelog entry.** The commitment-discount-eligibility feature shipped in v14, and this fixes a bug where the weekly refresh job has never completed successfully since — so open-data consumers have been getting stale data. That's user-facing enough to warrant a changelog entry (e.g., a "Fixed" line in `docs-mslearn/toolkit/changelog.md`).
3. See inline comment on the timeout comment mismatch (44 min vs. the 20-minute run you reported).

### 💡 Suggestions (2)

See inline comments — a small clarifying comment for a PowerShell idiom, and a concrete gap in the retry backoff clamp.

src/scripts/Update-CommitmentDiscountEligibility.ps1
```diff
@@ -161,228 +204,282 @@ function Get-RetailPriceSegment
...
+        UNDERCOUNT a family whose passes each missed a different meter (pass 1 sees {a,b},
+        pass 2 sees {a,c}: the union is 3, the max is 2). FamilyCounts is derived from the
+        sets for callers that only need the totals.
+
+        FamilyCounts counts DISTINCT COLLECTED meterIds per family, matching what the
+        sharded implementation persisted, so an existing baseline stays comparable. A
+        family bucket is created for every family seen even when it collects nothing, so
+        a family that legitimately has no eligible meters is recorded as 0 rather than
+        vanishing from the baseline.
     #>
```
flanakin, line 285: 🤖 [AI][Claude] 💡 **Suggestion**

The `-MaxSeconds` clamp two lines above only applies to the `Retry-After`-driven wait, not this exponential fallback. With `$maxRetries = 5`, retry 5 alone waits `2^5*10 = 320s`, which already exceeds the 300s cap used on the `Retry-After` path just above. If the intent is for `MaxSeconds` to bound every wait (not just the header-driven one), this line should be wrapped the same way:

```suggestion
    return [Math]::Min([int][Math]::Pow(2, $Attempt) * 10, $MaxSeconds)
```

src/scripts/Update-CommitmentDiscountEligibility.ps1
```diff
@@ -393,19 +490,19 @@ function Get-ShardShortfall
 
     $violations = @()
```
flanakin, line 499: 🤖 [AI][Claude] 💡 **Suggestion**

This merges two hashtables with `+` (verified locally — PowerShell unions the keys correctly). It works, but it's a bit less obvious than the explicit `@{ Keys = ...; FamilyKeys = ... }` construction used in the main return path a few lines down. A short comment noting that `+` merges the two hashtables' keys would help a future reader who isn't sure PowerShell supports that.

.github/workflows/opendata-commitment-eligibility.yml
RolandKrummenacher: The 44-minute figure is the accurate one — the 20m39s number in my earlier comment is what's stale, and I should have flagged that when the approach changed.

The two runs are different code:

| Run | SHA | Approach | Duration |
|---|---|---|---|
| [31674572978](https://github.com/microsoft/finops-toolkit/actions/runs/31674572978) | `528a5f1f` | single traversal per price type | 20m39s |
| [31710007476](https://github.com/microsoft/finops-toolkit/actions/runs/31710007476) | `9e38f182` | two traversals + drift verification | 44m00s (14:24:23 → 15:08:23) |

The verification pass you asked for in [your other comment](https://github.com/microsoft/finops-toolkit/pull/2251#discussion_r3775549468) roughly doubled the runtime, which is the trade I took deliberately — two independent traversals are what make the completeness claim true. So the comment and the 90-minute timeout are describing the code as it now stands.

I've left both as-is and corrected the PR description instead, so the 20-minute number doesn't outlive this thread.

src/scripts/Update-CommitmentDiscountEligibility.ps1
RolandKrummenacher: Agreed — fixed in ccffc75a. `MaxSeconds` was documented as the ceiling on the wait, but only the header path honored it.

Attempt 5 is reachable (`$maxRetries = 5`, and `$retries` is incremented before `Get-RetryDelay` is called), so the last retry waited 320s against a 300s cap. Worst-case total backoff for one page goes from 620s to 600s — small, but the inconsistency was in the part the docstring promised.

Added a regression test covering attempt 5 with and without a response object, plus attempt 4 to confirm the sub-ceiling backoff is unchanged. 33/33 unit tests pass (the new one included).

src/scripts/Update-CommitmentDiscountEligibility.ps1
RolandKrummenacher: Added in ccffc75a.

I also noted the behavior that makes `+` the right choice rather than merely a working one: it *throws* on a duplicate key instead of silently overwriting, so if `Get-EligibleMeter` ever starts returning `VerifyDrift`/`OnlyFirst`/`OnlySecond` itself, this line fails loudly rather than quietly picking a winner. Verified locally — `@{a=1} + @{a=9}` raises `Item has already been added`.

MSBrett (commented):
**Follow-up review — verifying the fixes from the earlier review rounds.**

I re-traced the full diff against the current HEAD, re-ran the test suite, and read every prior comment/reply/thread (Copilot's initial review, my own findings on the single-traversal guard, and the AI/Claude review on 2026-08-17), including thread resolution state.

**All previously raised issues check out as fixed in code, not just marked resolved:**
- `ConvertTo-SortedMap` gives deterministic `familycounts.json` key order (values reconcile exactly against the old `shardcounts.json`, just resorted).
- `#Requires -Version 7.0` is in place with a clear rationale comment.
- The `workflow_dispatch` `ref`/`dry_run` guard correctly rejects a non-`dev` ref unless `dry_run` is set — verified the bash string comparisons against all four paths (scheduled/empty, explicit `false`, `dev`+push, non-`dev`+dry-run).
- `Get-RetryDelay` reads the typed `HttpResponseHeaders.RetryAfter` (the old `Headers['Retry-After']` indexer silently returned `$null` — genuine prior bug, now fixed and regression-tested) and clamps both the header path and the exponential-backoff path to `-MaxSeconds`.
- `Get-VerifiedEligibleMeter`'s union-of-two-passes and symmetric-drift math check out by hand (recovery of a single-pass miss, offsetting drop/add detection, drift-at-boundary using `-gt`).
- Changelog entry added (`docs-mslearn/toolkit/changelog.md`), Pester 6 install added to `opendata-ci.yml` to unblock CI.
- No stray references to the old shard vocabulary (`Get-ServiceFamily`, `Add-BaselineShard`, `Invoke-ShardedUnion`, `Assert-DiscoveryConverged`, `Get-ShardShortfall`, `shardcounts.json`) anywhere in the repo.

**Validation performed:**
- `Invoke-Pester` on `Update-CommitmentDiscountEligibility.Tests.ps1`: **33/33 passed**.
- PSScriptAnalyzer: only pre-existing `PSAvoidUsingWriteHost` warnings (this script isn't in the enforced lint path); no new rule violations.
- Both workflow YAMLs parse cleanly; current PR checks are all green.

**One minor loose end, not a blocker:** the PR title ("...a single traversal per price type") still undersells the final design, which does two full traversals per price type (fetch + verification) — this was flagged in the 2026-08-17 review as a "should fix" alongside the stale body/changelog, and the body + changelog were fixed but the title wasn't. Leaving as-is is fine; flagging for the record.

No new issues found. LGTM.

MSBrett (approved):

PR #2259, opened 2026-08-13
fix(hubs): use an Azure Files private endpoint for deployment script storage in private mode

### LINKED ISSUE #2258: [Hubs] Private networking deployment fails — defaultOutboundAccess breaks deployment scripts (regression)
## 🐛 Problem

Deploying a FinOps hub with `enablePublicAccess = false` fails on the current `dev` branch. The deployment aborts at the first deployment script, before the app layer starts:

```
DeploymentScriptStorageAccountWithServiceEndpointEnabled

Storage account '<script storage>' has firewall settings enabled which are not supported for
deployment scripts. If providing subnets with managed identity, make sure the "Allow Azure services
on the trusted services list to access this storage account" is enabled and proper RBAC is set on
the given storage.
```

This is a **regression in the current release cycle**. The shipped v14 template (`docs/deploy/finops-hub-latest.json`) deploys the same scenario successfully.

Both private network modes are affected, since `privateRouting` is simply `!enablePublicAccess`.

## ⚠️ The error message points at the wrong thing

The script storage account named in the error is configured **identically** in the working v14 deployment and the failing one:

```
v14  (deploys)   publicNetworkAccess=Enabled  defaultAction=Deny  bypass=AzureServices  vnetRules=[script-subnet]
dev  (fails)     publicNetworkAccess=Enabled  defaultAction=Deny  bypass=AzureServices  vnetRules=[script-subnet]
```

The storage firewall is not the cause. Chasing it leads nowhere.

## 🔍 Actual cause

`Core/infrastructure.bicep` now sets `defaultOutboundAccess` explicitly on all three subnets:

```bicep
defaultOutboundAccess: !hub.options.natGateway
```

Setting this property — **at any value, on any subnet in the virtual network** — makes the Deployment Scripts service reject the script storage account. The NAT Gateway itself is not the problem.

## 🧪 Isolation

Nine deployments on a dedicated test subscription, storage-only, same region:

| # | Template / mode | `defaultOutboundAccess` set | NAT Gateway | Result |
| 1 | dev, private | yes (all subnets) | yes | ❌ |
| 2 | dev, vnet | yes (all subnets) | no | ❌ |
| 3 | **v14 shipped** | no | no | ✅ |
| 4 | dev, same params as #3 | yes (all subnets) | no | ❌ |
| 5 | dev, property removed | no | no | ✅ |
| 6 | dev, property removed on script subnet only | yes (2 subnets) | yes | ❌ |
| 7 | dev, script subnet untouched entirely | yes (2 subnets) | yes | ❌ |
| 8 | repeat of #5 | no | no | ✅ |
| 9 | dev, property removed, **NAT enabled** | no | **yes** | ✅ |

Failure occurs exactly when the property is set, without exception.

- **#3 vs #4** differ in nothing but this property (identical parameters, identical storage config, identical subnet prefixes, service endpoints and delegations).
- **#5 and #8** show the successes are reproducible, not luck.
- **#6 and #7** rule out the theory that only the script subnet matters — the property on *other* subnets is equally sufficient to break it.
- **#9** was run as a falsifiable prediction: NAT Gateway attached to script and Data Explorer subnets, property nowhere. It succeeded, which clears the NAT Gateway and confirms the property as the sole cause.

## 🤔 Expected

Deploying with public network access disabled completes successfully, with or without the NAT Gateway.

## 🔧 Environment

- **FinOps hub version:** current `dev` (15.0-dev). Last known good: v14 shipped template.
- **Deployment:** ARM template deployment, storage-only (`dataExplorerName: ''`), `enablePublicAccess: false`
- **Region:** westus

## 💡 Fix

Drop `defaultOutboundAccess` from the three subnet definitions. With the NAT Gateway attached, outbound traffic routes through the gateway regardless, so the explicit setting is redundant — and it is the only thing standing between the private modes and a working deployment.

Verified: rebuilt from source with the property removed and deployed with `enableNatGateway: true` and `enablePublicAccess: false` — deployment succeeds and all deployment scripts run.

PR follows.

RolandKrummenacher:
Fixes #2258

## 🐛 Problem

Deploying with `enablePublicAccess = false` fails on `dev` at the first deployment script:

```
DeploymentScriptStorageAccountWithServiceEndpointEnabled
Storage account '<script storage>' has firewall settings enabled which are not supported for
deployment scripts.
```

The shipped v14 template deploys the same scenario successfully, so this is a regression in the current cycle. Both private network modes are affected, since `privateRouting` is `!enablePublicAccess`.

**The error message is misleading.** The script storage account it names is configured identically in the working v14 deployment:

```
v14  (deploys)   publicNetworkAccess=Enabled  defaultAction=Deny  bypass=AzureServices  vnetRules=[script-subnet]
dev  (fails)     publicNetworkAccess=Enabled  defaultAction=Deny  bypass=AzureServices  vnetRules=[script-subnet]
```

## 🔍 Cause

`defaultOutboundAccess` is now set explicitly on all three subnets. Setting it — at any value, on any subnet in the virtual network — makes the Deployment Scripts service reject the script storage account.

## 🔧 Solution

Drop the property from the three subnet definitions. Attaching the NAT Gateway is what routes outbound traffic, so the explicit setting is redundant and the feature is unaffected.

A comment above the NAT Gateway block records why the property must not come back, including the fact that the Azure error points at the storage account rather than the subnet.

## 🧪 Validation

Nine deployments on a test subscription (storage-only, same region):

| # | Template / mode | property set | NAT Gateway | Result |
| 1 | dev, private | yes (all) | yes | ❌ |
| 2 | dev, vnet | yes (all) | no | ❌ |
| 3 | **v14 shipped** | no | no | ✅ |
| 4 | dev, same params as #3 | yes (all) | no | ❌ |
| 5 | dev, property removed | no | no | ✅ |
| 6 | dev, removed on script subnet only | yes (2) | yes | ❌ |
| 7 | dev, script subnet untouched | yes (2) | yes | ❌ |
| 8 | repeat of #5 | no | no | ✅ |
| 9 | dev, property removed, NAT on | no | yes | ✅ |

Failure occurs exactly when the property is set, without exception. #3 vs #4 differ in nothing else. #5/#8 show the successes reproduce. #6/#7 rule out a script-subnet-only explanation. #9 was run as a falsifiable prediction and cleared the NAT Gateway.

**This branch**: rebuilt from source and deployed with `enableNatGateway: true` and `enablePublicAccess: false` — succeeded in 18 minutes, all 7 deployment scripts ran, NAT Gateway attached to the script and Data Explorer subnets, property absent everywhere.

Single-file template change; no PowerShell, workbook or query code is touched.


flanakin (commented):
🤖 [AI][Claude] **PR Review**

**Summary:** This is a well-diagnosed, well-validated fix — removing `defaultOutboundAccess` from the three subnet definitions resolves a real deployment regression where the Deployment Scripts service rejects the script storage account whenever that property is set on any subnet in the vNet. The Bicep change itself is correct, minimal, and builds cleanly, and the comment explaining *why* the property must not come back is a nice touch. No blockers.

The only gap: this PR only touches `infrastructure.bicep`, but three other files still describe the old `defaultOutboundAccess` behavior that no longer exists, so they're now inaccurate. None of these lines are part of this PR's diff, so I can't attach inline comments — listing them here instead.

### ⚠️ Should fix (3) — stale references to the removed property

1. **`src/powershell/Public/Deploy-FinOpsHub.ps1`** (`NetworkMode` parameter help text) — currently says private mode gives you "private endpoints + NAT Gateway, subnets locked down with defaultOutboundAccess=false." That property is gone now. Suggest something like: "private endpoints + NAT Gateway for controlled outbound access — required when the 'Subnets should be private' policy is enforced."
2. **`src/templates/finops-hub/modules/fx/hub-types.bicep`** (`natGateway` option description) — says "When enabled, subnets disable Azure default outbound access and route through the NAT Gateway." The first half is no longer true; only the NAT Gateway attachment happens now. Suggest: "When enabled, the script and Data Explorer subnets route outbound traffic through the NAT Gateway."
3. **`src/templates/finops-hub/createUiDefinition.json`** (`enableNatGateway` checkbox label) — this is customer-facing Azure Portal text and currently reads "Deploy NAT Gateway and disable default outbound access," which no longer matches what the checkbox does. Suggest: "Deploy NAT Gateway for controlled outbound access."

### 💡 Suggestions (2)

1. **`src/scripts/Deploy-Hub.ps1`** — private mode help text says "subnets locked down," which still implies the removed behavior even though it doesn't name the property directly. Worth tightening for consistency with whatever wording is chosen for `Deploy-FinOpsHub.ps1`, though lower priority since it's not explicitly incorrect.
2. **Changelog** — no entry added. Defensible: the feature that introduced the bug (NAT Gateway for private subnets, #2163) is only in the commented-out "Unreleased" section of the changelog, not a released version, so this is arguably fixing pre-release work rather than a shipped regression. Worth a quick check with maintainers on whether the #2163 changelog entry (once uncommented for release) should just reflect the corrected behavior directly, with no separate "Fixed" line needed.

RolandKrummenacher:
Thanks — all three "should fix" items and the `Deploy-Hub.ps1` suggestion are addressed in acad6d38, along with **two more instances the review missed**.

### Fixed (6 strings)

| File | Was | Now |
|---|---|---|
| `src/powershell/Public/Deploy-FinOpsHub.ps1` | "subnets locked down with defaultOutboundAccess=false" | "NAT Gateway for controlled outbound access" |
| `src/scripts/Deploy-Hub.ps1` | "subnets locked down" | same wording as above, for consistency |
| `src/templates/finops-hub/createUiDefinition.json` | "Deploy NAT Gateway and disable default outbound access" | "Deploy NAT Gateway for controlled outbound access" |
| `src/templates/finops-hub/modules/fx/hub-types.bicep` | "subnets disable Azure default outbound access and route through the NAT Gateway" | "the script and Data Explorer subnets route outbound traffic through the NAT Gateway" |
| **`src/templates/finops-hub/main.bicep`** | same stale sentence | same correction |
| **`src/templates/finops-hub/modules/hub.bicep`** | same stale sentence | same correction |

The last two are the `enableNatGateway` `@description()` strings. They matter more than the others, because they render into the generated ARM template and the published parameter reference — the incorrect text would have reached template consumers, not just contributors reading source.

I used "the script and Data Explorer subnets" rather than a bare "subnets" because only those two are attached to the NAT Gateway; the private endpoint subnet is not.

`Deploy-Hub.ps1` line 92 was left alone deliberately: it documents `-Private` (equivalent to `vnet` mode), and "default outbound access" is still accurate there — that mode genuinely relies on Azure default outbound.

### On the changelog

I've left it out. The #2163 entry is still inside the commented-out Unreleased block, and its text is "Added VNet and private network modes, including opt-in NAT Gateway support for private mode" — it never described the `defaultOutboundAccess` behavior, so it stays accurate as written and needs no correction when it's uncommented for release. Happy to add a "Fixed" line if you'd rather have the bug recorded explicitly.

### Validation

- `bicep build` clean on both `main.bicep` and `modules/hub.bicep` (only the two pre-existing `no-unused-params` warnings in `Recommendations/app.bicep`)
- `createUiDefinition.json` parses; both `.ps1` files parse
- `Lint.Tests.ps1` — 2400/2400 pass

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Core/infrastructure.bicep
```diff
@@ -211,9 +208,15 @@ resource vNet 'Microsoft.Network/virtualNetworks@2023-11-01' = if (hub.options.p
...
 
 //------------------------------------------------------------------------------
-// NAT Gateway (provides outbound for script-subnet + dataExplorer-subnet when
-// defaultOutboundAccess is disabled; required by the 'Subnets should be private'
-// policy and the September 2025 implicit-outbound retirement)
+// NAT Gateway (provides explicit outbound for script-subnet + dataExplorer-subnet;
+// required by the 'Subnets should be private' policy and the September 2025
+// implicit-outbound retirement)
+//
+// Do not set defaultOutboundAccess on the subnets. Setting it -- at any value, on any
```
MSBrett: This blocks private deployments under the built-in **Subnets should be private** policy (`7bca8353-aa3b-429b-904a-9229c4385837`). The policy requires `properties.defaultOutboundAccess` to equal `false`; omitting the property leaves it unset, so policy denies the deployment. NAT Gateway attachment provides egress but does not satisfy this policy. Keep `defaultOutboundAccess: !hub.options.natGateway` on every subnet.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Core/infrastructure.bicep
```diff
@@ -211,9 +208,15 @@ resource vNet 'Microsoft.Network/virtualNetworks@2023-11-01' = if (hub.options.p
...
+// subnet in this virtual network -- makes the Deployment Scripts service reject the
+// script storage account with DeploymentScriptStorageAccountWithServiceEndpointEnabled,
```
MSBrett: The Storage service endpoint is the incompatible part, not `defaultOutboundAccess`. Microsoft's supported private-network design for Deployment Scripts uses a Storage `file` private endpoint, `privatelink.file.core.windows.net`, and disabled public access. Replace the service endpoint and subnet firewall rule with that design, then keep `defaultOutboundAccess: false` for private mode.

MSBrett (changes requested):
The current change removes a policy-required subnet setting. NAT Gateway attachment and Azure Policy compliance are separate requirements. The deployment-script storage path must change so both requirements can be met.

MSBrett:
## Findings and test evidence

NAT Gateway attachment is necessary for outbound traffic, but it does not meet the Azure Policy requirement by itself. Policy `7bca8353-aa3b-429b-904a-9229c4385837` requires each private mode subnet to set `defaultOutboundAccess` to `false`. The original PR removed that setting from all three subnets, so policy could deny the deployment.

The Deployment Scripts conflict comes from the Storage service-endpoint design. The fix keeps `defaultOutboundAccess: false` in private mode, keeps NAT on the script and Data Explorer subnets, and replaces the Storage service endpoint with the documented Azure Files private endpoint and `privatelink.file.core.windows.net` private DNS zone.

### Requirements traceability matrix

| Requirement | Evidence | Result |
|---|---|---|
| Private mode subnets set `defaultOutboundAccess: false`; VNet mode retains default outbound access | Compiled-ARM regression test checks all three conditional subnet expressions | Pass |
| Script and Data Explorer subnets use NAT | Compiled-ARM regression test checks two NAT associations | Pass |
| Deployment-script storage does not use a service endpoint | Regression test checks that the storage subnet firewall rule is absent | Pass |
| Deployment-script storage uses Azure Files privately | Regression test checks the `file` private endpoint, disabled public access, and `privatelink.file.core.windows.net` | Pass |
| Public Bicep entry points compile | `src/templates/finops-hub/main.bicep` and `src/templates/finops-hub/modules/hub.bicep` compiled locally | Pass |
| Focused networking regression test with Bicep installed | `HubsPrivateNetworking.Tests.ps1`: 1 passed, 0 failed | Pass |
| Focused networking regression test without Bicep | Existing repository skip convention exercised: 1 skipped, 0 failed | Pass |
| Review wording matches conditional behavior | Source comment and changelog now say **private mode subnets** | Pass |
| Runtime deployment completes in both network modes | Maintainer live validation from the branch head in a disposable westus subscription: both `private` and `vnet` deployments succeeded and all 7 deployment scripts executed ([evidence](https://github.com/microsoft/finops-toolkit/pull/2259#issuecomment-5352236080)) | Pass |

The fix and review remediation are pushed through commit [`264e5463`](https://github.com/microsoft/finops-toolkit/commit/264e5463d08eb30a2707c6d7ca6c5bb42e614cf0).

MSBrett (approved):

MSBrett:
Addressed the three suppressed suggestions from review 4973971190 in commit [$(264e5463d08eb30a2707c6d7ca6c5bb42e614cf0.Substring(0,8))](https://github.com/microsoft/finops-toolkit/commit/264e5463d08eb30a2707c6d7ca6c5bb42e614cf0):

| Requirement | Change | Result |
|---|---|---|
| Match the conditional subnet behavior | The source comment now says **private mode subnets** set defaultOutboundAccess to alse | Addressed |
| Keep the changelog accurate | The changelog now uses the same **private mode subnets** wording | Addressed |
| Do not fail unit tests when Bicep is unavailable | The focused test follows the existing Get-Command bicep and Set-ItResult -Skipped pattern | Addressed |

Validation: both public Bicep entry points compiled; the focused test passed with Bicep installed and skipped cleanly with Bicep absent.

RolandKrummenacher:
## Live validation of this branch

Deployed the current branch head end to end on a disposable test subscription, in both modes affected by the regression. This closes the one row marked *Incomplete* in the traceability matrix above.

| Check | `private` (NAT on) | `vnet` (NAT off) |
| --- | --- | --- |
| Deployment result | **Succeeded** | **Succeeded** |
| Deployment scripts executed | **7 / 7** | **7 / 7** |
| `defaultOutboundAccess` on all three subnets | `false` | `true` (`!natGateway`) |
| Service endpoint on script subnet | removed | removed |
| Script storage public access | `Disabled` | `Disabled` |
| Script storage virtual network rules | 0 | 0 |
| `privatelink.file.core.windows.net` zone + vNet link | present | present |

Storage-only (`dataExplorerName: ''`), westus, template built from source on this branch.

**The 7/7 is the part that matters.** In all five failing runs from the original investigation, the deployment aborted at the *first* deployment script, before any container group started — so "Succeeded" alone would not have been sufficient evidence. Here the container groups actually ran, and the `private` mode ends up with `defaultOutboundAccess: false` on all three subnets in the deployed resources, not just in the compiled ARM.

For completeness: the compiled-ARM regression tests in `HubsPrivateNetworking.Tests.ps1` verify the template *says* the right thing; the original failure only surfaced at execution time, on a template that was already valid. The two checks complement each other.

### One observation, not a blocker

In `vnet` mode the subnets end up with `defaultOutboundAccess: true`, since the expression is `!natGateway`. Policy `7bca8353-aa3b-429b-904a-9229c4385837` requires `false`, so `vnet` mode is not policy-compliant — only `private` mode is. That matches how the two modes are documented, but it does mean "VNet mode" alone will not satisfy the *Subnets should be private* policy. Worth a separate look if that was not the intent; nothing to change here.

### On the original approach in this PR

Removing `defaultOutboundAccess` was the wrong half of the fix — the policy flags the property being *absent* as well as being `true`, so it would have traded the deployment failure for a policy violation. The service-endpoint change addresses the same conflict from the side that is actually optional. In the nine runs behind #2258, the service endpoint on the script storage was the one variable never altered, which is why it did not surface as a candidate.

MSBrett:
Confirmed. This is the intended mode boundary, so no code change is needed:

- `vnet`: private endpoints with Azure default outbound access; it does **not** satisfy policy `7bca8353-aa3b-429b-904a-9229c4385837`.
- `private`: private endpoints plus NAT Gateway and `defaultOutboundAccess: false`; this is the policy-compliant mode.

The command help and the validation table already distinguish these modes. I updated the canonical RTM to credit this live validation and close the runtime-test gap. Thank you for testing both modes and confirming that all 7 deployment scripts ran.

PR #2261, opened 2026-08-17
fix(powershell): treat Start-FinOpsCostExport dates as UTC calendar dates

### LINKED ISSUE #2255: [PowerShell] Start-FinOpsCostExport shifts the export period by one day in positive UTC offsets
## 🐛 Problem

`Start-FinOpsCostExport` converts its date parameters to UTC before truncating them to a day. For any positive UTC offset that moves the date back by one day, so the exported period is shifted:

```powershell
# machine in UTC+01:00
Start-FinOpsCostExport -Name 'CostExport' -Scope $scope -StartDate '2026-01-01'
```

| | value |
| --- | --- |
| requested | 2026-01-01 |
| actually exported | **2025-12-31** through **2026-01-30** |

The user asks for January and gets 31 December added while 31 January is missing.

Reproducible without Azure:

```powershell
$in = [datetime]'2026-01-01'      # how PowerShell parses the parameter: local midnight
$in.ToUniversalTime().Date        # -> 2025-12-31 in UTC+1/+2
```

## 🔍 Where

`src/powershell/Public/Start-FinOpsCostExport.ps1` applies the same pattern in 7 places, e.g.:

```powershell
# L102 -- backfill start
$StartDate = (Get-Date -Day 1 -Hour 0 -Minute 0 -Second 0 -Millisecond 0).ToUniversalTime().Date

# L119-L126 -- explicit start/end
$StartDate = $StartDate.ToUniversalTime().Date
$EndDate   = $EndDate.ToUniversalTime().Date
$EndDate   = $StartDate.ToUniversalTime().Date.AddMonths(1).AddDays(-1)
```

`-Backfill` is affected the same way: the first of the current month becomes the last day of the previous month, so each backfilled window is off by a day.

No other command in `src/powershell/Public/` uses this pattern.

## 🌍 Why this has not shown up

The defect is asymmetric:

- **Negative offsets** (Americas): local midnight converts to a later hour on the *same* UTC day, so `.Date` is unchanged. No visible effect.
- **Positive offsets** (Europe, Asia): local midnight converts to the previous UTC day. Off by one, always.

CI runs on UTC, so it never reproduces there.

## 🧪 The unit tests currently encode the behavior

Three tests in `src/powershell/Tests/Unit/Start-FinOpsCostExport.Tests.ps1` fail on any machine east of UTC:

```
[-] Should call /run with default end date
[-] Should call /run for backfill
[-] Should adjust end date if it would be in the future
    Expected Invoke-Rest ... to be called at least 1 times, but was called 0 times
```

They build their expected values with `Get-Date -Month 1 -Day 1 -Hour 0 ... -AsUTC`, which converts local midnight to UTC and lands on 31 December, and then compare against a format string with a hardcoded `yyyy-01-01`. Both sides shift, so the `-ParameterFilter` never matches.

This is worth calling out because "just fix the tests" would cement the product behavior. The tests are a symptom, not the defect.

## 🤔 Question before a fix

Are `-StartDate` / `-EndDate` meant to be **calendar dates** or **instants**?

The help text reads like calendar dates ("Day to start pulling the data for", "Last day to pull data for"), and Cost Management export periods are day-granular, which suggests the intended handling is to keep the day the caller named and not convert time zones:

```powershell
$StartDate = [datetime]::SpecifyKind($StartDate.Date, [System.DateTimeKind]::Utc)
```

That is a behavior change for existing users at positive offsets (it would start exporting the day they actually asked for) and a no-op for everyone else, so it seems worth confirming the intent before changing it. Happy to submit a PR covering the command and the three tests together once the direction is confirmed.

## 🔧 Environment

- **FinOps hub version:** current `dev`; the pattern is unchanged in v13 and v14
- **PowerShell:** 7.6.3 on Windows, machine time zone UTC+01:00 (UTC+02:00 during DST)
- **Billing account type:** not relevant — the shift happens before any API call

## ℹ️ Additional context

Found while verifying the R15 release candidate end to end. The three failing tests are also the reason the local suite is not green outside the CI time zone; they are unrelated to the two workflow-pinning failures fixed by #2252.

RolandKrummenacher:
## 🛠️ Description

Fixes #2255.

`Start-FinOpsCostExport` converted `-StartDate` and `-EndDate` with `ToUniversalTime()` before truncating them to a day. For any positive UTC offset, local midnight falls on the *previous* UTC day, so `.Date` moved the requested period back by one day.

```powershell
# machine in UTC+01:00
$in = [datetime]'2026-01-01'   # how the binder parses the parameter: local midnight
$in.ToUniversalTime().Date     # -> 2025-12-31
```

The parameters are now treated as **UTC calendar dates**: the day the caller names is tagged as UTC rather than converted.

```powershell
[datetime]::SpecifyKind($StartDate.Date, [DateTimeKind]::Utc)
```

This matches how the parameters are documented ("Day to start pulling the data for", "Last day to pull data for") and how Cost Management export periods actually work — they are UTC and day-granular, so there is no instant to convert.

### The defect is asymmetric

| Local offset | Old behavior | New behavior |
| --- | --- | --- |
| UTC | `2026-01-01` | `2026-01-01` |
| UTC+01:00 (Europe) | `2025-12-31` ❌ | `2026-01-01` |
| UTC+05:30 (India) | `2025-12-31` ❌ | `2026-01-01` |
| UTC+09:00 (Japan) | `2025-12-31` ❌ | `2026-01-01` |
| UTC-05:00 (US East) | `2026-01-01` | `2026-01-01` |
| UTC-08:00 (US West) | `2026-01-01` | `2026-01-01` |

At UTC and negative offsets, local midnight converts to a later hour on the *same* UTC day, so `.Date` was already unchanged. **Behavior there is identical before and after this change.** CI runs on UTC, which is why this never reproduced there.

### `-Backfill` ran an extra month

The default backfill start date used the same pattern, landing one day early — which pushed the computed month count up by one. On a UTC+02:00 machine, `-Backfill 3` issued **5** export runs (April through August) instead of 4 (May through August). The default is now derived from the current UTC month, so the count is correct in every time zone.

<br>

## 🧪 Testing

Three tests in `Start-FinOpsCostExport.Tests.ps1` encoded the shifted behavior — they built their expected values with the same `ToUniversalTime()` conversion applied to both sides, so they failed on any machine east of UTC:

```
[-] Should call /run with default end date
[-] Should call /run for backfill
[-] Should adjust end date if it would be in the future
```

Fixing only the tests would have cemented the product behavior, so the assertions were rewritten to be time zone independent instead:

- Date expectations are now **literal** strings (`'2024-01-01T00:00:00Z'`) rather than expressions that shift in step with the code under test.
- Date-derived tests compute from `(Get-Date).ToUniversalTime().Date` so they hold at any offset.
- Added `-Exactly` to the invocation counts, which is what catches the extra `-Backfill` month.
- Added a test covering every `DateTimeKind` the parameter binder can produce (`Local`, `Unspecified`, and the `Utc` path via the existing tests).

Results on a UTC+02:00 machine (PowerShell 7.6.3, Pester 6.1.0):

- Before: `Tests Passed: 5, Failed: 3`
- After: `Tests Passed: 9, Failed: 0`
- Full unit suite: `Tests Passed: 2373, Failed: 0, Skipped: 4`
- `Invoke-ScriptAnalyzer` clean on both changed PowerShell files

Because the change is a no-op at UTC, the existing CI run is the control: it should stay green.

<br>

## 📝 Notes

- This is a **behavior change for users at positive UTC offsets** — they will start exporting the day they actually asked for, and `-Backfill n` will run `n + 1` exports instead of `n + 2`. It is a no-op for UTC and negative offsets.
- Parameter help and the [command reference page](https://github.com/microsoft/finops-toolkit/blob/dev/docs-mslearn/toolkit/powershell/cost/start-finopscostexport.md) now state the UTC calendar-date semantics explicitly.
- Out of scope, noted while tracing this: passing `-EndDate` without `-StartDate` leaves `$StartDate` at `DateTime.MinValue`, so the month count becomes ~24,000 and the run loop iterates accordingly. It is unrelated to time zones and is not touched here — happy to open a separate issue.

<br>

## ✅ Checklist

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Unit tests added/updated and passing
- [x] Documentation updated
- [x] Changelog updated

### DIFF
--- a/docs-mslearn/toolkit/changelog.md
+++ b/docs-mslearn/toolkit/changelog.md
@@ -3,7 +3,7 @@ title: FinOps toolkit changelog
 description: Review the latest features and enhancements in the FinOps toolkit, including updates to FinOps hubs, Power BI reports, and more.
 author: MSBrett
 ms.author: brettwil
-ms.date: 08/13/2026
+ms.date: 08/17/2026
 ms.topic: reference
 ms.service: finops
 ms.subservice: finops-toolkit
@@ -58,6 +58,11 @@ The following section lists features and enhancements that are currently in deve
 
 - **Changed**
   - Switched the reservations and benefits workbooks from the retired `ccmstorageprod` isfratioblob.csv to the FinOps toolkit [Instance size flexibility](open-data.md#instance-size-flexibility) open data file ([#2090](https://github.com/microsoft/finops-toolkit/issues/2090)).
+
+### [PowerShell module](powershell/powershell-commands.md)
+
+- **Fixed**
+  - Fixed [Start-FinOpsCostExport](powershell/cost/start-finopscostexport.md) exporting the wrong period for anyone running in a positive UTC offset. `-StartDate` and `-EndDate` are now treated as UTC calendar dates instead of being time zone converted, so the days you request are the days that get exported. Previously, local midnight converted to the previous UTC day, which moved the period back a day and made `-Backfill` run one extra month ([#2255](https://github.com/microsoft/finops-toolkit/issues/2255)).
 
 ### [Open data](open-data.md) updates
 
--- a/docs-mslearn/toolkit/powershell/cost/start-finopscostexport.md
+++ b/docs-mslearn/toolkit/powershell/cost/start-finopscostexport.md
@@ -3,7 +3,7 @@ title: Start-FinOpsCostExport command
 description: Initiate a Cost Management export run for the most recent period using the Start-FinOpsCostExport command in the FinOpsToolkit module.
 author: flanakin
 ms.author: micflan
-ms.date: 04/01/2026
+ms.date: 08/17/2026
 ms.topic: reference
 ms.service: finops
 ms.subservice: finops-toolkit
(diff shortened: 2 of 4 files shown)


MSBrett (commented):
Reviewed this — the fix is correct and well-verified. I re-ran the unit tests under `TZ=Asia/Tokyo` and `TZ=America/Los_Angeles` to reproduce the exact asymmetric bug from #2255, and they pass on this branch in both (and fail as expected on old logic). Lint suite is clean too.

One non-blocking item: this branch currently shows as `CONFLICTING` against `dev` — looks like just the `ms.date` field in `changelog.md` (`08/17` vs `08/19`), which matches the repo's documented conflict pattern. A `git merge origin/dev` with `ms.date` set to today should clear it.

The two out-of-scope issues you called out in the description (missing-`-StartDate` month-count blowup, and the millisecond- vs. day-truncated backfill boundary) both check out on my read — good catch calling those out separately rather than scope-creeping this fix.

MSBrett (approved):
Approving — my earlier review found only two Low/informational items (the pending `ms.date` merge conflict, now resolved via `git merge origin/dev` + this branch's own automated `ms.date` bot commit; and two pre-existing, explicitly out-of-scope issues unrelated to this fix). No Critical/High/Medium findings. Tests pass (verified under `TZ=Asia/Tokyo` and `TZ=America/Los_Angeles` to reproduce the original bug conditions), lint is clean, and CI is green.

MSBrett (approved):

PR #1800, opened 2025-08-11
Break out bicep modules for hubs extensibility

flanakin:
## 🛠️ Description
This PR reorganizes FinOps hubs bicep into a cleaner, more extensible structure while fixing build issues and improving maintainability.

- Moved all JSON schemas into relevant module directories
- Split large monolithic files (e.g., 5K+ line dataFactory.bicep) into smaller modules
- Moved all framework modules into the `fx` folder
- Organized apps into `X.Y/Z` folders for each publisher/app:
  - Microsoft.CostManagement/Exports – Cost export functionality
  - Microsoft.CostManagement/ManagedExports – Managed exports
  - Microsoft.FinOpsHubs/Analytics – Data Explorer and analytics
  - Microsoft.FinOpsHubs/Core – Core hub infrastructure
  - Microsoft.FinOpsHubs/RemoteHub – Remote hub functionality
- Resolved all Bicep compilation errors/warnings
- Added `#disable-next-line BCP318` with explanations for null safety
- Fixed Build-Toolkit.ps1 param generation bug
- Updated PowerShell build scripts with correct paths

### 🔬 How did you test this change?

> - [ ] 🤏 Lint tests
> - [ ] 🤞 PS -WhatIf / az validate
> - [x] 👍 Manually deployed + verified
> - [ ] 💪 Unit tests
> - [ ] 🙌 Integration tests

### 🙋‍♀️ Do any of the following that apply?

> - [ ] 🚨 This is a breaking change.
> - [ ] 🤏 The change is less than 20 lines of code.

### 📑 Did you update `docs/changelog.md`?

> - [x] ✅ Updated changelog (required for `dev` PRs)
> - [ ] ➡️ Will add log in a future PR (feature branch PRs only)
> - [ ] ❎ Log not needed (small/internal change)

### 📖 Did you update documentation?

> - [ ] ✅ Public docs in `docs` (required for `dev`)
> - [ ] ✅ Internal dev docs in `src` (required for `dev`)
> - [ ] ➡️ Will add docs in a future PR (feature branch PRs only)
> - [x] ❎ Docs not needed (small/internal change)

PR #2262, opened 2026-08-17
fix(hubs): assign requested storage roles to apps that don't create storage (#2253)

### LINKED ISSUE #2253: [Hubs] Managed exports fail with Unauthorized — RBAC Administrator role is never assigned (regression since v13)
## 🐛 Problem

Managed exports never create any Cost Management exports. The `config_ConfigureExports` pipeline runs, reaches the export creation activity, and fails with:

```
{"error":{"code":"Unauthorized","message":"The user does not have authorization to perform
'Microsoft.Authorization/roleAssignments/write' action on specified storage account, please use a
storage account with sufficient permissions. If the permissions have changed recently then retry
after some time."}}
```

Cost Management needs the caller to be able to write role assignments on the destination storage account, because the export is created with a system-assigned identity (`"identity": { "type": "systemAssigned" }`) for which it must grant access to the destination.

The template is aware of this and passes the **Role Based Access Control Administrator** role for the Data Factory identity — but the assignment is never actually created, so the Data Factory identity cannot create exports.

## 🔍 Root cause

The Managed Exports app registers itself with only the `DataFactory` feature while passing `storageRoles`:

```bicep
// src/templates/finops-hub/modules/Microsoft.CostManagement/ManagedExports/app.bicep#L69-L82
module appRegistration '../../fx/hub-app.bicep' = {
  name: 'Microsoft.CostManagement.ManagedExports_Register'
  params: {
    features: [
      'DataFactory'          // <-- 'Storage' is not declared
    ]
    storageRoles: [
      // RBAC Administrator -- used to create Cost Management exports (which require access to grant access)
      'f58310d9-a9f6-439a-9e8d-f62e7b41a168'
    ]
  }
}
```

`hub-app.bicep` gates all storage role assignments on that feature:

```bicep
// src/templates/finops-hub/modules/fx/hub-app.bicep#L40
var usesStorage = contains(features, 'Storage')

// src/templates/finops-hub/modules/fx/hub-app.bicep#L287-L298
resource storageRoleAssignments 'Microsoft.Authorization/roleAssignments@2022-04-01' = [
  for role in factoryStorageRoles: if (usesDataFactory && usesStorage) {
    ...
  }
]
```

Because `usesStorage` is `false` for this app, the entire loop is skipped and the RBAC Administrator role it passes is silently dropped. Managed Exports is the only app in the template that passes `storageRoles` at all — and the only one of those that does not declare the `Storage` feature.

This is visible in the shipped template as well: the role GUID appears exactly once in `azuredeploy.json`, as the parameter value handed to `Microsoft.CostManagement.ManagedExports_Register`, immediately next to `"features": ["DataFactory"]`.

## 📉 Regression

This worked in v12, where the same role was coupled directly to the managed exports option and assigned without a feature condition:

```bicep
// v12 -- src/templates/finops-hub/modules/dataFactory.bicep
var storageRbacRoles = union(
  [ /* Storage Account Contributor, Storage Blob Data Contributor, Reader */ ],
  // Only use User Access Administrator if managed exports are enabled for least privileged access
  !enableManagedExports ? [] : [ '18d7d88d-d35e-4fb5-a5c3-7773c20a72d9' ]
)

resource factoryIdentityStorageRoleAssignments 'Microsoft.Authorization/roleAssignments@2022-04-01' = [
  for role in storageRbacRoles: { ... }   // no feature condition
]
```

The app modularization (#1800) moved roles into a per-app `storageRoles` parameter and introduced the feature gate; the coupling to `enableManagedExports` was lost in the process. Present in v13, v14 and current `dev`.

## 👣 Repro steps

1. Deploy a FinOps hub with `enableManagedExports = true` and at least one subscription scope in `scopesToMonitor`.
2. Grant the hub managed identity **Cost Management Contributor** on that scope, as documented.
3. Write to `config/settings.json` in the hub storage account (any write) so the `config_SettingsUpdated` trigger fires.
4. `config_ConfigureExports` runs and fails at the `... open month focus export` activity with the `Unauthorized` error above (3 attempts).
5. No export is created in Cost Management, so nothing lands in the `msexports` container and no data is ingested.

Checking the Data Factory identity afterwards shows four role assignments — Data Factory Contributor on the factory, plus Storage Account Contributor, Storage Blob Data Contributor and Reader on the hub storage account. RBAC Administrator is not among them.

## 🤔 Expected

When `enableManagedExports` is enabled, the deployment assigns the RBAC Administrator role to the Data Factory identity on the hub storage account, so `config_ConfigureExports` can create Cost Management exports. This matches the documented behavior in [Hub template](https://microsoft.github.io/finops-toolkit/hubs/template), which lists the role under the assignments made "as part of the deployment", noting only that it is *"not applied when enableManagedExports is disabled"*.

## 🔧 Environment

- **FinOps hub version:** current `dev` (15.0-dev); also present in v13 and v14. Last known good: v12.
- **Billing account type:** reproduced on a subscription scope, which is documented as supported for managed exports (MCA billing accounts/profiles are explicitly not supported).
- **Cost Management export:** FocusCost

Reproduced on two independent deployments with different shapes (with and without Data Explorer, public networking), with the same result.

## 💡 Suggested fix

Declare the `Storage` feature on the Managed Exports app registration so the role assignment loop runs:

```bicep
features: [
  'DataFactory'
  'Storage'
]
```

## ℹ️ Additional context

Two related observations found while investigating, both minor — happy to split them into their own issues if preferred:

1. The role table in the hub template docs still lists **User Access Administrator**. That GUID no longer appears anywhere in the shipped template since it was replaced by RBAC Administrator, so the documented role cannot be found by anyone looking for it.

2. In the same pipeline run, the `Save Scopes` activity of `config_ConfigureExports` fails when the settings file contains exactly one scope:

   ```
   The variable 'scopesArray' of type 'Array' cannot be initialized or updated with value of type 'Object'.
   ```

   The settings file serializes a single scope as an object rather than an array, because sorting a one-element collection in PowerShell yields a scalar. This is non-blocking — the following `Save Scopes as Array` activity recovers and the scope reaches export creation — but it shows up as a failed activity on every run with a single scope and makes troubleshooting harder.

RolandKrummenacher:
## 🛠️ Description

Fixes #2253.

Managed exports never created any Cost Management exports. `config_ConfigureExports` reached the export creation activity and failed with:

```
{"error":{"code":"Unauthorized","message":"The user does not have authorization to perform
'Microsoft.Authorization/roleAssignments/write' action on specified storage account, ..."}}
```

Cost Management creates exports with a system-assigned identity and must grant that identity access to the destination, so the caller needs to write role assignments on the destination storage account. The Managed Exports app requests **Role Based Access Control Administrator** for the Data Factory identity via the `storageRoles` parameter of `fx/hub-app.bicep`, but the assignment was never created.

### Root cause

The storage role assignment loop was gated on the `Storage` feature:

```bicep
var usesStorage = contains(features, 'Storage')

resource storageRoleAssignments 'Microsoft.Authorization/roleAssignments@2022-04-01' = [
  for role in factoryStorageRoles: if (usesDataFactory && usesStorage) { ... }
]
```

Apps only declare `Storage` when they *create* the publisher storage account. Managed Exports uses the account created by `Microsoft.CostManagement.Exports`, so `usesStorage` is `false` and every role it passed was silently dropped at build time. It is the only app that passes `storageRoles`, so it was the only one affected.

This regressed in the app modularization (#1800): in v12 the same role was coupled directly to `enableManagedExports` and assigned with no feature condition.

### Fix

Roles requested via `storageRoles` are assigned whenever the app uses Data Factory, since the storage account is shared across all apps from the same publisher. The base data management roles (Storage Account Contributor, Storage Blob Data Contributor, Reader) stay tied to the `Storage` feature, so nothing changes for any other app.

```bicep
var factoryStorageRoles = union(storageRoles, !usesStorage ? [] : [ /* base data management roles */ ])

resource storageRoleAssignments 'Microsoft.Authorization/roleAssignments@2022-04-01' = [
  for role in factoryStorageRoles: if (usesDataFactory) { ... }
]
```

<details>
<summary>Why not declare the <code>Storage</code> feature on Managed Exports (the fix suggested in the issue)?</summary>

`usesStorage` doesn't only gate role assignments — it also deploys the storage account, both private endpoints, the private DNS zone groups, the ADF managed private endpoint, the storage linked service, and the `Get`/`ApproveStoragePrivateEndpointConnections` nested deployments. Adding the feature to Managed Exports would duplicate all of that against a storage account another app already owns, and under private routing `Microsoft.CostManagement.ManagedExports` has no `dependsOn` sequencing against `Microsoft.FinOpsHubs.Analytics`, so the two would race on identically named private endpoint resources and nested deployments.

Fixing the gate instead keeps the blast radius to the one missing role assignment and also fixes the class of bug for any future app that requests storage roles.
</details>

### Docs

Item 1 from the issue: the role table and the `enableManagedExports` tooltip/description still named **User Access Administrator**, which hasn't appeared anywhere in the template since v13 replaced it with RBAC Administrator. Corrected in `template.md`, `main.bicep`, `hub.bicep`, and `createUiDefinition.json`.

Item 2 from the issue (`Save Scopes` failing when the settings file has exactly one scope) is **not** addressed here — it is non-blocking, the recovery activity already handles it, and the real fix belongs in the PowerShell that serializes `settings.json`. Happy to file it separately.

## 🧪 Testing

- `bicep build src/templates/finops-hub/main.bicep` — clean (only pre-existing `no-unused-params` warnings in `Recommendations/app.bicep`).
- Verified against the compiled ARM template that `Microsoft.CostManagement.ManagedExports_Register` now emits a role assignment for `f58310d9-a9f6-439a-9e8d-f62e7b41a168` on the storage account, and that `factoryStorageRoles` is unchanged for every other app.
- New `src/powershell/Tests/Unit/HubsAppStorageRoles.Tests.ps1` regression suite (6 tests): the assignment loop isn't gated on `usesStorage`, `storageRoles` always flows into it, every app requesting storage roles declares `DataFactory`, and Managed Exports still requests RBAC Administrator. Confirmed the suite fails against the pre-fix condition.
- Full `Tests/Unit` (2379 passed, 4 skipped) and `Tests/Lint` (3587 passed) suites pass.

Not yet validated with a live hub deployment — worth a deploy test with `enableManagedExports = true` before merge.

### Reviewer tasks

- [ ] Reviewed the fix approach (gate change vs. declaring the `Storage` feature)
- [ ] Verified the role assignment appears on a test deployment

### Contributor tasks

- [x] Updated the changelog
- [x] Updated docs
- [x] Added tests


MSBrett (approved):
Independently re-verified the fix: `bicep build` is clean, and the new `HubsAppStorageRoles.Tests.ps1` suite passes with the fix and fails against the pre-fix code exactly as described. The deployment-ordering concern raised for the rejected alternative doesn't apply to the shipped fix, since `cmManagedExports` already has an implicit dependency on `cmExports` (via the required `exports` parameter) — so the shared storage account is guaranteed to exist before the role assignment runs. No functional or security issues found.

Also merged `origin/dev` into this branch to resolve the changelog.md conflict (adjacent-bullet insertion + differing `ms.date`); all other files auto-merged cleanly, and the full test suite still passes post-merge.

flanakin (approved):

PR #2272, opened 2026-08-20
Fix Claude plugin marketplace and agent discovery

MSBrett:
## 🛠️ Description

Correct the Claude Code marketplace source so it resolves to the repository's actual plugin root, enable strict marketplace validation, and load the shared agents through Claude's standard `agents/` auto-discovery. Agent files now use plain `.md` names so Claude exposes the intended identifiers without an `.agent` suffix. The MCP configuration remains declared explicitly.

Validated against Anthropic's current documentation:
- [Plugin manifest schema](https://code.claude.com/docs/en/plugins-reference#plugin-manifest-schema)
- [Marketplace plugin entries and strict mode](https://code.claude.com/docs/en/plugin-marketplaces#plugin-entries)
- [Repository-root-relative plugin sources](https://code.claude.com/docs/en/plugin-marketplaces#relative-paths)
- [Plugin subagent discovery](https://code.claude.com/docs/en/sub-agents#choose-the-subagent-scope)

A clean installation from this PR's online branch with Claude Code CLI 2.1.235 reports one skill, five correctly named agents, and one MCP server.

Fixes # N/A

## 📷 Screenshots

N/A

> - [ ] 🤏 Lint tests
> - [x] 💪 Unit tests
> - [ ] 🙌 Integration tests

Additional validation:
- `claude plugin validate . --strict`
- `claude plugin validate src/templates/agent-plugin --strict`
- Clean online-branch marketplace installation and component inventory with Claude Code CLI 2.1.235

### 📦 Deploy to test?

> - [ ] Hubs + ADX (managed)
> - [ ] Hubs (manual)
> - [ ] Hubs (no data)
> - [ ] Workbooks
> - [ ] Alerts


src/powershell/Tests/Unit/AgentPlugins.Tests.ps1
```diff
@@ -48,11 +48,20 @@ Describe 'Agent plugin manifest' {
...
 
+    It 'Declares resolvable paths for every Claude component' {
+        $claude = Get-Content (Join-Path $script:Plugin '.claude-plugin/plugin.json') -Raw | ConvertFrom-Json
+        $paths = @($claude.commands) + @($claude.skills) + @($claude.mcpServers) + @($claude.outputStyles)
+
+        $paths.Count | Should -Be 4
+        $paths | ForEach-Object {
+            Join-Path $script:Plugin $_ | Should -Exist
+        }
+    }
```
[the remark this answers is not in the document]
MSBrett: Updated the test to require at least the four expected components while allowing additional paths, and each path is now explicitly checked for a non-empty value and existence. The targeted Pester suite passes all 24 tests.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

.claude-plugin/marketplace.json
```diff
@@ -9,7 +9,8 @@
     {
       "name": "microsoft-finops-toolkit",
       "version": "15.0.0",
-      "source": "./plugins/microsoft-finops-toolkit",
+      "source": "./src/templates/agent-plugin",
```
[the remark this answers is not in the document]
MSBrett: No source change is needed here. Both paths are tracked symlinks into this same marketplace repository (`docs-mslearn` and `src/queries`), and Claude's documented marketplace caching behavior dereferences targets elsewhere within the marketplace and copies their contents into the plugin cache. A clean install of this PR at `214dbcb4` contains 124 Learn Markdown files and 40 query files (including `queries/INDEX.md`) with no remaining symlinks, so the query and documentation workflows have their required runtime files. `Build-AgentPlugin.ps1` remains the expansion path for other packaged distribution targets. Reference: https://code.claude.com/docs/en/plugins-reference#share-files-within-a-marketplace-with-symlinks

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

RolandKrummenacher (commented):
Verified the premise empirically rather than by reading alone: installed both the `dev` and PR versions of the plugin into isolated `CLAUDE_CONFIG_DIR`s and compared `claude plugin details`, and reproduced the symlink-source failure in a scratch marketplace.

Confirmed:

- On `dev`, the `.agent.md` filenames cause Claude to derive agent ids from the filename stem and ignore frontmatter `name:`, so agents surface as `azure-capacity-manager.agent`, `ftk-database-query.agent`, etc. This PR's version correctly exposes 5 unsuffixed agents, 1 skill, 1 MCP server.
- The `./plugins/microsoft-finops-toolkit` source is genuinely unusable on Windows checkouts (the git symlink materializes as a text file, so install fails with `ENOTDIR`).
- `claude plugin validate --strict` passes on both manifests, and the validator does reject unknown fields, so `strict` is a recognized marketplace field.
- All 24 Pester tests pass locally.

The fix works for Claude Code. My concern is what it costs on the other platform.

## Renaming the agents breaks GitHub Copilot CLI discovery

(Left here rather than inline, since the five renames carry no diff hunks to anchor to.)

The root `plugin.json` exists specifically for Copilot and declares `"agents": "./agents/"`. GitHub's plugin format requires `NAME.agent.md`, so dropping the suffix takes Copilot from 5 agents to 0 — while the changelog still describes the plugin as shared between Claude Code and Copilot CLI. Claude auto-discovers `agents/` regardless of what the manifest says, so adding a second directory doesn't resolve this; the Copilot-suffixed variant probably needs to be generated at package time in `Build-AgentPlugin.ps1`.

## README is now out of date

`src/templates/agent-plugin/README.md:7` isn't in this PR, but it now contradicts both changes — it still says the plugin directory is "used by both marketplace manifests", and that Claude declares only its output style.

Remaining comments inline.

src/powershell/Tests/Unit/AgentPlugins.Tests.ps1
```diff
@@ -167,7 +181,7 @@ Describe 'Plugin discovery and marketplaces' {
     It 'Uses repository-root-relative marketplace sources' {
         $marketplaces = @{
             '.github/plugin/marketplace.json' = './plugins/microsoft-finops-toolkit'
-            '.claude-plugin/marketplace.json' = './plugins/microsoft-finops-toolkit'
+            '.claude-plugin/marketplace.json' = './src/templates/agent-plugin'
```
RolandKrummenacher: Only the Claude marketplace manifest was repointed — `.github/plugin/marketplace.json` still resolves to `./plugins/microsoft-finops-toolkit`, so half the reported bug survives for Copilot CLI users on Windows checkouts (symlink materializes as a text file, install fails with `ENOTDIR`).

This test now asserts two *different* source values under a name that claims they're consistent ("Uses repository-root-relative marketplace sources"), so nothing will flag the stale one. If the divergence is intentional, worth renaming the test and adding a comment explaining why the Copilot manifest keeps the symlink path.

src/powershell/Tests/Unit/AgentPlugins.Tests.ps1
```diff
@@ -64,13 +74,17 @@ Describe 'Agent plugin manifest' {
 }
 
 Describe 'Agent plugin components' {
-    It 'Ships agent definitions as NAME.agent.md files' {
-        $agents = Get-ChildItem (Join-Path $script:Plugin 'agents') -Filter '*.agent.md'
-        $agents.Count | Should -BeGreaterThan 0
+    It 'Ships agent definitions as Claude-compatible NAME.md files' {
+        $agents = Get-ChildItem (Join-Path $script:Plugin 'agents') -Filter '*.md'
+
+        $agents.Count | Should -Be 5
```
RolandKrummenacher: The hardcoded `Should -Be 5` combined with the widened `*.md` glob makes this brittle in two directions: adding a sixth agent fails here, and dropping an `agents/README.md` into the directory fails both this count and the front-matter assertion below — in each case with a message that points at the wrong problem. `Should -BeGreaterThan 0` plus the suffix assertion already covers the regression this PR is guarding against.

Separately: there's no test covering the behavior the PR actually fixes — that the installed agent ids come out unsuffixed. The suffix check on filenames is a proxy for it.

src/templates/agent-plugin/.claude-plugin/plugin.json
```diff
@@ -14,5 +14,6 @@
   "skills": [
     "./skills/"
   ],
+  "mcpServers": "./.mcp.json",
```
RolandKrummenacher: This is a no-op — I verified `dev` already loaded exactly 1 MCP server through auto-discovery of `.mcp.json` at the plugin root. Declaring it explicitly just merges with the auto-discovered config and adds a third spelling of the same path (`.mcp.json` in the root manifest, `./.mcp.json` here), plus two exact-string assertions in the test file to keep in sync.

Not harmful, just extra surface. If the intent is to be explicit for readers, a comment in the README would carry it more cheaply.

MSBrett: Removed the redundant Claude `mcpServers` declaration and its exact-path assertions. The test now documents root `.mcp.json` auto-discovery, and an isolated Claude marketplace install still reports the expected `azure-mcp-server` configuration.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/powershell/Tests/Unit/AgentPlugins.Tests.ps1
MSBrett: Repointed the Copilot marketplace to `./src/templates/agent-plugin` as well, so neither host depends on the repository symlink on Windows. The test now enforces the same cross-platform source for both manifests, and the README documents that distribution layout.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/powershell/Tests/Unit/AgentPlugins.Tests.ps1
MSBrett: Relaxed the agent count to require at least one definition and added a direct identity invariant: every agent filename stem must equal its frontmatter `name`, which specifically catches the `.agent` suffix regression. I also retested from an isolated `COPILOT_HOME`; Copilot CLI 1.0.80 installed the marketplace and successfully invoked `microsoft-finops-toolkit:ftk-database-query`, so the shared `NAME.md` agents remain discoverable by both hosts.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/powershell/Tests/Unit/AgentPlugins.Tests.ps1
```diff
@@ -64,15 +74,23 @@ Describe 'Agent plugin manifest' {
...
 
 Describe 'Agent plugin components' {
-    It 'Ships agent definitions as NAME.agent.md files' {
-        $agents = Get-ChildItem (Join-Path $script:Plugin 'agents') -Filter '*.agent.md'
+    It 'Ships agent definitions as Claude-compatible NAME.md files' {
+        $agents = Get-ChildItem (Join-Path $script:Plugin 'agents') -Filter '*.md'
+
         $agents.Count | Should -BeGreaterThan 0
+        $agents.Name | ForEach-Object {
+            $_ | Should -Not -Match '\.agent\.md$'
```
RolandKrummenacher, line 82: This inverts the previous guard, so the `.agent.md` naming can no longer be restored without also editing this test — a follow-up PR fixing Copilot discovery would fail CI here.

If the rename is reverted, this assertion has to go with it. Worth keeping the naming change and its enforcement in the same commit either way, so CI never encodes a convention the plugin manifests contradict.

src/templates/agent-plugin/README.md
```diff
@@ -4,9 +4,9 @@ This plugin is the canonical source for the FinOps Toolkit agent plugins.
 
 ## Distribution and runtime conventions
 
-- `plugins/microsoft-finops-toolkit` is the marketplace source used by both marketplace manifests.
+- Both marketplace manifests point directly to this directory to avoid relying on repository symlink support.
```
RolandKrummenacher: This claim isn't achieved on the branch. `skills/finops-toolkit/references/docs-mslearn` and `references/queries` are still committed symlinks — I checked this branch out in a fresh Windows worktree and they materialize as ~30-byte text files, so a direct install from this directory still has broken skill references.

The marketplace `source` change is a real improvement regardless; it's just this sentence that overstates it. Either resolve those two symlinks as well, or soften the wording to describe what the change actually does.

.claude-plugin/marketplace.json
RolandKrummenacher, line 12: With both manifests now pointing here, the `plugins/microsoft-finops-toolkit` symlink has no remaining references anywhere on the branch — no grep hits — but is still committed. Worth removing it in this PR rather than leaving a dead pointer for the next reader to trip over.

Separately, `strict: true` on the next line is the documented default for marketplace plugin entries. Valid, but a no-op, and the new test now pins it. Fine to keep if it's deliberate belt-and-braces.

RolandKrummenacher (changes requested):
Re-reviewed at 2a0755b. The intent here is right — pointing the marketplace `source` at a real directory instead of a symlink is a genuine fix, and the modified tests pass in a clean worktree (24/24). One blocker.

## Blocking: the `.agent.md` → `.md` rename breaks Copilot CLI agent discovery

GitHub Copilot CLI requires the `.agent.md` extension to discover agents. The root `plugin.json` still declares `"agents": "./agents/"` for Copilot, and `.github/plugin/marketplace.json` exists solely to serve it — so after this change Copilot finds no agents at all. f3935ac2 (#2167) renamed these five files *to* `.agent.md` for exactly that reason.

The rename also isn't needed on the Claude side. Claude Code derives the subagent identifier from the front-matter `name:` field, and the docs state plainly that the filename doesn't have to match — so `finops-practitioner.agent.md` was already exposed as `finops-practitioner`, never `finops-practitioner.agent`.

Suggested resolution: keep `.agent.md`, and revert the test assertion that now enforces the opposite (inline comment below). If something in Claude discovery was genuinely failing, could you capture the actual failure? That would point at the real cause, which this rename may be masking rather than fixing.

One caveat on my side: the Copilot breakage is from its documentation and the stated intent of #2167 — I have not run Copilot CLI against this branch. If you have a run showing `.agent.md` files still being discovered there, that settles it and I'll withdraw the blocker.

## Non-blocking

A symlink that is now unreferenced but still committed, and a README claim the branch doesn't actually deliver. Both inline.

## Checked and cleared

Relative `source` paths do resolve from the repo root, so `./src/templates/agent-plugin` is correct; dropping `agents` and `mcpServers` from `.claude-plugin/plugin.json` is fine because Claude auto-discovers both from the plugin root; all five agents' front-matter `name:` values match their new basenames.

src/templates/agent-plugin/agents/finops-practitioner.md
RolandKrummenacher, line 1: **Blocker** — this rename, and the other four in `agents/`, breaks GitHub Copilot CLI agent discovery, which requires the `.agent.md` extension. Full reasoning and suggested resolution in the review summary.

src/templates/agent-plugin/README.md
RolandKrummenacher: Withdrawing the runtime half of this. @MSBrett already answered the same point on the earlier `marketplace.json` thread with a clean install at `214dbcb4` — 124 Learn Markdown files and 40 query files, no remaining symlinks — because Claude's marketplace install dereferences in-marketplace symlink targets into the plugin cache. My check was a plain `git checkout` in a Windows worktree, which is the wrong path to test; it doesn't tell you what the install produces.

So the skill references are not broken, and the only thing left here is wording: the sentence says the change avoids relying on repository symlink support, while `references/docs-mslearn` and `references/queries` are still symlinks that the install path resolves. Purely editorial — resolving this thread.

src/templates/agent-plugin/agents/finops-practitioner.md
MSBrett: Thanks for flagging this. The PR already contains the branch-level evidence covered by your caveat:

- In your initial review, you reported testing isolated Claude installs of `dev` and this PR. You observed that `dev` exposed the `.agent.md` files with `.agent` in their IDs, while this PR exposed all five agents with the intended unsuffixed IDs.
- In [this earlier thread](https://github.com/microsoft/finops-toolkit/pull/2272#discussion_r3865911321), I reported testing the shared `NAME.md` files through an isolated `COPILOT_HOME` with Copilot CLI 1.0.80. The marketplace installed successfully, discovered the agents, and successfully invoked `microsoft-finops-toolkit:ftk-database-query`.

So although GitHub's authoring documentation uses the `.agent.md` convention, the current Copilot CLI plugin loader does discover these plain `.md` agent files. Keeping `.md` is necessary for Claude to expose the intended IDs, and the same files remain functional in Copilot CLI.

Given your note that a successful Copilot CLI run against this branch would settle the concern, could you please withdraw the blocker?

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

RolandKrummenacher (approved):
Withdrawing the blocker.

On the `.agent.md` naming: my condition was that the rename and its CI enforcement land together so nothing on the branch contradicts the convention. That holds — there are no remaining `.agent.md` references anywhere in the repo. Combined with your Copilot CLI 1.0.80 run against an isolated `COPILOT_HOME`, which discovered and invoked the agents from the plain `NAME.md` files, the naming question is settled for both hosts.

On the `plugins/microsoft-finops-toolkit` symlink: it's now unreferenced repo-wide, but it already exists on `dev` — this PR only orphans it rather than introducing it. That makes the cleanup optional here; happy for it to be a follow-up. Leaving the thread open so it doesn't get lost.

.claude-plugin/marketplace.json
MSBrett: Good catch. Removed the now-unreferenced `plugins/microsoft-finops-toolkit` symlink in `8da80ac95`.

Both marketplace manifests already point directly to `./src/templates/agent-plugin`, and there are no remaining references to the old symlink path. The targeted `AgentPlugins.Tests.ps1` suite still passes all 24 tests. I left the explicit `strict: true` setting unchanged as the intentional belt-and-braces choice noted here.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

.claude-plugin/marketplace.json
```diff
@@ -9,7 +9,8 @@
...
+      "strict": true,
```
flanakin: 🤖 [AI][Claude Code] ⚠️ **Should fix**

`"strict": true` is the documented default, so this line changes nothing.

Anthropic's [marketplace reference](https://code.claude.com/docs/en/plugin-marketplaces#optional-plugin-fields) defines it as *"Controls whether `plugin.json` is the authority for component definitions (default: true)"*, and the Strict mode section adds that `strict: true` "is the default and works for most plugins."

Two follow-ons:

- The PR description calls this "enable strict marketplace validation." `strict` has nothing to do with validation — that is the `--strict` flag on `claude plugin validate`, which treats warnings as errors. The two are unrelated mechanisms.
- `src/powershell/Tests/Unit/AgentPlugins.Tests.ps1:196` adds a dedicated test pinning the redundant field, locking the misreading into CI. Worth removing alongside it.

Verified locally: dropping the line keeps `claude plugin validate . --strict` passing.

**Suggestion:**

```suggestion
      "source": "./src/templates/agent-plugin",
```

src/templates/agent-plugin/README.md
flanakin: 🤖 [AI][Claude Code] ⚠️ **Should fix**

This claim is not fully delivered. The directory both manifests now point at still contains two committed symlinks:

```
$ git ls-files -s src/templates/agent-plugin | awk '$1=="120000"'
120000 84cf3d03…  src/templates/agent-plugin/skills/finops-toolkit/references/docs-mslearn
120000 8c5e6d61…  src/templates/agent-plugin/skills/finops-toolkit/references/queries
```

186 references across the plugin's own agents, commands, and `SKILL.md` resolve through them (`references/queries/INDEX.md`, `references/docs-mslearn/framework/finops-framework.md`, and so on). A marketplace install clones the repo and uses this directory in place — it does not run `Build-AgentPlugin.ps1`, which is what replaces those symlinks with real copies at package time. So on a Windows checkout without symlink support they materialize as one-line text files and every one of those references breaks: the same failure class this PR fixes, degrading skill content instead of failing the install.

`.plugin` at the repo root has the same shape — still a symlink, and `AgentPlugins.Tests.ps1:177` already carries a text-file fallback that acknowledges it.

Either narrow the claim to what is true (the marketplace **source path** no longer depends on symlink support) or make those two paths real directories so the sentence holds end to end.

src/powershell/Tests/Unit/AgentPlugins.Tests.ps1
```diff
@@ -164,17 +182,21 @@ Describe 'Plugin discovery and marketplaces' {
...
 
-        foreach ($marketplace in $marketplaces.Keys)
+        foreach ($marketplace in $marketplaces)
         {
             $json = Get-Content (Join-Path $script:RepoRoot $marketplace) -Raw | ConvertFrom-Json
             $entry = $json.plugins | Where-Object { $_.name -eq 'microsoft-finops-toolkit' }
-            $entry.source | Should -Be $marketplaces[$marketplace]
+            $entry.source | Should -Be './src/templates/agent-plugin'
         }
     }
```
flanakin: 🤖 [AI][Claude Code] 💡 **Suggestion**

The test name promises a cross-platform check, but the body only compares `source` to one literal string. The bug it guards against is *"`source` points at a symlink"* — and a future move to a different symlinked path would fail here only because the string changed, so the natural fix is to update the literal, and CI goes green again with a symlink source. It guards the answer, not the rule.

Asserting the property instead catches any symlinked source, and matches what the name already claims.

Verified: this passes as written, a negative control against `.plugin` (a real symlink) fails as expected, and the full file is still 24/24.

**Suggestion:**

```suggestion
    It 'Uses a cross-platform source path in both marketplaces' {
        foreach ($marketplace in @('.github/plugin/marketplace.json', '.claude-plugin/marketplace.json'))
        {
            $json = Get-Content (Join-Path $script:RepoRoot $marketplace) -Raw | ConvertFrom-Json
            $entry = $json.plugins | Where-Object { $_.name -eq 'microsoft-finops-toolkit' }
            $entry.source | Should -Match '^\./'

            $source = Get-Item (Join-Path $script:RepoRoot ($entry.source -replace '^\./', '')) -Force
            $source.LinkType | Should -BeNullOrEmpty
            Join-Path $source.FullName 'plugin.json' | Should -Exist
        }
    }
```

src/powershell/Tests/Unit/AgentPlugins.Tests.ps1
```diff
@@ -43,16 +43,26 @@ Describe 'Agent plugin manifest' {
...
         $root = Get-Content (Join-Path $script:Plugin 'plugin.json') -Raw | ConvertFrom-Json
         $claude = Get-Content (Join-Path $script:Plugin '.claude-plugin/plugin.json') -Raw | ConvertFrom-Json
 
         $root.mcpServers | Should -Be '.mcp.json'
         $claude.mcpServers | Should -BeNullOrEmpty
-        $claude.agents | Should -BeNullOrEmpty
         Join-Path $script:Plugin '.mcp.json' | Should -Exist
     }
 
+    It 'Declares resolvable paths for every Claude component' {
```
flanakin: 🤖 [AI][Claude Code] 💡 **Suggestion**

"every Claude component" overstates the scope — this checks `commands`, `skills`, and `outputStyles` only, deliberately skipping `agents` and `mcpServers` because this PR moved those to auto-discovery. Naming that explicitly keeps the intent readable for whoever next wonders why agents are missing here.

**Suggestion:**

```suggestion
    It 'Declares resolvable paths for every explicitly declared Claude component' {
```

flanakin (commented):

**Summary:** The core fix is right and verified — pointing both marketplace sources at a real directory resolves the Windows `ENOTDIR` install failure. All 24 Pester tests pass, `claude plugin validate --strict` is green on both manifests, every agent's front-matter `name:` matches its new basename, and no stale `.agent.md` or `plugins/microsoft-finops-toolkit` references remain anywhere in the repo. Four findings, none blocking.

### ⚠️ Should fix (2)

1. `"strict": true` is the documented default, so it is a no-op — and the PR describes it as "strict marketplace validation," conflating it with the unrelated `claude plugin validate --strict` flag. A new test pins the redundant field into CI.
2. The README claims the manifests "avoid relying on repository symlink support," but the directory both manifests now point at still ships two committed symlinks that 186 references across the plugin's own agents, commands, and `SKILL.md` depend on.

### 💡 Suggestions (2)

1. `Uses a cross-platform source path in both marketplaces` asserts a literal string rather than the property that actually broke, so it does not guard against a future symlinked source.
2. `Declares resolvable paths for every Claude component` covers only the explicitly declared components, not every one.

Changelog correctly skipped: v15 is unreleased (`package.json` is `15.0.0-dev.0`, latest tag is `v14`) and the plugin was added in this same cycle, so this fixes something that never shipped.

flanakin (approved):
LGTM

src/powershell/Tests/Unit/AgentPlugins.Tests.ps1
MSBrett: Strengthened the marketplace check to require a relative, non-symlink plugin directory that contains `plugin.json`, rather than pinning one literal source path.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

src/powershell/Tests/Unit/AgentPlugins.Tests.ps1
MSBrett: Thanks. The `NAME.md` assertion intentionally enforces the Claude-compatible convention adopted in this PR. Any intentional discovery reversal should update the implementation and its corresponding test together.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

src/templates/agent-plugin/README.md
MSBrett: Clarified the README: the marketplace manifests bypass only the repository-level `.plugin` discovery symlink. It no longer implies the selected plugin directory has no other symlink dependencies.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

.claude-plugin/marketplace.json
MSBrett: Removed the redundant explicit `strict` field and its dedicated test. Claude marketplace strict mode remains enabled through its documented default.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

src/powershell/Tests/Unit/AgentPlugins.Tests.ps1
MSBrett: Renamed the test to clarify that it validates explicitly declared Claude components. Auto-discovered agents and MCP servers are not included.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

PR #2291, opened 2026-08-26
fix(hubs): make ADF trigger startTime Z-suffix conditional on UTC fallback

### LINKED ISSUE #2157: [Bug] InvalidWorkflowTriggerRecurrence error during deployment — ADF trigger startTime missing zone designator
# GitHub Issue: ADF Schedule Triggers fail to start — missing UTC zone designator in startTime

## Title
`InvalidWorkflowTriggerRecurrence` error during deployment — ADF trigger `startTime` missing zone designator

## Repository
https://github.com/microsoft/finops-toolkit

---

### Describe the bug

FinOps Hub v14 fresh deployments fail at the `Microsoft.FinOpsHubs.Init` and `Microsoft.FinOpsHubs.StartTriggers` deployment script steps. The `Init-DataFactory.ps1` script cannot start the ADF schedule triggers because the `startTime` values are missing the UTC zone designator (`Z`).

### Error message

```
Microsoft.Rest.Azure.CloudException: HTTP Status Code: BadRequest
Error Code: BadRequest
Error Message: ErrorCode=InvalidWorkflowTriggerRecurrence, 
ErrorMessage=The recurrence of trigger has an invalid start time 
'2023-01-01T01:01:00.0000000' zone designator for time zone 'UTC'. 
Expected start time format with zone designator is 'yyyy-MM-ddTHH:mm:ssZ'.
```

### Affected triggers

| Trigger Name | Current startTime (broken) | Expected startTime |
|---|---|---|
| `config_DailySchedule` | `2023-01-01T01:01:00` | `2023-01-01T01:01:00Z` |
| `config_MonthlySchedule` | `2023-01-05T01:11:00` | `2023-01-05T01:11:00Z` |
| `queries_DailySchedule` | `2023-01-01T01:01:00` | `2023-01-01T01:01:00Z` |

### Steps to reproduce

1. Deploy FinOps Hub v14 using the `finops-hub-latest.json` ARM template (fresh deployment, not upgrade)
2. Wait for the deployment to reach the `Microsoft.FinOpsHubs.Init` step
3. The `Init-DataFactory.ps1` script attempts to start the triggers
4. Azure ADF rejects the trigger start due to invalid `startTime` format
5. The script retries 20 times, each time seeing "Trigger is Stopped, expected Started"
6. Deployment fails

### Expected behavior

The ADF schedule triggers should have `startTime` values with the UTC zone designator (`Z` suffix) so they can be started successfully during deployment.

### Environment

- **FinOps Toolkit version**: v14 (April 2026)
- **Template**: `finops-hub-latest.json`
- **Deployment type**: Fresh deployment (not upgrade)
- **Regions tested**: Sweden Central, Norway East
- **Deployment method**: Azure Portal ARM template deployment

### Suggested fix

In the Bicep template(s) that define the ADF schedule triggers, append `Z` to the `startTime` values:

```bicep
// Before (broken)
startTime: '2023-01-01T01:01:00'

// After (fixed)
startTime: '2023-01-01T01:01:00Z'
```

### Workaround

After deployment fails, manually update the trigger `startTime` values via Azure CLI:

```powershell
$rg = "<resource-group>"
$adf = "<data-factory-name>"

az datafactory trigger update --resource-group $rg --factory-name $adf `
    --name "config_DailySchedule" `
    --set "properties.recurrence.startTime=2023-01-01T01:01:00Z"

az datafactory trigger update --resource-group $rg --factory-name $adf `
    --name "config_MonthlySchedule" `
    --set "properties.recurrence.startTime=2023-01-05T01:11:00Z"

az datafactory trigger update --resource-group $rg --factory-name $adf `
    --name "queries_DailySchedule" `
    --set "properties.recurrence.startTime=2023-01-01T01:01:00Z"
```

Then start all triggers manually:

```powershell
$triggers = @("queries_DailySchedule", "msexports_ManifestAdded", "config_SettingsUpdated", 
              "config_DailySchedule", "config_MonthlySchedule", "ingestion_ManifestAdded")

foreach ($t in $triggers) {
    az datafactory trigger start --resource-group $rg --factory-name $adf --name $t -o none
}
```

### Additional context

- Reproduced independently on two separate tenants/subscriptions with fresh v14 deployments
- The v14 release notes mention fixing "US Government cloud time zone deployment failures" — this may be a related regression
- Upgrades from v13 may not hit this issue if triggers were already started with valid `startTime` values from a previous version

### Labels
`bug`, `hubs`

flanakin:
## Summary

Fixes #2157 — FinOps hub deployments fail in regions not covered by `timeZones.bicep`'s region→Windows-timezone map (e.g. Sweden Central) with `InvalidWorkflowTriggerRecurrence`.

**Root cause:** Three ADF schedule triggers hardcode `startTime` without a trailing `Z` (UTC zone designator):
- `config_DailySchedule` and `queries_DailySchedule`: `'2023-01-01T01:01:00'`
- `config_MonthlySchedule`: `'2023-01-05T01:11:00'`

These triggers set `timeZone` from `timeZones.bicep`'s region lookup, which falls back to the literal `'UTC'` for any region not in its explicit map (confirmed: Sweden Central is one such region; likely also Qatar Central, Poland Central, Italy North, Israel Central, though the map should be checked directly rather than assuming this list is exhaustive). Azure Data Factory requires `startTime` to end in `Z` (format `yyyy-MM-ddTHH:mm:ssZ`) whenever `timeZone` resolves to `'UTC'` — otherwise trigger activation fails with `InvalidWorkflowTriggerRecurrence`, which surfaces during deployment as `Init-DataFactory.ps1`'s `Start-AzDataFactoryV2Trigger` retrying up to `MAX_RETRIES = 20` and then failing the whole deployment.

**Fix:** Make the `Z` suffix conditional on the resolved `timeZone` actually being `'UTC'`:

```bicep
startTime: timeZones.outputs.Timezone == 'UTC' ? '2023-01-01T01:01:00Z' : '2023-01-01T01:01:00'
```

Applied to all three triggers:
- `src/templates/finops-hub/modules/Microsoft.CostManagement/ManagedExports/app.bicep` — `config_DailySchedule` and `config_MonthlySchedule`
- `src/templates/finops-hub/modules/Microsoft.FinOpsHubs/IngestionQueries/app.bicep` — `queries_DailySchedule`

This preserves the existing local-wall-clock scheduling behavior for every region that **is** mapped (no `Z`, non-UTC `timeZone`), and only adds `Z` for the fallback case, fixing activation there.

**Explicitly not done, and why:**
- Did not unconditionally append `Z` to all three literals — that would silently switch every *mapped* region from local-time scheduling to UTC wall-clock scheduling, a user-facing behavior change for currently-working regions.
- Did not expand the region→timezone map to reduce how often the UTC fallback triggers — that's a separate, non-durable maintenance treadmill, not a fix for this defect.

**Out of scope:** The issue reporter also mentioned a failure in Norway East. Norway East **is** present in `timeZones.bicep`'s map (`'W. Europe Standard Time'`, non-UTC), so the root cause above does not explain a Norway East failure. This looks like a second, distinct issue that needs more diagnostic detail from the reporter — it is not addressed by this PR.

## Test plan

- [x] `bicep build` on both changed files completes cleanly (only the expected `@validate()` experimental-feature warning, unrelated to this change)
- [x] Confirmed compiled ARM JSON emits the conditional `if(equals(...Timezone..., 'UTC'), '...Z', '...')` expression for all three `startTime` properties
- [x] `pwsh -Command "./src/scripts/Test-PowerShell.ps1 -Lint"` — 3585/3585 tests passed
- [x] Manual validation: deploy to an unmapped region (e.g. Sweden Central) and confirm triggers activate without `InvalidWorkflowTriggerRecurrence`

---

> - [x] 🤏 Lint tests
> - [x] 🤞 PS -WhatIf / az validate
> - [ ] 💪 Unit tests
> - [ ] 🙌 Integration tests

### 📦 Deploy to test?

> - [x] Hubs + ADX (managed)
> - [x] Hubs (manual)
> - [ ] Hubs (no data)
> - [ ] Workbooks
> - [ ] Alerts

⚛️ [Catalyst] Added the release checklist section and opted into test deployments so the trigger activation can be verified before v15 ships.


src/templates/finops-hub/modules/Microsoft.CostManagement/ManagedExports/app.bicep
```diff
@@ -112,7 +112,11 @@ resource dataFactory 'Microsoft.DataFactory/factories@2018-06-01' existing = {
         recurrence: {
           frequency: 'Hour'
           interval: 24
-          startTime: '2023-01-01T01:01:00'
+          // Data Factory requires a trailing 'Z' on startTime when timeZone resolves to 'UTC' (the
+          // fallback for regions not in timeZones.bicep's map); a missing 'Z' fails trigger activation
+          // with InvalidWorkflowTriggerRecurrence. Mapped, non-UTC regions must keep the no-'Z' format
+          // so they continue scheduling on local wall-clock time.
+          startTime: timeZones.outputs.Timezone == 'UTC' ? '2023-01-01T01:01:00Z' : '2023-01-01T01:01:00'
```
RolandKrummenacher, line 119: The conditional and the comment are both right, and I like that the second occurrence points back here rather than repeating the rationale.

Non-blocking suggestion: pin this in `HubsAdfTriggerTimeZones.Tests.ps1`. Its `Fallback` context already asserts the `?? 'UTC'` fallback against the Bicep source, so the natural companion is an assertion that every `startTime` in these two files carries the `Z` on the UTC branch and omits it on the other — something to the effect of every `startTime:` line matching `Timezone == 'UTC' ? '<ts>Z' : '<ts>'`.

Without it, the invariant lives only in a comment, and `91ce5a5e` is the precedent for extending this specific test file so the fix cannot silently regress. If you extend it to the `docs/deploy` artifacts the way that commit did for the mappings, it also catches the shipped-template drift I raised in the review body.

RolandKrummenacher (changes requested):
The Bicep change is correct and I would approve it as written — the reasoning in the PR description holds up under checking:

- `timeZones.bicep:63` falls back to `'UTC'` via `timezoneobject[?loc] ?? 'UTC'`, and `swedencentral` is genuinely absent from the map, so the reported region does land on the fallback.
- The ternary keeps the no-`Z` form for all 36 mapped, non-UTC regions, so no currently-working region changes scheduling behavior. ARM `equals()` is case-sensitive and no map value is `'UTC'` in any casing, so the condition can only match the fallback.
- All three `startTime` occurrences under `src/templates/` are covered — there are no others.
- Declining to unconditionally append `Z`, and declining to grow the region map, are both the right calls for this PR.

I did not independently run `bicep build`, but the change is a plain ternary over two string literals in a property that already references `timeZones.outputs.Timezone`, so I'm not worried about the compiled output.

**What blocks: the shipped deploy templates still reproduce #2157.**

`docs/deploy/*.json` are regenerated only at release time, so on the `flanakin/v15-prep` base as it stands today:

| artifact | `startTime` | `swedencentral` | fallback |
|---|---|---|---|
| `finops-hub-latest.json` | no `Z` (3x) | absent | `'UTC'` |
| `finops-hub-14.0.json` | no `Z` (3x) | absent | `'UTC'` |
| `finops-hub-preview.json` | no `Z` (2x) | absent | `'Universal Coordinated Time'` |

Both `latest` and `14.0` already emit the `'UTC'` fallback and have no `swedencentral` entry, and their `startTime` values have no `Z` — which is exactly the failing combination. So after this merges, anyone hitting the **Deploy to Azure** button in Sweden Central still gets `InvalidWorkflowTriggerRecurrence`, and the issue this PR closes remains reproducible through the primary deployment path until v15 ships.

That is the same situation as `91ce5a5e`, which patched these artifacts surgically for this defect class in these same files, for the same triggers, on the grounds that they lag the src fixes. I would rather not close #2157 with the button still broken — either patch the three literals in `latest` and `14.0` here, or leave #2157 open with a note that the shipped-artifact half is outstanding.

**`finops-hub-preview.json` has a second problem, pre-existing on the base and not caused by this PR:** its fallback is still the display name `'Universal Coordinated Time'`, which ADF rejects outright — `91ce5a5e` corrected the *mapping values* in that file but not the fallback. The `Fallback` context in `HubsAdfTriggerTimeZones.Tests.ps1:139` only asserts against `$bicepContent`, so this passes CI. Since v15-prep users deploy from the preview artifact, and it is the same fallback path this PR is fixing, it seems in scope to fix here — your call.

**Non-blocking:**

- No changelog entry under `Unreleased > FinOps hubs > Fixed`. The recent fix commits on `dev` each add one.
- No regression test pins the startTime/`Z` invariant — there are zero `startTime` hits across `src/powershell/Tests` and `src/scripts`, so a template regeneration or refactor can silently drop this with CI green. `91ce5a5e` extended `HubsAdfTriggerTimeZones.Tests.ps1` for exactly this reason; a case in its `Fallback` context would cover it. See the inline comment.

**On Norway East:** your out-of-scope reasoning is right. `norwayeast` maps to `'W. Europe Standard Time'` in the src Bicep, in `finops-hub-latest.json`, and in `finops-hub-14.0.json`, so this root cause does not explain the reporter's Norway East error, which named time zone `'UTC'`. Agreed that needs separate diagnostics.

src/templates/finops-hub/modules/Microsoft.CostManagement/ManagedExports/app.bicep
flanakin, line 119: nit: Should we update these to 2026? Not sure if there's an issue with a 3-year-old start date. Just a thought. Applies to the other 2 instances as well.

src/templates/finops-hub/modules/Microsoft.CostManagement/ManagedExports/app.bicep
flanakin, line 1: Holding for validation

src/templates/finops-hub/modules/Microsoft.CostManagement/ManagedExports/app.bicep
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Added a `startTime UTC suffix` context to `HubsAdfTriggerTimeZones.Tests.ps1` asserting every `startTime:` line in the two src files matches `Timezone == 'UTC' ? '<ts>Z' : '<ts>'` (same timestamp on both branches, differing only by the trailing `Z`), plus the compiled `if(equals(...))` form in `finops-hub-14.0.json`. See the PR-level reply for why `finops-hub-latest.json`/`finops-hub-preview.json` are intentionally excluded from that assertion.

flanakin: 🤖 [AI][Claude Code] ✅ **Answered**

Confirmed via the Data Factory schedule trigger docs: ADF walks forward from `startTime` by the recurrence interval regardless of how old it is — only the anchor hour/minute (and `monthDays` for the monthly trigger) matter, not the year. There's no documented limit, quota, or deprecation tied to a stale `startTime` (the only "can't be in the past" restriction applies to `endTime`, which these triggers don't set). See https://learn.microsoft.com/azure/data-factory/how-to-create-schedule-trigger#starttime-property.

Recommend leaving 2023 as-is — bumping to 2026 would be pure churn with zero functional effect.

src/templates/finops-hub/modules/Microsoft.CostManagement/ManagedExports/app.bicep
flanakin: 🤖 [AI][Claude Code] 💬 **Status**

CI's automated PR deployments (Hubs manual, Hubs + ADX managed) succeeded, but neither targets an unmapped region like Sweden Central, so they don't exercise the UTC-fallback path this fix targets. Manual validation in an unmapped region is still the open item to fully close this out.

flanakin:
🤖 [AI][Claude Code] ✅ **Implemented**

Addressed the review:

- **Shipped deploy templates**: Patched the 3 `startTime` occurrences in `finops-hub-14.0.json` to the same UTC-conditional `Z` suffix (compiled to `if(equals(reference('timeZones').outputs.Timezone.value, 'UTC'), '...Z', '...')`), matching the precedent in 91ce5a5e. Intentionally **not** touching `finops-hub-latest.json` — per the PR author, it gets fully regenerated when v15 ships, so patching it now is unnecessary busywork.
- **`finops-hub-preview.json` fallback bug**: Left as-is for the same reason — it's a generated v15-prep artifact that will be replaced by the release regen, so it's not worth patching in isolation here.
- **Changelog**: Added a Fixed entry under `Unreleased > FinOps hubs` referencing #2157/#2291.
- **Regression test**: Extended `HubsAdfTriggerTimeZones.Tests.ps1` with a `startTime UTC suffix` context pinning the Z/no-Z pairing in both src files and in the now-patched `finops-hub-14.0.json`. `finops-hub-latest.json`/`finops-hub-preview.json` are intentionally excluded from that assertion since they still ship the pre-fix, unconditional `startTime` by design.

All 2283 PowerShell unit tests pass, and `bicep build` on both changed src files is clean.

flanakin:
🤖 [AI][Claude Code] **PR Update Summary**

**Addressed:** 3 thread(s) + 1 review body

- ✅ Implemented: 2 (deploy artifact patch + regression test; changelog entry)
- 💬 Answered: 2 (startTime date question; validation status)

**Key changes:**
- Patched `finops-hub-14.0.json`'s 3 `startTime` literals with the same UTC-conditional `Z` suffix as the src fix (compiled ARM `if(equals(...))` form) — `finops-hub-latest.json`/`finops-hub-preview.json` intentionally left alone since both get regenerated at v15 release.
- Added a changelog entry under `Unreleased > FinOps hubs > Fixed`.
- Extended `HubsAdfTriggerTimeZones.Tests.ps1` with a new `startTime UTC suffix` context (6 test cases) pinning the Z/no-Z pairing so it can't silently regress.
- Confirmed via Microsoft Learn docs that ADF's `startTime` age has no functional effect — recommended keeping the existing 2023 dates.

Full unit suite: 2283/2283 passing. `bicep build` clean on both changed src files.

docs/deploy/finops-hub-14.0.json
```diff
@@ -14033,7 +14033,7 @@
                           "recurrence": {
                             "frequency": "Hour",
                             "interval": 24,
-                            "startTime": "2023-01-01T01:01:00",
+                            "startTime": "[if(equals(reference('timeZones').outputs.Timezone.value, 'UTC'), '2023-01-01T01:01:00Z', '2023-01-01T01:01:00')]",
```
RolandKrummenacher, line 14036: Patching this artifact is the right call, and I'm withdrawing my objection to leaving `finops-hub-latest.json` alone.

I checked how the deploy path actually resolves rather than assuming: `aka.ms/finops/hubs/deploy` does point at `finops-hub-latest.json`, but GitHub Pages serves `main`/`docs`, not `dev` — so nothing merged here moves the live button either way, and `latest.json` picks up the src fix when it's regenerated at release.

The distinction you drew is the correct one, and it's a cleaner rule than what `91ce5a5e` did:

- `finops-hub-14.0.json` is a frozen snapshot of a shipped release. It is never regenerated, so a manual patch is the only fix it will ever get. ✅
- `finops-hub-latest.json` is regenerated every release from src, which now carries the fix.

(`91ce5a5e` patched all three because it was correcting invalid *mapping values* — wrong on their face in any artifact, regenerated or not. This is a src fix that reaches `latest.json` through the normal path, so it isn't the same case.)

src/powershell/Tests/Unit/HubsAdfTriggerTimeZones.Tests.ps1
```diff
@@ -142,4 +184,22 @@ Describe 'HubsAdfTriggerTimeZones' {
             $bicepContent | Should -Match "\?\?\s*'UTC'"
         }
     }
+
+    Context 'startTime UTC suffix' {
```
RolandKrummenacher, line 188: This is exactly what I was after — the Z/no-Z pairing is now pinned rather than living only in a comment, so a template regen or refactor can't drop it with CI green.

Asserting `$Matches[1] | Should -BeExactly $Matches[2]` is a nice touch: it catches a timestamp drifting between the two branches, which a looser "does it end in Z" check would miss.

src/powershell/Tests/Unit/HubsAdfTriggerTimeZones.Tests.ps1
```diff
@@ -142,4 +184,22 @@ Describe 'HubsAdfTriggerTimeZones' {
             $bicepContent | Should -Match "\?\?\s*'UTC'"
```
RolandKrummenacher, line 184: Not blocking this PR — flagging so it doesn't get lost.

This `Fallback` context asserts only against `$bicepContent`, so it verifies the src fallback and nothing else. `finops-hub-preview.json` still ships `'Universal Coordinated Time'` as its fallback — a display name ADF rejects outright, not a Windows time zone ID — and no test covers it.

That's pre-existing and genuinely out of scope here, but it means the gap is invisible to CI rather than merely unfixed. I'll open a separate issue for it.

docs-mslearn/toolkit/changelog.md
```diff
@@ -34,6 +34,7 @@ The following section lists features and enhancements that are currently in deve
   - Replaced redundant `tolower()` comparisons in hub KQL with case-insensitive operators (`has`, `=~`, `!~`) so the engine can use the term index instead of scanning every row ([#2213](https://github.com/microsoft/finops-toolkit/issues/2213)).
   - Replaced whole-term `contains` matches with `has` across hub KQL and the query catalog (resource ID paths, licensing phrases, SKU description terms) and added a per-row operator-equivalence regression harness with unit test coverage ([#2220](https://github.com/microsoft/finops-toolkit/pull/2220)).
 - **Fixed**
+  - Fixed ADF schedule trigger deployment failures (`InvalidWorkflowTriggerRecurrence`) in regions not covered by the region-to-time-zone map (for example, Sweden Central), where the trigger's `startTime` was missing the required `Z` suffix for the `UTC` fallback time zone ([#2157](https://github.com/microsoft/finops-toolkit/issues/2157), [#2291](https://github.com/microsoft/finops-toolkit/pull/2291)).
```
RolandKrummenacher, line 37: Changelog entry confirmed — right section, and it references both the issue and the PR consistently with the surrounding entries.

RolandKrummenacher (approved):
Approving — all three substantive items from my review landed, and I'm withdrawing the `finops-hub-latest.json` objection (reasoning inline on the `14.0.json` patch).

Two things to carry forward, neither blocking:

- **The branch conflicts with `dev`** and isn't mergeable as-is.
- **`finops-hub-preview.json`'s `'Universal Coordinated Time'` fallback** is still outstanding — pre-existing and correctly out of scope here, but nothing in CI catches it. Details inline on the `Fallback` context. I'll file it separately so it doesn't get lost in a resolved thread.

Manual validation in an unmapped region is still unchecked on the test plan. I don't think it needs to gate the merge — the change is a ternary over two string literals and the compiled ARM form is correct in all three places — but it's the one thing that would fully confirm #2157 is closed.

src/templates/finops-hub/modules/Microsoft.CostManagement/ManagedExports/app.bicep
flanakin: 🤖 [AI][Claude Code] ✅ **Validated**

Deployed a storage-only hub with managed exports + recommendations enabled to `swedencentral` (unmapped region) via `Deploy-Hub.ps1`, using this branch's fix. Confirmed `timeZone` resolved to the `UTC` fallback as expected, and all three affected triggers activated successfully:

```
queries_DailySchedule    RuntimeState: Started   startTime=2023-01-01 01:01:00  timeZone=UTC
config_DailySchedule     RuntimeState: Started   startTime=2023-01-01 01:01:00  timeZone=UTC
config_MonthlySchedule   RuntimeState: Started   startTime=2023-01-05 01:11:00  timeZone=UTC
```

No `InvalidWorkflowTriggerRecurrence` — this is the exact failure mode #2157 reported, and it's fixed. Cleaned up the validation resource group afterward.

PR #2284, opened 2026-08-26
fix(open-data,hubs): align commitment discount eligibility with FOCUS spend and usage categories

RolandKrummenacher:
## 🛠️ Description

Fixes #2279.

`x_CommitmentDiscountSpendEligibility` reported **reservation** pricing and `x_CommitmentDiscountUsageEligibility` reported **savings plan** pricing. That is the reverse of the FOCUS `CommitmentDiscountCategory` definition the toolkit uses everywhere else: a reservation is a **Usage** commitment (you commit to a quantity of usage) and a savings plan is a **Spend** commitment (you commit to an amount of money).

`IngestionSetup_v1_2.kql` contained both mappings ten lines apart — it assigned `CommitmentDiscountCategory` correctly, then assigned the eligibility columns the opposite way, under a `// TODO: Check this to ensure it's correct` comment. That TODO is now resolved.

### What changed

| Component | Change |
|---|---|
| `src/scripts/Update-CommitmentDiscountEligibility.ps1` | Mapping extracted into a new `New-EligibilityRow` helper and corrected. The change-summary key is now built from the row's own properties, so it can't drift from the CSV column order. |
| `src/open-data/CommitmentDiscountEligibility.csv` | The two columns are swapped for all 92,624 meters. 36,287 rows change value; the 56,337 meters eligible for both are unaffected. |
| `IngestionSetup_v1_0.kql`, `IngestionSetup_v1_2.kql` | The spend column now reads `spMeters`, the usage column reads `riMeters`. |
| `docs-mslearn/toolkit/open-data.md` | Corrects the two column descriptions and the sample rows, and states the FOCUS reasoning. |
| `src/powershell/Tests/Unit/Update-CommitmentDiscountEligibility.Tests.ps1` | Five new tests pin the direction and the published column order. |

The exclusion of reservation price rows from their own eligibility (`and x_SkuPriceType != 'ReservedInstance'`) travels with `riMeters` to the usage column: it describes what an on-demand charge could be covered by, not the commitment row itself.

The best practice query in `docs-mslearn/best-practices/compute.md` already used the correct mapping and is unchanged. `data-dictionary.md` and the Copilot Studio `schema-reference.md` also already described the columns per FOCUS, and now match the data.

## 🧪 Verification

[...]

PR #2298, opened 2026-09-02
Fix opendata-ci auto-commit gap and add ISF/eligibility PowerShell commands

flanakin:
## 🛠️ Description

`opendata-ci.yml` used `git commit -a` after regenerating PowerShell functions from `src/open-data/*.csv`. `git commit -a` only stages **modified/deleted tracked files** — it silently skips new untracked files. `InstanceSizeFlexibility.csv` (#2199) and `CommitmentDiscountEligibility.csv` (#2251) never had their generated `Get-OpenData*.ps1` functions committed, so they stayed untracked forever and every PR touching either CSV (most recently #2295) saw a "no diff" CI result that masked the standing gap. Fixed by switching to `git add -A` before `git commit`.

Also adds `Get-FinOpsInstanceSizeFlexibility` and `Get-FinOpsCommitmentDiscountEligibility` public cmdlets, matching the existing `Get-FinOpsPricingUnit`/`Get-FinOpsService` pattern — these two datasets had no PowerShell-facing lookup at all, only direct CSV consumption (Power BI, Optimization Engine).

### Bugs found and fixed along the way

- `Get-FinOpsCommitmentDiscountEligibility`'s initial draft used the same `Select-Object -Property * -Unique` post-filter as the other `Get-FinOps*` cmdlets. At 92k+ rows this hung for minutes. `MeterId` is already unique in the source data, so `-Unique` was dropped (both from the cmdlet and its test's `BeforeAll`).
- `Build-OpenData.ps1`'s `Write-Command`/`Out-File` generation path silently dropped rows on large datasets on some runs (662 rows missing from one `CommitmentDiscountEligibility` generation, 1 row from an `InstanceSizeFlexibility` generation). Regenerating fixed it both times. Not root-caused further — flagging here as a latent reliability risk in the generator for anyone touching it next.
- Merging in `dev` picked up #2284, which swapped which eligibility column maps to reservations vs. savings plans. Corrected the new cmdlet's help text/docs to match and regenerated the private function against the updated CSV.

Fixes # (opendata-ci not running as expected on PRs like #2295)

## 📷 Screenshots

N/A — PowerShell module change

> - [x] 🤏 Lint tests
> - [x] 💪 Unit tests
> - [ ] 🙌 Integration tests

### 📦 Deploy to test?

> - [ ] Hubs + ADX (managed)
> - [ ] Hubs (manual)
> - [ ] Hubs (no data)
> - [ ] Workbooks
> - [ ] Alerts

> - [x] ✅ Public docs in `docs-mslearn` (required for `dev`)

🤖 [AI] Generated with [Claude Code](https://claude.com/claude-code)


src/powershell/Public/Get-FinOpsCommitmentDiscountEligibility.ps1
```diff
@@ -0,0 +1,61 @@
...
+
+<#
+    .SYNOPSIS
+    Gets commitment discount eligibility for a meter.
+
+    .PARAMETER MeterId
+    Optional. The meter ID to look up. Accepts wildcards. Default = * (all).
+
+    .PARAMETER SpendEligibility
+    Optional. Filters to meters with the specified savings plan (spend commitment) eligibility. FOCUS classifies a savings plan as a spend commitment because you commit to an amount of money. Expected values: Eligible, Not Eligible. Default = null (all).
```
RolandKrummenacher, line 12: `Default = null (all)` doesn't match the implementation — the default is `"*"` (line 41), and passing `$null` explicitly doesn't mean "all": `[string] $null` coerces to `""`, and `"Eligible" -like ""` is `$false`, so `-SpendEligibility $null` returns **0 rows** rather than everything. Verified against the branch.

`Get-FinOpsPricingUnit:58` guards its equivalent case (`$null -eq $BlockSize -or ...`); this cmdlet doesn't. Either fix the help text to say `Default = * (all)` — matching how `MeterId` is documented two lines up — or add the null guard. Same wording on line 15 for `UsageEligibility`, and both are repeated in `get-finopscommitmentdiscounteligibility.md`.

src/powershell/Public/Get-FinOpsInstanceSizeFlexibility.ps1
```diff
@@ -0,0 +1,55 @@
...
+            -and $_.InstanceSizeFlexibilityGroup -like $InstanceSizeFlexibilityGroup
+    } `
+    | ForEach-Object {
+        [PSCustomObject]@{
+            InstanceSizeFlexibilityGroup = $_.InstanceSizeFlexibilityGroup
+            ArmSkuName                   = $_.ArmSkuName
+            Ratio                        = $_.Ratio
+        }
+    } `
+    | Select-Object -Property * -Unique
```
RolandKrummenacher, line 54: This is a no-op: the source has 1,353 rows and 1,353 distinct `(InstanceSizeFlexibilityGroup, ArmSkuName, Ratio)` triples — `ArmSkuName` alone is already unique — so nothing is ever deduplicated. It costs roughly 0.8s of the cmdlet's ~1.1s runtime.

Same reasoning you used to drop `-Unique` from the sibling cmdlet — worth dropping here too, or adding the equivalent "already unique per row" comment if you'd rather keep it defensively.

.github/workflows/opendata-ci.yml
```diff
@@ -42,7 +42,8 @@ jobs:
         {
           git config --global user.name "${{ env.CI_COMMIT_AUTHOR }}"
           git config --global user.email "finopstoolkit@users.noreply.github.com"
-          git commit -a -m "${{ env.CI_COMMIT_MESSAGE }}"
+          git add -A
+          git commit -m "${{ env.CI_COMMIT_MESSAGE }}"
```
RolandKrummenacher, line 46: Confirmed this is the actual gap: `git commit -a` stages only modifications to already-tracked files, so newly generated open-data files were never committed. Worth noting the failure was silent because a failing `git commit` is followed by `git push`, which exits 0 — the job stayed green while doing nothing. Might be worth a `set -e` or an explicit check later, but not in this PR.

RolandKrummenacher (changes requested):
Thanks for splitting the CI fix out — `git add -A` is the right call, and I confirmed the old `git commit -a` genuinely skipped newly generated (untracked) files. The generated data itself checks out: both private functions match their CSVs exactly (92,624 and 1,353 rows, field for field), so the "generator dropped rows" hazard isn't present in this output. Eligibility semantics line up with the post-#2284 orientation, and the 6 new tests pass locally under Pester 6.

Requesting changes on one thing that's much cheaper to decide now than after this ships, plus a small correctness bug (inline).

**The 16.2 MB generated file is the one I'd like to settle before merge.** GitHub won't let me anchor a comment there (no diff entry for a file that size), so it goes here.

`src/powershell/Private/Get-OpenDataCommitmentDiscountEligibility.ps1` is 16.2 MB / 92,633 lines, and `FinOpsToolkit.psm1:7-17` dot-sources every `Private/*.ps1` eagerly on import. Measured cold import: **0.72s without this file, 3.6-4.8s with it** — every user of every cmdlet pays that, and `Build-PsModule.ps1:83` ships the whole Private folder to PSGallery.

Second cost: `Tests/Lint/Lint.Tests.ps1:43-44` runs `Invoke-ScriptAnalyzer -IncludeRule $_` once **per rule** per file. I measured 10.9s for a single rule on this file, against 1.57s for the previously largest generated file. At ~75 rules that's roughly **+14 minutes of lint on every `src/powershell/**` PR** locally, and more on the Windows CI runner.

Embedding generated data as a function is the established pattern here, so this isn't a wrong approach — it's that this file is about 10x larger than anything we've embedded before, while the source CSV is only 5.5 MB. `Import-Csv` on demand, or lazy-loading behind a script-scoped cache, would avoid both costs. Happy to be overruled if the data has to be in-module, but I'd rather make that call now than after v15 is on the gallery.

Two more that can't be anchored because the files aren't in this diff:

**`src/templates/agent-plugin/skills/finops-toolkit/SKILL.md`** — the "Open data commands" table still lists only the five existing references. The plugin bundles the two new reference docs but never points at them, and no test enforces that table, so it will stay stale silently.

**`docs/open-data.md:40`** — the Jekyll instance size flexibility tile keeps its hidden placeholder button (`visibility:hidden`) where every other dataset tile has a "See PowerShell" link. The docs-mslearn twin was updated in this PR, so the two sites now disagree on whether a cmdlet exists.

RolandKrummenacher:
Heads-up on an interaction with #2308, which grows `src/open-data/InstanceSizeFlexibility.csv` from 1,353 to 2,573 rows (restoring retired SKUs the Catalogs API no longer returns).

`Get-OpenDataInstanceSizeFlexibility.ps1` in this PR is generated from that CSV at its current 1,353 rows, so the merge order matters:

- **This PR first** — the file becomes tracked on `dev`, and I regenerate it on #2308 before merging. No action for you.
- **#2308 first** — this branch would carry a generated file built from the old CSV. Open Data CI would not catch it either, because its `paths: src/open-data/*.csv` filter does not match this PR.

So merging this one first is the cleaner order, and I will handle the regeneration on my side.

Worth noting for the auto-commit fix itself: Open Data CI ran green on #2308 (head `77949cd8`) and generated nothing, because `git commit -a` skipped the still-untracked ISF file — exactly the gap `git add -A` closes here.

RolandKrummenacher:
Update on the interaction I flagged earlier: still there, but much smaller than it was.

#2308 no longer backfills retired SKUs — that part was dropped. It now only widens the region sweep, so the CSV goes from 1,353 to **1,363** rows rather than 2,573. `Get-OpenDataInstanceSizeFlexibility.ps1` here is generated at 1,353, so it would still be stale by 10 rows.

Same mechanism as before, so merging this one first remains the cleaner order and I will regenerate on my side afterwards. If #2308 lands first instead, note that Open Data CI would not catch it: its `paths: src/open-data/*.csv` filter does not match this PR.

docs-mslearn/toolkit/powershell/data/get-finopscommitmentdiscounteligibility.md
```diff
@@ -0,0 +1,94 @@
...
+| ---------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| MeterId          | Optional. The meter ID to look up. Accepts wildcards. Default = \* (all).                                                                                                                                                                                  |
+| SpendEligibility | Optional. Filters to meters with the specified savings plan (spend commitment) eligibility. FOCUS classifies a savings plan as a spend commitment because you commit to an amount of money. Expected values: Eligible, Not Eligible. Default = null (all). |
+| UsageEligibility | Optional. Filters to meters with the specified reservation (usage commitment) eligibility. FOCUS classifies a reservation as a usage commitment because you commit to a quantity of usage. Expected values: Eligible, Not Eligible. Default = null (all).  |
```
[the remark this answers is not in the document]
flanakin: 🤖 [AI] ✅ **Implemented**

Updated the help text to `Default = * (all)` to match the actual parameter default and behavior.

docs-mslearn/toolkit/powershell/data/open-data-commands.md
```diff
@@ -19,6 +19,8 @@ The FinOps toolkit open data PowerShell commands help you integrate datasets int
 
 ## Commands
 
+- [Get-FinOpsCommitmentDiscountEligibility](Get-FinOpsCommitmentDiscountEligibility.md) – Gets commitment discount eligibility for a meter.
+- [Get-FinOpsInstanceSizeFlexibility](Get-FinOpsInstanceSizeFlexibility.md) – Gets the instance size flexibility group and ratio for an ARM SKU.
 - [Get-FinOpsPricingUnit](Get-FinOpsPricingUnit.md) – Gets an Azure region ID and name.
```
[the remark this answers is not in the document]
flanakin: 🤖 [AI] ✅ **Implemented**

Fixed the copy-pasted description in both command list docs.

docs-mslearn/toolkit/powershell/powershell-commands.md
```diff
@@ -70,6 +70,8 @@ The FinOps toolkit PowerShell module includes commands to manage FinOps solution
 
 ### Open data commands
 
+- [Get-FinOpsCommitmentDiscountEligibility](data/Get-FinOpsCommitmentDiscountEligibility.md) – Gets commitment discount eligibility for a meter.
+- [Get-FinOpsInstanceSizeFlexibility](data/Get-FinOpsInstanceSizeFlexibility.md) – Gets the instance size flexibility group and ratio for an ARM SKU.
 - [Get-FinOpsPricingUnit](data/Get-FinOpsPricingUnit.md) – Gets an Azure region ID and name.
```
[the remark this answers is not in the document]
flanakin: 🤖 [AI] ✅ **Implemented**

Fixed the copy-pasted description in both command list docs.

src/powershell/Public/Get-FinOpsCommitmentDiscountEligibility.ps1
```diff
@@ -0,0 +1,61 @@
...
+
```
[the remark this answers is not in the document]
flanakin: 🤖 [AI] ✅ **Implemented**

Updated the help text to `Default = * (all)` to match the actual parameter default and behavior.

src/powershell/Public/Get-FinOpsCommitmentDiscountEligibility.ps1
flanakin: 🤖 [AI] ✅ **Implemented**

Good catch — this was a real bug, not just a docs mismatch. Fixed both: help text now says `Default = * (all)`, and added a guard so an explicit `$null` (which binds to `""` for a `[string]` parameter) is normalized to `"*"` before filtering, so it now returns all rows instead of 0. Added a regression test for the explicit-`$null` case.

src/powershell/Public/Get-FinOpsInstanceSizeFlexibility.ps1
flanakin: 🤖 [AI] ✅ **Implemented**

Dropped `-Unique` here too and added the same "already unique per row" comment used on the sibling cmdlet, in both the function and its test's `BeforeAll`.

.github/workflows/opendata-ci.yml
flanakin: 🤖 [AI] 👍 Acknowledged — thanks for confirming the root cause. Agreed a `set -e` or explicit exit-code check would help catch this class of failure earlier; leaving that for a separate follow-up as you suggested.

flanakin:
🤖 [AI] Addressing the review-body items that couldn't be anchored to a diff line:

**`SKILL.md` "Open data commands" table** — ✅ **Implemented**. Added the two missing rows (`get-finopscommitmentdiscounteligibility.md`, `get-finopsinstancesizeflexibility.md`); both docs were already bundled under `references/`, just unlisted.

**`docs/open-data.md:40` (ISF tile missing "See PowerShell")** — ✅ **Implemented**. Added the link, matching the pattern used by the Pricing units / Regions / Services tiles.

**16.2 MB `Get-OpenDataCommitmentDiscountEligibility.ps1` (import time + lint cost)** — 🤔 **Needs discussion, but agreed and planned**. You're right that this isn't a wrong approach, just wrong at this scale — we don't want to lazy-load just this one dataset and leave the other five inconsistent. Rather than block this PR on a design-relayout, I've scoped a follow-up PR that reworks all six `Get-OpenData*` functions (and `Build-OpenData.ps1`, the generator) to lazy-load via `Import-Csv` with a script-scoped cache, replacing the embedded-literal pattern everywhere at once. Will land as its own PR referencing this thread.

@RolandKrummenacher — also thanks for the heads-up on the #2308 merge-order interaction; agreed this PR should land first and appreciate you handling the regeneration on your side.

flanakin:
🤖 [AI] **PR Update Summary**

**Addressed:** 8 thread(s)/comment(s)

- ✅ Implemented: 6
- 🤔 Needs discussion: 1 (16.2 MB generated file — agreed, scoped as a separate follow-up PR reworking all six Get-OpenData* functions + the generator to lazy-load)
- 👍 Acknowledged: 1 (silent-failure root cause confirmation)

**Key changes:** Fixed the SpendEligibility/UsageEligibility explicit-$null bug (was silently returning 0 rows), corrected copy-pasted/stale docs (Get-FinOpsPricingUnit description, eligibility parameter defaults, SKILL.md table, ISF tile link), and dropped a no-op -Unique call from Get-FinOpsInstanceSizeFlexibility. All 10 relevant Pester tests pass; PSScriptAnalyzer clean on changed files.

Issue #2301, opened 2026-09-06
Fix ConsumptionV1 schema regression breaking the Reservations Usage workbook


helderpinto:
## 🐛 Problem

The `ConsumptionV1` Log Analytics table schema in `Setup-LogAnalyticsTablesAndDCRs.ps1` defines the reservation ID column as `ReservationId_s`.

This column was previously named `ReservationId_g`. The Reservations Usage workbook and existing queries continue to reference `ReservationId_g`, so newly created tables use an incompatible schema and cause workbook queries to fail.

## 👣 Repro steps

1. Deploy or update the Azure Optimization Engine using DCR-based ingestion.
2. Run `Setup-LogAnalyticsTablesAndDCRs.ps1` against an environment where the `AzureOptimizationConsumptionV1_CL` table does not already exist.
3. Confirm that the table is created with a `ReservationId_s` column.
4. Open the Reservations Usage workbook.
5. Run a workbook query that references `ReservationId_g`.
6. Observe that the query fails because `ReservationId_g` cannot be resolved.

## 🤔 Expected

The `ConsumptionV1` schema should define the column as `ReservationId_g` to preserve backward compatibility with the existing Log Analytics schema and Reservations Usage workbook queries.

Existing environments that already received the incorrect `ReservationId_s` schema should also have a supported migration or repair path, since the setup script currently skips schema updates for existing tables.

## 🔧 Environment

- **Component:** Azure Optimization Engine
- **Ingestion method:** DCR-based Log Analytics ingestion
- **Affected table:** `AzureOptimizationConsumptionV1_CL`
- **Affected workbook:** Reservations Usage

## ℹ️ Additional context

The incorrect schema definition is:

```powershell
@{ name = "ReservationId_s"; type = "string" }
```

It should be:

```powershell
@{ name = "ReservationId_g"; type = "string" }
```

The Reservations Usage workbook consistently references `ReservationId_g`, including filters, joins, summaries, and workbook parameter fields.

Changing the schema definition fixes newly created tables. Existing tables created with `ReservationId_g` require additional handling because `Setup-LogAnalyticsTablesAndDCRs.ps1` intentionally skips schema updates when a table already exists.

Suggested acceptance criteria:
- New ConsumptionV1 tables define `ReservationId_g` instead of `ReservationId_s`.
- The generated DCR maps the source `ReservationId` field to `ReservationId_g`.
- Existing installations with `ReservationId_s` have a documented or automated migration path.
[...]

PR #2308, opened 2026-09-08
[Open Data] Fix ISF ratio scale, region coverage and culture handling, and keep retired-SKU reservations visible

### LINKED ISSUE #2300: InstanceSizeFlexibility incomplete
## 🐛 Problem
`src/open-data/InstanceSizeFlexibility.csv` is incomplete. Multiple instance size flexibility groups that exist in Microsoft's published ISF ratio blob (`isfratioblob.csv`) are missing entirely from the open data file, even though those SKUs are still deployed and covered by active reservations.

Examples of groups that are completely absent:

| InstanceSizeFlexibilityGroup | ArmSkuName | Ratio |
| --- | --- | --- |
| Av2 Series | Standard_A1_v2 | 1 |
| Av2 Series | Standard_A2_v2 | 2.1 |
| Av2 Series | Standard_A4_v2 | 4.4 |
| Av2 Series | Standard_A8_v2 | 9.2 |
| D Series High Memory | Standard_D11 | 1 |
| D Series High Memory | Standard_D12 | 2 |
| D Series High Memory | Standard_D13 | 4 |
| D Series High Memory | Standard_D14 | 8 |
| DS Series High Memory | Standard_DS11 | 1 |
| DS Series High Memory | Standard_DS12 | 2 |
| DS Series High Memory | Standard_DS13 | 4 |
| DS Series High Memory | Standard_DS14 | 8 |

## 👣 Repro steps
compare `src/open-data/InstanceSizeFlexibility.csv`  with https://aka.ms/isf and filter out the ArmSkuName starts with "Standard_" and maybe filter out the promo's 

## 🤔 Expected
A complete list for  `src/open-data/InstanceSizeFlexibility.csv` with all the vm's that are currently still being used

## 🔧 Environment
- **FinOps hub version:** 15
- **Power BI report type:** ADX (Azure Data Explorer)

## ℹ️ Additional context

[isf-missing.csv](https://github.com/user-attachments/files/31829784/isf-missing.csv)

## 🙋‍♀️ Ask for the community

We could use your help:

### LINKED ISSUE #2309: [AOE] Reservation workbooks show absolute quantities on the wrong ratio scale
## Problem

The reservation workbooks convert quantities into "smallest SKU" units by multiplying with the ISF `Ratio`:

```kql
| extend AvgRIsUsedInSmallestRatio = Ratio * AvgRIsUsedDaily
| summarize TotalReservedQuantity_s = sum(todouble(TotalReservedQuantity_s) * Ratio), ... by ISFGroup
```

That conversion only holds if the smallest SKU in a group has ratio `1`. Since #2222 the workbooks read `src/open-data/InstanceSizeFlexibility.csv`, which comes from the Catalogs API and is **not** normalized — [Microsoft's own ISF documentation](https://learn.microsoft.com/azure/cost-management-billing/reservations/instance-size-flexibility#normalize-isf-ratios) states it plainly:

> The raw ISF ratios from the API and powershell don't always start at `1` for the smallest SKU in a group. For example, the `BS Series` group starts at `0.25` and the `Ddsv5 Series` starts at `2`.

Measured on the published dataset: **162 of 319 groups start at `1`, 157 do not**. The retired `isfratioblob.csv` the workbooks read before #2222 was normalized (432 of 433 groups started at `1`).

## Impact

**Utilization percentages are correct.** `UsedQuantity / TotalReservedQuantity * 100` carries the factor on both sides, so it cancels.

**Absolute quantities are wrong** by each group's constant factor. `AvgRIsUsedDaily` derives from `Quantity * RINormalizationRatio`, so it is already in reservation units of the purchased SKU; multiplying by a non-normalized `Ratio` does not land in smallest-SKU units, which is what the column name claims and what the displayed `TotalReservedQuantity` implies.

## Fix

Normalize inside the workbook queries — divide each group's ratios by that group's minimum:

```kql
let ISFGroups = externaldata(ISFGroup:string, ArmSKUName:string, Ratio:double)
    [@"https://raw.githubusercontent.com/microsoft/finops-toolkit/dev/src/open-data/InstanceSizeFlexibility.csv"]
    with(ignoreFirstRecord=true)
| extend ArmSKUName = tolower(ArmSKUName)
| summarize MinRatio = min(Ratio) by ISFGroup
| join kind=inner (...) on ISFGroup
| extend Ratio = Ratio / MinRatio;
```

Contained to AOE, leaves the open data and the Power BI models untouched.

Affected files, all reading the same CSV via `externaldata()`:

- `src/optimization-engine/views/workbooks/reservations-usage.json`
- `src/optimization-engine/views/workbooks/reservations-potential.json`
- `src/optimization-engine/views/workbooks/benefits-simulation.json`

## Alternative

Publish the open data normalized instead, which would restore parity with the retired file for every consumer at once. That is a larger decision: it changes published values in 156 of 318 groups. It would not affect the Power BI models, which use the flexibility group but never the ratio.

## Not a regression from #2300

Pre-existing since #2222 migrated the workbooks off the retired blobs. #2300 changed no published ratio — it only added rows.

RolandKrummenacher:
## Summary

Fixes #2300 and #2309. Three parts: the reported symptom is fixed in the consumer, and the generator defects the investigation turned up are fixed in the data pipeline.

The published dataset goes from **1,353 to 1,363 SKUs** — no row removed, ten added by a wider region sweep. Schema unchanged at three columns.

## Part 1 — Reservations on retired SKUs disappeared from the reports

The reservation workbooks join the ISF open data with `kind=inner`. That data comes from the Catalogs API, which only returns SKUs you can still **purchase** — so an active reservation on a size Azure no longer sells has no row to match and is **dropped from the report entirely**. Not shown at zero utilization: absent. The customer keeps paying for a reservation the tool stops mentioning.

```kql
AzureOptimizationReservationsUsageV1_CL
| where ProvisioningState_s in ('Succeeded','Expiring') and todatetime(ExpiryDate_s) > now()
...
| join kind=inner ( ISFGroups ) on $left.SKUName_s == $right.ArmSKUName
```

All seven ISF joins in `reservations-usage.json` become `leftouter`, with the group and ratio falling back to the SKU itself:

```kql
| extend ISFGroup = coalesce(ISFGroup, SKUName_s), Ratio = coalesce(Ratio, 1.0)
```

**The `coalesce` is not optional.** A plain `leftouter` is worse than today: `Ratio` and `ISFGroup` come back null, `TotalReservedQuantity_s * Ratio` becomes null, and `summarize ... by ISFGroup` collapses every unmatched reservation into a single null bucket with an empty utilization column.

**`Ratio = 1` is correct here.** `Util7Days_s` comes from Azure and already accounts for instance size flexibility, so the percentage is right as-is. The ratio only converts between group units, a no-op for a group of one.

`reservations-potential.json` and `benefits-simulation.json` keep `kind=inner`. They model what to *buy*, and there restricting to purchasable SKUs is correct.

**Known limitation.** Drilldowns fall back to the *consumed* size while the overview falls back to the *reserved* size. For a reservation on `Standard_DS4_v2` running `DS3_v2` underneath, those don't meet — which sizes belong to a retired group is exactly what the ISF data no longer carries. A drilldown on an unmatched reservation therefore shows usage of the identical SKU only. Better than nothing, not a full restoration.

## Part 2 — Three generator defects

**Region coverage.** The sweep used a hardcoded list of 26 regions, so SKUs that launch in only a handful of regions never entered the dataset, and the list had to be hand-maintained. It now enumerates every physical region from the ARM locations API — 63 today. Note `regionType` is nested under `metadata`, not at the top level, and 46 of the 109 entries are logical groupings (`global`, `unitedstates`, `europe`) that are not valid catalog scopes. An empty enumeration fails the run rather than sweeping a partial set. Measured: **7.9 minutes** for 63 regions × 3 reserved types, against a 60-minute timeout.

**Culture handling.** Ratios were parsed and written with the current culture:

| culture | `TryParse("2.1")` | `Export-Csv` writes |
| --- | --- | --- |
| en-US | `2.1` | `"2.1"` |
| de-DE | **`21`** | **`"2,1"`** |
| fr-CH | **fails, row dropped** | **`"2,1"`** |

CI runs on `ubuntu-latest` and was never affected, but the README documents running the generator by hand, so a contributor in a comma-decimal locale would have published a corrupt file — `"2,1"` breaks the `externaldata(... Ratio:double)` joins and both Power BI partitions typed as `number`. Both parse sites and the write path now pin `InvariantCulture`.

**Zero ratios.** The API returns `azure_redis_cache_isolated_i100` with a ratio of `0`, and `benefits-simulation.json` divides by `Ratio`. Non-positive ratios are dropped, with a unit test asserting the invariant on the published file.

## Part 3 — Ratios were on the wrong scale

The Catalogs API leaves ratios unnormalized: **only 52 of its 211 groups have a smallest SKU of `1`**. [Microsoft's documentation calls this out](https://learn.microsoft.com/azure/cost-management-billing/reservations/instance-size-flexibility#normalize-isf-ratios) — `BS Series` starts at `0.25`, `Ddsv5 Series` at `2` — and publishes a normalization step of its own. The retired `isfratioblob.csv` this dataset replaced was normalized, so #2222 silently changed the scale under every consumer.

That matters because the Optimization Engine converts quantities into units of a group's smallest SKU — `AvgRIsUsedInSmallestRatio` is the variable's own name — and multiplying by a ratio that doesn't start at `1` inflates the absolute figures by a per-group constant. Utilization percentages carry the factor on both sides and were never wrong; the displayed quantities were.

It also makes Part 1 coherent. An unmatched reservation falls back to `Ratio = 1`, which is implicitly "smallest SKU = 1"; next to an unnormalized group that would put two rows of one table in different units.

Each group is now normalized. `-Normalize` is replaced by `-Raw`, so the switch names the departure from the default rather than the default itself.

**Parity with the retired file is close but not total**, and the gap is worth understanding rather than glossing. Of the 1,291 rows the two files share, **1,223 now agree**. The 68 that don't split cleanly:

| | groups | rows | why |
| --- | --- | --- | --- |
| Different anchor | 12 | 52 | The API no longer returns the historically smallest SKU, so normalization anchors on a different one. `BS Series High Memory` is down to `Standard_B20ms` alone, which normalizes to `1` rather than `40.2` — internally consistent, since no other SKU of that group remains. |
| Different proportions | 11 | 16 | The API and the retired file disagree on the proportions themselves, not the unit. All dedicated host groups, 5-10% apart. The API is the newer source. |

Measured impact on consumers: **Power BI is unaffected** — `x_CommitmentDiscountFlexRatio` appears in no DAX measure and no report visual, only in linguistic metadata; the visuals use the flexibility *group*. The Optimization Engine's absolute quantities change, which is the point.

## What this deliberately does not do

An earlier revision backfilled 1,220 retired SKUs from the ratio file the API replaced. That was dropped. Those rows would be frozen forever — nothing can refresh or correct them once the source blob is gone — inside a dataset whose contract is a weekly refresh from the authoritative API. Fixing the consumer avoids carrying dead reference data.

Fixes #2309 as part of this.

## Test plan

- [x] `Invoke-Pester ./src/powershell/Tests/Unit/* ./src/powershell/Tests/Lint/*` — **6,010 passed, 0 failed** (4 pre-existing skips)
- [x] Generator run end-to-end against the Catalogs API: 63 regions, 7.9 min
- [x] Verified against v15: 0 rows removed, 10 added; ratios rescaled per group by normalization
- [x] Verified against the retired `isfratioblob.csv`: 1,223 of 1,291 shared rows agree, the rest classified above
- [x] Every published group has a smallest ratio of `1`, asserted by a unit test
- [x] 22 unit tests for the generator, including all four region-enumeration paths and the `de-DE` round trip
- [x] Open **Reservations usage** against a subscription holding a reservation on a retired size and confirm it now appears with a correct utilization percentage
- [x] Confirm reservations on current sizes are unchanged — same groups, quantities and percentages

**The workbook change is reasoned, not verified.** No test in this repo executes workbook KQL, and I have no Optimization Engine workspace with reservation data to run it against. The join semantics and the null behaviour that motivates the `coalesce` want confirming by someone who does.


RolandKrummenacher:
The ratio-scale follow-up flagged in the description is now tracked separately in #2309, so it doesn't hold up this PR. That one is about the Optimization Engine's math; this PR leaves every published ratio unchanged.

RolandKrummenacher:
Depends on merge order with #2298, which adds `src/powershell/Private/Get-OpenDataInstanceSizeFlexibility.ps1` — generated from this CSV at its pre-change 1,353 rows.

This PR takes the CSV to 2,573 rows, so that generated file has to be rebuilt. It is deliberately **not** included here: both PRs would otherwise add the same new file and collide.

Plan: let #2298 merge first, then merge `dev` into this branch and run `Build-OpenData.ps1 -PowerShell -Name InstanceSizeFlexibility` so the regenerated file lands with the data it describes.

Note that Open Data CI passing on this PR does not mean the file was regenerated — on `dev` the workflow uses `git commit -a`, which skips the still-untracked file. #2298 fixes that with `git add -A`.

Hub KQL is unaffected: `Build-OpenData.ps1 -Hubs` uses an allowlist that excludes this dataset, and the committed `OpenDataFunctions.kql` contains no ISF data.

src/scripts/Update-InstanceSizeFlexibility.ps1
```diff
@@ -214,12 +264,16 @@ function Get-IsfRecords
         # 'arm_sku_name_placeholder'); skip them so they don't leak into the public CSV.
         if ($armSkuName -like '*placeholder*') { continue }
 
-        if ($flexGroup -and $ratio -and $armSkuName)
+        # A ratio of 0 carries no flexibility information and consumers divide by it -- the
+        # Optimization Engine's benefits simulation does exactly that -- so drop it rather than
+        # publish a division by zero. The API returns one today (Azure Redis Cache Isolated).
+        $parsedRatio = 0.0
+        if ($flexGroup -and $armSkuName -and [double]::TryParse($ratio, [ref]$parsedRatio) -and $parsedRatio -gt 0)
         {
```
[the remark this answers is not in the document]
RolandKrummenacher: Fixed in 45923113 — and the finding is bigger than the read side.

Measured across cultures, parsing `"2.1"`:

| culture | `TryParse` (current) | `Export-Csv` writes |
| --- | --- | --- |
| en-US | `2.1` | `"2.1"` |
| de-DE | **`21`** | **`"2,1"`** |
| fr-CH | **fails, row dropped** | **`"2,1"`** |

So the write path corrupts the published file too, which this comment does not cover: under a comma-decimal culture `Export-Csv` emits `"2,1"`, breaking the Optimization Engine`s `externaldata(ISFGroup:string, ArmSKUName:string, Ratio:double)` and both Power BI partitions typed as `number`.

Both parse sites and the write path now pin `InvariantCulture` (with `NumberStyles::Float` on the parses). Added a regression test that drives the generator under `de-DE` and asserts `"2.1"` and `"4.5"` survive the round trip — without the fix the carried-forward row comes back as `21`.

CI itself was never affected: it runs on `ubuntu-latest`. The exposure is the manual run the README documents.

docs-mslearn/toolkit/changelog.md
```diff
@@ -78,6 +78,11 @@ The following section lists features and enhancements that are currently in deve
 
 - **Added**
   - Added a new [Instance size flexibility](open-data.md#instance-size-flexibility) dataset that maps each ARM SKU to its instance size flexibility group and ratio, sourced from the Azure Reservations Catalogs API. It replaces the deprecated ISF ratio files hosted on `ccmstorageprod.blob.core.windows.net` ([#2090](https://github.com/microsoft/finops-toolkit/issues/2090)).
+- **Fixed**
+  - Fixed the instance size flexibility dataset omitting SKUs that are no longer available to purchase. The Catalogs API only returns purchasable SKUs, so retired series that reservations still cover (Av2, D, DS, Dv2, Dv3, Ev3, F, G, H, LS, NC, NV, and others) were missing. The dataset is now additive and backfilled from the retired ratio file, growing from 211 to 318 flexibility groups and from 1,353 to 2,574 SKUs ([#2300](https://github.com/microsoft/finops-toolkit/issues/2300)).
```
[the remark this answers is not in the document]
RolandKrummenacher: Fixed in 45923113 — the count is now 2,573.

The cause was not the header, though: the entry was written when the dataset had 2,574 SKUs, and a later commit dropped the one row whose ratio was `0` (`azure_redis_cache_isolated_i100`, which the benefits simulation workbook divides by). The number was simply stale. Verified against the file: 2,573 SKUs across 318 groups.

docs-mslearn/toolkit/changelog.md
```diff
@@ -78,6 +81,11 @@ The following section lists features and enhancements that are currently in deve
...
+  - Fixed ratios being parsed and written with the current culture. On a comma-decimal machine `2.1` was read as `21` or rejected, and written back as `2,1`, which would break every consumer of the file ([#2308](https://github.com/microsoft/finops-toolkit/pull/2308)).
+  - Fixed the dataset publishing unnormalized ratios. The Catalogs API leaves them unnormalized, so a group's smallest SKU is usually not `1`, while the retired ratio file it replaced was normalized and the Optimization Engine converts quantities into units of a group's smallest SKU. Absolute quantities in the reservation workbooks were inflated by a per-group constant as a result. Each group is now normalized, matching the retired file for 1,223 of the 1,291 rows they share ([#2309](https://github.com/microsoft/finops-toolkit/issues/2309)).
+  - Fixed the dataset publishing a SKU whose ratio is `0`, which the Optimization Engine's benefits simulation divides by. Non-positive ratios are now dropped ([#2308](https://github.com/microsoft/finops-toolkit/pull/2308)).
```
flanakin: nit: We don't need to doc changes to a new dataset
If there's anything you want to call out that we're doing diff from the original dataset, we could add those as nested bullets (e.g., benefits for moving to this from the one that's going away).

src/optimization-engine/views/workbooks/reservations-usage.json
flanakin, line 1: @helderpinto Can you review?

flanakin (approved):

src/optimization-engine/views/workbooks/reservations-usage.json
helderpinto: @RolandKrummenacher, @flanakin , I just tested the Reservations Usage workbook in a customer with impacted sizes and it now shows correctly those sizes when no ISF group is matched. Also the other non-impacted sizes continue to show correctly their usage and savings.

docs-mslearn/toolkit/open-data.md
```diff
@@ -193,9 +193,9 @@ Sample data:
 
 A few important notes about the data:
 
-- It covers Virtual Machines, Redis Cache, and Dedicated Host.
+- It covers Virtual Machines, Redis Cache, and Dedicated Host across every physical Azure region.
 - `ArmSkuName` is unique across the file and can be used as a join key on its own.
-- Ratios are the raw Microsoft values within each group (the smallest SKU isn't always `1`).
+- Ratios are normalized so the smallest SKU in each flexibility group has a ratio of `1`. Compare ratios only within a flexibility group.
```
helderpinto, line 198: I am not sure whether generating a CSV whose ISF ratios are inconsistent with what Cost Management exports report in the `RINormalizationRatio` property of the `AdditionalInfo` column is a good idea.

helderpinto: I am not seeing the full picture, but once I address #2309 I'll have a better idea.

helderpinto (commented):
The PR description should be updated to remove the "Fixes #2309" statement, because it does not address the so called ratio scale issue.

flanakin (commented):

**Summary:** Strong, well-reasoned PR. The generator fixes (invariant culture, region enumeration, non-positive ratio drop, normalization) are correct and well tested, and the `leftouter` + `coalesce` pattern is the *only* safe form here — KQL strings have no null, so `leftouter` yields `""` and `coalesce` (documented as "first non-null, or non-empty for string") handles both the string and the `double` case where `isnull()` would silently fail. Verified locally: 23 unit tests pass (the description says 22), all 3,593 lint tests pass, and the published CSV holds every claimed invariant — 1,363 rows, 213 groups, every group normalized to a smallest ratio of 1, `ArmSkuName` globally unique, no non-positive ratios, invariant decimal formatting. The Power BI claim also checks out: `x_CommitmentDiscountFlexRatio` appears only in linguistic metadata, in no DAX measure or visual.

One blocker: the fix stops three joins short of complete, and the reported symptom survives in one of them.

### 🚫 Blockers (1)

1. `reservations-usage.json` has **10** ISF joins, not 7. Three were already `leftouter` before this PR and still have no `coalesce` guard — including "Unused Reservations over time (by VM count)" (line 1093), which multiplies by a null `Ratio` and drops retired-SKU reservations exactly as #2300 describes.

### ⚠️ Should fix (3)

1. The Optimization Engine changelog section now has two separate `- **Fixed**` blocks, the new one placed before `- **Added**`.
2. The instance size flexibility sample table in `open-data.md` shows rows that exist nowhere in the published file.
3. `leftouter` widens the two ISF-grouped tables to every reservation type, with dead-end drilldowns — an undocumented scope change.

### 💡 Suggestions (1)

1. The ARM locations call has no retry, unlike every Catalogs call.

---

**Not verified, and I couldn't verify it either:** no test in this repo executes workbook KQL, so the runtime behavior of these queries is reasoned, not run. Everything I checked above is static analysis plus the Pester suites and the published CSV.

src/optimization-engine/views/workbooks/reservations-usage.json
```diff
@@ -1310,7 +1310,7 @@
...
...otalReservedQuantity_s) * todouble(Util7Days_s) / 100)\r\n    | extend UnusedQuantity30d = todouble(TotalReservedQuantity_s) - (todouble(TotalReservedQuantity_s) * todouble(Util30Days_s) / 100)\r\n    | project ReservationId_g, ReservationName_s=DisplayName_s, SKUName_s=tolower(SKUName_s), Location_s, TotalReservedQuantity_s, Term_s, AppliedScopeType_s, UnusedQuantity, UnusedQuantity30d\r\n) on ReservationId_g\r\n| project-away ReservationId_g1\r\n| join kind=leftouter ( ISFGroups ) on $left.SKUName_s == $right.ArmSKUName\r\n| extend ISFGroup = coalesce(ISFGroup, SKUName_s), Ratio = coalesce(Ratio, 1.0)\r\n| summarize TotalUnusedCost=sum(TotalUnusedCost), TotalReservedQuantity_s=sum(todouble(TotalReservedQuantity_s)*Ratio), UnusedQuantity=sum(UnusedQuantity*Ratio), UnusedQuantity30d=sum(UnusedQuantity30d*Ratio) by ISFGroup, Location_s, Term_s, AppliedScopeType_s\r\n| extend Util7Days_s = (1-UnusedQuantity/TotalReservedQuantity_s)*100, Util30Days_s = (1-UnusedQuantity30d/TotalReservedQuantity_s)*100\r\n| project-reorder ISFGroup, TotalUnusedCost, Location_s, Term_s, AppliedScopeType_s, TotalReservedQuantity_s, Util7Days_s, UnusedQuantity, Util30Days_s\r\n| order by TotalUnusedCost",
```
flanakin: 🤖 [AI][Claude Code] 🚫 **Blocker**

The PR description says "all seven ISF joins in `reservations-usage.json` become `leftouter`." There are **ten**. Three were already `leftouter` before this PR, so they never appeared in the diff — and none of them got the `coalesce` guard. Your own reasoning in the description ("a plain `leftouter` is worse than today") applies to all three verbatim:

| Line | Query | What happens on an unmatched SKU |
| --- | --- | --- |
| 1093 | Unused Reservations over time (by VM count) | `UnusedVMs = iif('{UseISF}' == 'Yes', UnusedHours * Ratio / 24, ...)` → `Ratio` is null → `UnusedVMs` is null → **the reservation contributes nothing to the chart**. `SKUName` becomes `strcat("", " ", Location_s)` = `" eastus"`. |
| 1060 | Cost of Unused Reservations over time (by SKU) | `strcat(ISFGroup, " ", Location_s)` collapses every unmatched SKU in a region into a single mislabeled `" eastus"` series. |
| 1488 | Reservations grid | `ISFGroup` is projected directly → blank cell for every unmatched reservation. |

Line 1093 is the important one: with **Use ISF = Yes** — a first-class workbook toggle — a reservation on a retired size still silently disappears from the chart. That is the symptom [#2300](https://github.com/microsoft/finops-toolkit/issues/2300) reports, in the same workbook, and the changelog entry added here claims it's fixed.

The fix is the line you already wrote, inserted after each of the three joins:

```kusto
| extend ISFGroup = coalesce(ISFGroup, SKUName_s), Ratio = coalesce(Ratio, 1.0)
```

Line 1488 uses `skuName` as the alias rather than `ArmSKUName`, but the left column is still `SKUName_s`, so the same line works there too.

Worth a short KQL comment next to one of them, too: `coalesce` is load-bearing precisely because it treats *empty* as missing for strings. A future "simplification" to `iif(isnull(ISFGroup), ...)` would silently reintroduce the empty-group bucket, since [KQL strings have no null](https://learn.microsoft.com/kusto/query/scalar-data-types/null-values) and `isnull("")` is `false`.

src/optimization-engine/views/workbooks/reservations-usage.json
```diff
@@ -536,7 +536,7 @@
...
..._s and Term_s\r\n| join kind=leftouter ( OnDemandPriceSheet ) on $left.MeterId_g == $right.MeterID_g\r\n| extend DiscountPercent = (1 - ReservationPrice/OnDemandUnitPrice) * 100\r\n| extend AvgRIsUsedInSmallestRatio = Ratio * AvgRIsUsedDaily\r\n| summarize TotalReservedQuantity_s=sum(todouble(TotalReservedQuantity_s)*Ratio), AvgRIsUsedDaily=sum(AvgRIsUsedInSmallestRatio), UsedQuantity=sum(UsedQuantity*Ratio), UsedQuantity30d=sum(UsedQuantity30d*Ratio), AvgDiscountPercent=avg(DiscountPercent) by ISFGroup, Location_s, Term_s, AppliedScopeType_s\r\n| extend Util7Days_s = UsedQuantity/TotalReservedQuantity_s*100, Util30Days_s = UsedQuantity30d/TotalReservedQuantity_s*100\r\n| extend AvgRIUsagePercentInSmallestRatio = round(AvgRIsUsedDaily / TotalReservedQuantity_s * 100, 1)\r\n| extend AvgDiscountPercent=iif(AvgDiscountPercent > 0.0, AvgDiscountPercent, 0.0)\r\n| extend SavingsMargin=round(todouble(Util7Days_s))-100.0+AvgDiscountPercent \r\n| project-away AvgRIUsagePercentInSmallestRatio, AvgRIsUsedDaily\r\n| project-reorder ISFGroup, Location_s, Term_s, AppliedScopeType_s, TotalReservedQuantity_s, Util7Days_s, UsedQuantity, Util30Days_s, UsedQuantity30d\r\n| order by Util7Days_s asc",
```
flanakin: 🤖 [AI][Claude Code] ⚠️ **Should fix**

The `inner` join wasn't only an ISF lookup — it was also the filter that kept this table to reservation types that *have* ISF. `AzureOptimizationReservationsUsageV1_CL` carries every reservation type (the export runbook writes `ResourceType` from `reservedResourceType`), and the workbook's own **Resource Type** picker defaults to `value::all`.

So after this change, SQL Database, Cosmos DB, App Service, Databricks and every other reserved type flow into "Reservation Usage Details grouped by ISF group" (and into "Unused Reservations Details" at line 1313), each as a pseudo-group named by its lowercase SKU. Two consequences:

1. The **ISF group** column becomes a mix of real group names and raw SKU names with nothing distinguishing them.
2. Clicking one of those rows drills into an empty chart — the drilldowns key on `AdditionalInfo_s.ServiceType`, which consumption rows for those types don't populate, so `VMSize` is empty and `where ISFGroup == '{selectedISFGroup:value}'` matches nothing. Previously those rows never appeared, so the user never got a dead-end click.

Arguably a net improvement over silently hiding them, and it does match what the Resource Type picker already offers. But it's a scope change the PR description doesn't mention. Either scope the fallback to types that can have ISF:

```kusto
| where ResourceType in ('VirtualMachines', 'RedisCache', 'DedicatedHost')
```

(added to the `AzureOptimizationReservationsUsageV1_CL` subquery, with `ResourceType` added to its `project`), or call the widening out in the changelog alongside the known limitation you already documented.

docs-mslearn/toolkit/changelog.md
```diff
@@ -60,6 +60,9 @@ The following section lists features and enhancements that are currently in deve
 
 ### [Optimization Engine](optimization-engine/overview.md)
 
+- **Fixed**
```
flanakin: 🤖 [AI][Claude Code] ⚠️ **Should fix**

This adds a second `- **Fixed**` block to the Optimization Engine section — there's already one at line 70 with the #2301 entry. Every other section in the file uses a single **Added** → **Changed** → **Fixed** sequence, and a duplicate bullet heading renders as two separate "Fixed" lists in the same section.

Move this entry into the existing `- **Fixed**` block at line 70 and drop the new heading and the blank line after it:

```markdown
- **Fixed**
  - Fixed a regression in the `AzureOptimizationConsumptionV1_CL` schema that was breaking the Reservations Usage workbook ([#2301](...)).
  - Fixed reservations disappearing from the reservation workbooks when their SKU is no longer sold. ...
```

Minor, in the same block: the region-coverage entry under **Open data** is attributed to [#2300](https://github.com/microsoft/finops-toolkit/issues/2300), which is the reservations-disappearing report. The region sweep is a generator defect this investigation turned up, not what #2300 describes — [#2309](https://github.com/microsoft/finops-toolkit/issues/2309) or the PR link would be more accurate.

docs-mslearn/toolkit/open-data.md
```diff
@@ -193,9 +193,9 @@ Sample data:
 
 A few important notes about the data:
 
-- It covers Virtual Machines, Redis Cache, and Dedicated Host.
+- It covers Virtual Machines, Redis Cache, and Dedicated Host across every physical Azure region.
```
flanakin, line 196: 🤖 [AI][Claude Code] ⚠️ **Should fix**

The sample table just above (lines 190-192) shows rows that don't exist in the file this page documents:

```
| Av2 Series | Standard_A1_v2 | 1    |
| Av2 Series | Standard_A2_v2 | 2.1  |
| Av2 Series | Standard_A4_v2 | 4.44 |
```

There is no `Av2 Series` group and no `Standard_A*_v2` SKU anywhere in `InstanceSizeFlexibility.csv` — not in this revision and not on `dev`. The naming isn't representative either: 671 of the 1,363 published rows use the `Virtual Machines <family> Linux` form, e.g. `Virtual Machines DCadsv6-series Linux`.

Pre-existing (it came in with #2090), but #2090 is still unreleased and ships in the same version as this change, so it hasn't reached anyone yet — and this PR is already editing the notes directly under the table. Replacing it with three real rows now avoids publishing a sample no reader can find in the data:

```markdown
| InstanceSizeFlexibilityGroup | ArmSkuName        | Ratio |
| ---------------------------- | ----------------- | ----- |
| Basv2 Series                 | Standard_B2als_v2 | 4     |
| Basv2 Series                 | Standard_B2as_v2  | 8     |
| Basv2 Series                 | Standard_B16as_v2 | 64.04 |
```

Those three are verbatim from the CSV in this PR. A group in the `Virtual Machines ... Linux` style would be even more representative of what a reader actually sees.

src/scripts/Update-InstanceSizeFlexibility.ps1
```diff
@@ -101,6 +116,35 @@ if (-not $SubscriptionId)
...
 }
+if (-not $Location)
+{
+    # Physical regions only: the response also carries logical groupings ('global', 'unitedstates',
+    # 'europe', 'asiapacific'), which are not valid catalog scopes -- 46 of the 109 entries at the
+    # time of writing. Note regionType is nested under metadata, not at the top level.
+    # Az.Accounts is the only module the workflow installs, so this uses the REST API rather than
+    # Get-AzLocation, which lives in Az.Resources.
+    $locationsUri = "https://management.azure.com/subscriptions/$SubscriptionId/locations?api-version=$LocationApiVersion"
+    $locationsResponse = Invoke-AzRestMethod -Uri $locationsUri -Method GET
```
flanakin: 🤖 [AI][Claude Code] 💡 **Suggestion**

Every Catalogs call goes through `Invoke-CatalogsApi`, which retries 429 and 5xx up to five times with exponential backoff. This call — now the first thing the weekly run does, and a hard prerequisite for all 189 of them — has none. A single transient 500 or 429 from ARM fails the whole refresh before it fetches a single catalog page, and the next attempt is a week away (or a manual `workflow_dispatch`).

Given that the script already treats a partial region sweep as fatal, it's worth being at least as resilient here as it is for the calls that follow. Extracting the retry loop from `Invoke-CatalogsApi` into a small `Invoke-ArmRequest` helper both calls share would cover it without duplicating the backoff logic.

src/optimization-engine/views/workbooks/reservations-usage.json
RolandKrummenacher: Fixed in 6fcf79f8 — all three now carry the guard, and all ten ISF joins are `leftouter` + guarded.

- **1093** (Unused Reservations over time, by VM count): `ISFGroup = coalesce(ISFGroup, SKUName_s), Ratio = coalesce(Ratio, 1.0)`. The `Ratio` half is the one that mattered — it's what kept the reservation off the chart entirely under Use ISF = Yes.
- **1060** (Cost of Unused Reservations over time, by SKU): `ISFGroup` only; `Ratio` is never referenced in that query. The `strlen(SKUName) < 2` → `'Canceled RIs'` branch still works, and for the reason you gave: the join to `AzureOptimizationReservationsUsageV1_CL` there is `leftouter`, so a canceled RI has `ISFGroup` *and* `SKUName_s` empty, `coalesce` returns empty for strings, and `SKUName` stays `" "`.
- **1488** (Reservations grid): `ISFGroup` only, same reason — `Ratio` isn't projected.

The KQL comment went in at 1060, where the empty-vs-null distinction is load-bearing twice over.

src/optimization-engine/views/workbooks/reservations-usage.json
RolandKrummenacher: Scoped, in 6fcf79f8. Two deviations from the suggestion, both deliberate:

- The filter goes *inside* the `AzureOptimizationReservationsUsageV1_CL` subquery rather than after the join, so `project` doesn't need to carry `ResourceType` — one line instead of two, at both 539 and 1313.
- `in~` rather than `in`, per the KQL section of `Coding-guidelines.md`.

The type list is a `let IsfResourceTypes = dynamic(['VirtualMachines', 'RedisCache', 'DedicatedHost'])` with a comment, matching the generator's own default `-ReservedResourceType` — i.e. the types the dataset can ever cover.

Two things worth recording that weren't in the analysis:

**Nothing is hidden product-wide either way.** Both affected grids sit in groups gated on `UseISF == 'Yes'` (`reservationAnalysisGroupISF`, and the unused-reservations equivalent). Each has a twin gated on `UseISF == 'No'` — "Reservation Usage Details" and "Unused Reservations Details" — with no ISF join at all, which already lists every reserved resource type. So scoping restores the pre-PR split rather than dropping anything from the workbook.

**The dead-end drilldown is confirmed, not inferred.** The Use ISF = No drilldown keys on `{selectedReservation:value}` and carries an explicit fallback for rows with no `ServiceType`:

```kusto
| extend VMSize = tostring(parse_json(AdditionalInfo_s).ServiceType)
| extend ConsumedSize = iif(isnotempty(VMSize), VMSize, strcat(MeterSubCategory_s, ' ', MeterName_s))
```

That fallback exists because those rows are real. The ISF drilldown filters on `ISFGroup == '{selectedISFGroup:value}'` *before* reaching its copy of the same line, and `ISFGroup` is `coalesce(ISFGroup, VMSize)` — so for exactly those rows it's empty and matches nothing.

docs-mslearn/toolkit/changelog.md
RolandKrummenacher: Moved into the existing `- **Fixed**` block; the section now reads Added → Changed → Fixed like every other one.

On the attribution point — the region-coverage entry is gone entirely rather than relinked. Per your other comment, the whole instance size flexibility `- **Fixed**` block was dropped, since the dataset is new in this same release. What survives is a nested bullet under **Added** describing the region coverage as a property of the dataset, with no issue link, which sidesteps the question.

docs-mslearn/toolkit/open-data.md
RolandKrummenacher: Replaced in 6fcf79f8, with the `Virtual Machines ... Linux` form you flagged as more representative. I picked the group anchored at ratio `1` so the sample also demonstrates the normalization note directly below it:

| InstanceSizeFlexibilityGroup         | ArmSkuName        | Ratio |
| ------------------------------------ | ----------------- | ----- |
| Virtual Machines Dadsv6-series Linux | Standard_D2ads_v6 | 1     |
| Virtual Machines Dadsv6-series Linux | Standard_D4ads_v6 | 2     |
| Virtual Machines Dadsv6-series Linux | Standard_D8ads_v6 | 4     |

Verbatim from the CSV in this PR. Minor correction to the count: 664 rows use that form, not 671.

src/scripts/Update-InstanceSizeFlexibility.ps1
RolandKrummenacher: Done in 6fcf79f8 — extracted `Invoke-ArmRequest` with the backoff lifted out of `Invoke-CatalogsApi`, and both call sites go through it.

The helper takes a `Description` used verbatim in every message, so the existing wording is unchanged: `HTTP 403 enumerating regions: ...` and `Failed after 5 retries on VirtualMachines/eastus page 3: ...` read exactly as before, and the existing tests pass untouched.

New coverage: `Retries a transient failure when enumerating regions instead of failing the run` — 503 on the first locations call, success on the second, asserts the sweep still runs and the output is written. 24/24 in that file.

docs-mslearn/toolkit/changelog.md
RolandKrummenacher: Dropped the `- **Fixed**` block. Three nested bullets under **Added** now cover what differs from the retired `isfratioblob.csv`:

- ratio scale — same normalization as the retired file, with the row-level comparison
- full physical-region coverage instead of a fixed list
- weekly refresh from the live catalog, so retired SKUs age out without anyone republishing

The culture and zero-ratio entries are gone outright — they were generator defects in a dataset that hasn't shipped, so there's nothing for a reader to have been affected by.

src/optimization-engine/views/workbooks/reservations-usage.json
RolandKrummenacher: @helderpinto one behavior change since you tested: the unmatched-SKU fallback is now scoped to the reserved resource types that have instance size flexibility (`VirtualMachines`, `RedisCache`, `DedicatedHost`), so the Use ISF = Yes views no longer pick up SQL Database / Cosmos DB / App Service reservations the way they did in the build you ran against.

The retired-VM-SKU path you verified is unchanged — that's still `leftouter` + `coalesce`, now applied to three more joins that were already `leftouter` before this PR.

RolandKrummenacher:
## Review feedback addressed — ready for another look

Everything from the 9 Sep review is in, in `6fcf79f8`, plus a `dev` merge in `d4b8e31e`. Six threads resolved, one left open deliberately. All checks green.

@flanakin — your approval on 9 Sep 20:19 predates the agent review you posted at 23:58 the same evening, so it doesn't cover the blocker or anything below. Re-approval would be appreciated.

### What changed

| Finding | Resolution |
| --- | --- |
| 🚫 Ten ISF joins, three unguarded | `coalesce` guard added at 1060, 1093 and 1488. All ten are now `leftouter` + guarded. `Ratio` is only coalesced where it's actually referenced; 1060's `'Canceled RIs'` branch is preserved, and a KQL comment there records why `coalesce` and not `isnull` |
| ⚠️ `inner` → `leftouter` widened scope | Scoped. `let IsfResourceTypes = dynamic(['VirtualMachines', 'RedisCache', 'DedicatedHost'])` + `where ResourceType in~ (IsfResourceTypes)` in the two `AzureOptimizationReservationsUsageV1_CL` subqueries |
| ⚠️ Duplicate `- **Fixed**` block | Merged into the existing one; the section reads Added → Changed → Fixed |
| ⚠️ Sample table rows don't exist | Replaced with three `Virtual Machines Dadsv6-series Linux` rows, verbatim from the CSV in this PR, anchored at ratio `1` so the sample demonstrates the normalization note below it |
| 💡 No retry on the first ARM call | Extracted `Invoke-ArmRequest`; both call sites use it, message wording unchanged, new test covers a 503 during region enumeration |
| nit: don't document fixes to a new dataset | The instance size flexibility `- **Fixed**` block is gone. Three nested bullets under **Added** now cover what differs from the retired `isfratioblob.csv` |

### The one judgment call worth a second look

I scoped the fallback rather than documenting the widening. Two things that weren't in the original analysis pushed it that way:

**Nothing is hidden either way.** Both affected grids sit in groups gated on `UseISF == 'Yes'`. Each has a twin gated on `UseISF == 'No'` — "Reservation Usage Details" and "Unused Reservations Details" — with no ISF join at all, already listing every reserved resource type. Scoping restores the pre-PR split rather than dropping anything from the workbook.

**The dead-end drilldown is confirmed, not inferred.** The `Use ISF = No` drilldown carries an explicit fallback for rows with no `ServiceType`:

```kusto
| extend VMSize = tostring(parse_json(AdditionalInfo_s).ServiceType)
| extend ConsumedSize = iif(isnotempty(VMSize), VMSize, strcat(MeterSubCategory_s, ' ', MeterName_s))
```

That line exists because those rows are real. The ISF drilldown filters on `ISFGroup == '{selectedISFGroup:value}'` before reaching its copy of it, and `ISFGroup` is `coalesce(ISFGroup, VMSize)` — so for exactly those rows it's empty and matches nothing.

### @helderpinto

Two things for you, since you're the only real-data check this PR has:

- The scoping landed **after** the build you tested. The `Use ISF = Yes` views no longer pick up SQL Database / Cosmos DB / App Service reservations. The retired-VM-SKU path you verified is unchanged.
- Your `RINormalizationRatio` thread is the one I left open on purpose, since you tied it to #2309. If the ratio scale here turns out to conflict with what Cost Management reports in `AdditionalInfo`, that's worth settling before this ships rather than after.

### Verification

- 833/833 Pester tests locally after the `dev` merge (24 generator including the new retry test, 445 KQL join-kind, 144 KQL operator, 220 ADF trigger)
- Workbook JSON parses; prettier clean; PSScriptAnalyzer shows no new findings
- `dev` merge conflict was `ms.date` only, resolved to today per AGENTS.md; both sides' changelog entries kept

PR #2310, opened 2026-09-09
[Optimization Engine] Keep reservations visible when their SKU has no ISF ratio

### LINKED ISSUE #2300: InstanceSizeFlexibility incomplete

RolandKrummenacher:
## Summary

Fixes #2300. The reservation workbooks join the instance size flexibility open data with `kind=inner`. That data comes from the Azure Catalogs API, which only returns SKUs you can still **purchase** — so an active reservation on a size Azure no longer sells has no row to match and is **dropped from the report entirely**. Not shown at zero utilization: absent. The customer keeps paying for a reservation the tool stops mentioning.

Reservations run one or three years and routinely outlive the sellability of their size, so this is reachable in ordinary use. It is what the reporter of #2300 hit.

## Change

All seven ISF joins in `reservations-usage.json` become `leftouter`, with the group and ratio falling back to the SKU itself:

```kql
| join kind=leftouter ( ISFGroups ) on $left.SKUName_s == $right.ArmSKUName
| extend ISFGroup = coalesce(ISFGroup, SKUName_s), Ratio = coalesce(Ratio, 1.0)
```

**The `coalesce` is not optional.** A plain `leftouter` is worse than today: `Ratio` and `ISFGroup` come back null, `TotalReservedQuantity_s * Ratio` becomes null, and `summarize ... by ISFGroup` collapses every unmatched reservation into a single null bucket with an empty utilization column.

**`Ratio = 1` is correct for these rows.** `Util7Days_s` comes from Azure and already accounts for instance size flexibility, so the percentage is right as-is. The ratio only converts between group units, which is a no-op for a group of one.

`reservations-potential.json` and `benefits-simulation.json` keep `kind=inner`. They model what to *buy*, and there restricting to purchasable SKUs is correct.

## Known limitation

Drilldowns fall back to `coalesce(ISFGroup, VMSize)` — the *consumed* size — while the overview falls back to `SKUName_s`, the *reserved* size. For a reservation on `Standard_DS4_v2` running `DS3_v2` underneath, those don't meet: which sizes belong to a retired group is exactly the information the ISF data no longer carries. So a drilldown on an unmatched reservation shows usage of the identical SKU only, not the group's flexibility.

Better than showing nothing, but not a full restoration. Recovering it would need the retired ratios, which is what #2300 originally proposed and #2308 explicitly stopped short of.

## Why not fix the data instead

An earlier revision of #2308 backfilled 1,220 retired SKUs from the ratio file the API replaced. It was dropped: those rows would be frozen forever, since nothing can refresh or correct them once the source is gone, inside a dataset whose contract is a weekly refresh from the authoritative API. Fixing the consumer avoids carrying dead reference data.

## Test plan

- [ ] Open **Reservations usage** against a subscription holding a reservation on a retired size (`Av2`, `Dv2`, `Dv3`, `Ev3`, `F`, `G`, `H`, `NC`, `NV`, …) and confirm it now appears with a correct utilization percentage
- [ ] Confirm reservations on current sizes are unchanged — same groups, same quantities, same percentages
- [ ] Confirm the unused-cost view lists the same reservations as the utilization view

**This is reasoned, not verified.** No test in this repo executes the workbook KQL, and I have no Optimization Engine workspace with reservation data to run it against. The join semantics and the null behaviour that motivates the `coalesce` should be confirmed by someone who does before merging.


RolandKrummenacher:
Closing — folded into #2308 instead. Splitting this out was my call and the wrong one; the change belongs in the same PR as the rest of the ISF work. Same commit, same content, now on \`RolandKrummenacher/isf-legacy-floor\`.

PR #2326, opened 2026-09-16
Document how the schema's columns nest

### LINKED ISSUE #2325: Hub database guide does not say how the schema's columns nest
## 🐛 Problem

`src/queries/finops-hub-database-guide.md` defines each column on its own, but never says how the columns nest. A reader who has a breakdown by `ServiceCategory` and wants to drill into it has nothing in the guide to follow, and no way to tell which of the available columns is "one level down".

Searching the guide for `hierarch` returns a single hit, and it is the *financial* hierarchy inside an example query — not the organizational or service structure.

This matters most for the service columns, because there are two chains and they are easy to mistake for one:

```
FOCUS:  ServiceCategory -> ServiceSubcategory -> ServiceName
Azure:  x_SkuMeterCategory -> x_SkuMeterSubcategory -> SkuMeter
```

They are parallel classifications. A single meter category can sit under several service categories, so drilling from one chain into the other does not partition anything, and a meter category's total is not a subset of any one service category. Nothing in the guide warns about this.

## 👣 Repro steps

```kusto
Costs()
| where ChargePeriodStart >= startofmonth(ago(30d))
| summarize ServiceCategories = dcount(ServiceCategory) by x_SkuMeterCategory
| where ServiceCategories > 1
| order by ServiceCategories desc
```

Run on two unrelated hubs, this returns a non-empty result on both — meter categories that map to more than one service category, some to eight.

## 🤔 Expected

A short section stating the organizational chain, the two service chains, and the fact that they do not nest — so a reader can drill down without inventing a structure.

## 🔧 Environment

- **Affected file:** `src/queries/finops-hub-database-guide.md` on `dev`
- **FinOps hub version:** any

## ℹ️ Additional context

PR follows. Related but filed separately: the same guide's `Costs()` table reference has drifted from the deployed schema.

RolandKrummenacher:
Fixes #2325.

## 🛠️ Changes

Adds a **Column relationships** section to the hub database guide, between *Key enrichment columns* and *Example queries*, plus its table-of-contents entries.

It covers three things the guide did not state:

- **The organization chain** — `BillingAccountName → SubAccountName → x_ResourceGroupName → ResourceName`, a strict tree where each level partitions the one above.
- **The two service chains** — the FOCUS one and the Azure meter one, side by side rather than as a single six-level path.
- **Why they cannot be mixed** — a meter category can appear under several service categories, so crossing the chains does not partition anything. A query is included so the reader can confirm it against their own data rather than taking it on faith.

It also notes that `ServiceSubcategory` and `SkuMeter` are FOCUS 1.2 columns, so on a hub whose exports predate 1.2 they are present but empty — an empty level meaning "not in this export" rather than "no spend".

## ✅ Verification

The claim that the service chains do not nest was checked on two unrelated hubs. On both, the query in the new section returns meter categories mapping to more than one service category, some to eight.

Markdown only — no queries or schema are changed, so there is nothing to regress.

## 📝 Notes

Related and filed separately, since they are different problems: the same guide's `Costs()` table reference has drifted from the deployed schema. I will open that one after this, to keep the two reviewable on their own.


RolandKrummenacher:
Correction to the framing above, after finding prior art I had missed.

`src/templates/finops-hub-copilot-studio/knowledge/schema-reference.md` already documents both chains — an **Organization hierarchy** table and a **Service hierarchy** section carrying the same point in one line: *"FOCUS levels (first 3) don't align 1:1 with Azure meter levels (last 3)."*

So this content is not new to the toolkit; it is absent from the **database guide** while living in a template knowledge file. That makes this PR a consolidation step rather than an addition, which is the more useful way to read it: the guide is the shared artifact that the plugins and agents build on, and the template file is a copy that has to be maintained separately.

Two things I would still argue for keeping from this PR over the template's wording:

1. **The consequence, not just the fact.** "Don't align 1:1" tells a reader the chains differ; it does not tell them that a meter category's total is therefore not a subset of any one service category, which is the part that produces wrong breakdowns.
2. **A verification query**, so the reader confirms it on their own data instead of trusting either document.

One correction to the template's text while it is in view, which I verified rather than assumed: it says `ServiceSubcategory` is *"Currently blank in Microsoft data."* That is version-dependent, not universal. Measured over the same 30-day window on two unrelated hubs, it is populated on every row on one and empty on every row on the other — consistent with it being a FOCUS 1.2 column, present but unfilled where the exports predate 1.2. The note added in this PR states it that way.

RolandKrummenacher:
Superseded by #2328 — same commit, moved to a branch in this repo for the same reason as #2327. The prior-art correction I posted below is folded into the new PR's description.

PR #2330, opened 2026-09-16
fix(hubs): paginate Azure Resource Graph queries past 1,000 rows

### LINKED ISSUE #2320: ADF queries_ResourceGraph_ExecuteQuery only fetches first 1000 results
## 🐛 Problem
When fetching resource graph queries with more than 1000 results it would only fetch 1000 results. For example the src/templates/finops-hub/modules/Microsoft.FinOpsHubs/Recommendations/queries/Recommendations-Microsoft-AdvisorCost.json could return more then 1000 results.

Change should happen in src/templates/finops-hub/modules/Microsoft.FinOpsHubs/AzureResourceGraph/app.bicep

## 👣 Repro steps
TODO: Add repro steps below:
1. add recommandation for the finops hub
2.  execute the pipeline  queries_ExecuteETL  in ADF
3. check how many results end up in ingestion recommendation blob storage
4. would give you 1000 rows back.

## 🤔 Expected
return all records back.

## 📷 Screenshots

## 🔧 Environment
- **FinOps hub version:** 14
- **Billing account type:**  MCA (Microsoft Customer Agreement) 
- **Power BI report type:** Storage-based / ADX (Azure Data Explorer)
- **Cost Management export:** FOCUS 1.2-preview

## ℹ️ Additional context

## 🙋‍♀️ Ask for the community

We could use your help:

flanakin:
🤖 [AI]

## Summary

- Azure Resource Graph caps a single query response at 1,000 rows and returns a `$skipToken` when more results exist, but the `queries_ResourceGraph_ExecuteQuery` pipeline ran a single Copy activity with no pagination, silently truncating larger result sets (for example `Recommendations-Microsoft-AdvisorCost`).
- The ADF REST connector's `paginationRules` can only inject a continuation value into a URL, query string, or header — not a POST body — so it can't express ARG's `$skipToken` continuation on its own.
- Replaced the single Copy activity with an `Until` loop: each iteration runs the query via a Web activity to check for results and read `$skipToken`, then re-runs the same page via Copy (to keep schema mapping/translator support) and writes it to its own `_{page}.parquet` file. The loop stops once ARG stops returning a token.

## Test plan

- [x] `bicep build src/templates/finops-hub/main.bicep --stdout` compiles cleanly
- [x] `Invoke-Pester -Path ./src/powershell/Tests/Unit/HubsIngestionQueries.Tests.ps1` — 192/192 passed
- [ ] Manual: deploy a hub, run `queries_ExecuteETL` against a tenant with >1,000 Advisor cost recommendations, confirm multiple `_{page}.parquet` files land in ingestion storage and all rows reach the destination table

Fixes #2320


src/templates/finops-hub/modules/Microsoft.FinOpsHubs/AzureResourceGraph/app.bicep
```diff
@@ -102,115 +102,245 @@ resource pipeline_ExecuteQuery 'Microsoft.DataFactory/factories/pipelines@2018-0
...
-                  storeSettings: {
-                    type: 'AzureBlobFSWriteSettings'
-                  }
-                  formatSettings: {
-                    type: 'ParquetWriteSettings'
-                  }
+              }
+            }
+            {
+              name: 'Fetch ARG Page'
```
RolandKrummenacher, line 136: **Web activity output limit.** `Fetch ARG Page` now pulls a full 1,000-row page into the activity output. ADF caps a Web activity response payload at roughly 4 MB, and the activity fails hard on `The length of execution output is over limit` rather than truncating.

`Recommendations-Microsoft-AdvisorCost` is the dataset most likely to hit it: `x_RecommendationDetails` is `bag_merge` of the recommendation bag plus the whole `properties.extendedProperties` blob, stringified. At around 4 KB per row (routine for Advisor extended properties), a full page crosses the limit. Because page 1 is the first thing the loop does, a tenant with more than 1,000 Advisor cost recommendations would go from silently truncating at 1,000 rows to ingesting nothing.

Two ways out: send `options.$top` (say 250) so a page can't get that large, or page with `options.$skip` off the old `| count` probe, which keeps the probe response small and also removes the need to read `$skipToken` out of a full response.

RolandKrummenacher (commented):
Reviewed the pagination rework end to end. The `Until` shape is right and the diagnosis matches what I see in the module.

Things I checked that are fine, so you don't have to re-derive them: the `_{page}` suffix doesn't break ADX table routing (the table comes from the folder path, not the file name) or the drop-extents cleanup (per-file `drop-by:` tags plus a per-run `ingestionId` means no stale pages and no upgrade hazard); manifest gating means a mid-loop failure can't half-ingest; and the request-body JSON escaping and the zero-row exit both behave the same as the old `| count` path.

Five things below. The first two are the ones I'd want resolved before this merges — the Web activity payload limit could turn a truncation bug into a hard failure on the very dataset in the issue, and ARG paging without a sort can duplicate and drop rows.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/AzureResourceGraph/app.bicep
```diff
@@ -102,115 +102,245 @@ resource pipeline_ExecuteQuery 'Microsoft.DataFactory/factories/pipelines@2018-0
...
+                value: {
                   // Query text is from trusted config/queries/*.json files; no escaping needed.
-                  requestBody: {
-                    value: '@concat(\'{ "query": "\', pipeline().parameters.query, \' | extend x_SourceName=\\"\', pipeline().parameters.querySource, \'\\", x_SourceType=\\"\', pipeline().parameters.queryType, \'\\", x_SourceProvider=\\"\', pipeline().parameters.queryProvider, \'\\", x_SourceVersion=\\"\', pipeline().parameters.queryVersion, \'\\"" }\')'
-                    type: 'Expression'
-                  }
-                  additionalHeaders: {
-                    'Content-Type': 'application/json'
-                  }
+                  value: '@concat(\'{ "query": "\', pipeline().parameters.query, \' | extend x_SourceName=\\"\', pipeline().parameters.querySource, \'\\", x_SourceType=\\"\', pipeline().parameters.queryType, \'\\", x_SourceProvider=\\"\', pipeline().parameters.queryProvider, \'\\", x_SourceVersion=\\"\', pipeline().parameters.queryVersion, \'\\""\', if(equals(variables(\'skipToken\'), \'\'), \'\', concat(\', "options": { "$skipToken": "\', variables(\'skipToken\'), \'" }\')), \' }\')'
```
RolandKrummenacher, line 130: **Paged ARG results need a deterministic sort.** ARG documents that paging is non-deterministic unless the query sorts on a unique field — rows can repeat across pages or be skipped. `Recommendations-Microsoft-AdvisorCost.json` ends at `| join kind=leftouter (...) on SubAccountId | project-away SubAccountId1` with no `order by`, so nothing pins the order.

This PR is what makes that reachable; single-page queries were never affected. The double fetch compounds it: the `$skipToken` read by `Fetch ARG Page` is replayed by `Write ARG Page` as an independent ARG evaluation, so the page written to Parquet isn't necessarily the page whose token was read.

A tenant with 2,500 recommendations can end up with 2,500 rows that duplicate some recommendation IDs and omit others. The duplicates then survive pre-ingest cleanup, because each page lands as its own extent under a distinct `drop-by:` file tag.

Appending a sort on a unique column (`x_RecommendationId`, or `id` before the project) to the composed query text here covers both halves.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/AzureResourceGraph/app.bicep
```diff
@@ -102,115 +102,245 @@ resource pipeline_ExecuteQuery 'Microsoft.DataFactory/factories/pipelines@2018-0
...
+              ]
+              policy: {
+                secureOutput: false
+                secureInput: false
+              }
+              userProperties: []
+              typeProperties: {
+                variableName: 'hasMorePages'
+                value: {
+                  value: '@and(variables(\'hasResults\'), not(equals(variables(\'skipToken\'), \'\')))'
```
RolandKrummenacher, line 232: **An empty page carrying a `$skipToken` stops the loop.** ARG can return an empty `data` array while still handing back a continuation token; the documented client pattern loops on the token alone, not on row count.

When that happens mid-result-set, `hasResults` is false, `hasMorePages` goes false, the `Until` exits, and every remaining page is dropped — silent truncation, which is the bug this PR fixes, just at a different row count.

Suggest `@not(equals(variables('skipToken'), ''))` here. `hasResults` is already doing its real job at `If Page Has Results` below, where it correctly keeps an empty page from hitting a schema-mapping error.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/AzureResourceGraph/app.bicep
```diff
@@ -102,115 +102,245 @@ resource pipeline_ExecuteQuery 'Microsoft.DataFactory/factories/pipelines@2018-0
...
+                        value: '@string(add(int(variables(\'pageNumber\')), 1))'
+                        type: 'Expression'
+                      }
+                    }
+                  }
+                ]
+              }
             }
           ]
+          timeout: '0.02:00:00'
```
RolandKrummenacher, line 343: **Two hours may be tight at two ARG round trips per page.** A tenant with 100,000 Advisor cost recommendations needs 100 pages, each a Web activity (5 min timeout) plus a Copy (10 min timeout). A steady few seconds per page is fine, but ARG throttling pushes it over — and `Write ARG Page` has `retry: 0`, so a single 429 fails the run outright.

The failure mode is at least clean: `IngestionQueries` hits `Query Engine Pipeline Failed`, never writes the manifest, and ADX keeps its prior extents, so nothing half-ingests. But the dataset stops refreshing silently while `Delete Old Files Loop` has already removed the previously ingested Parquet files.

Worth either raising the cap or deriving it from the expected row count.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/AzureResourceGraph/app.bicep
```diff
@@ -102,115 +102,245 @@ resource pipeline_ExecuteQuery 'Microsoft.DataFactory/factories/pipelines@2018-0
...
+                      }
+                    ]
+                    outputs: [
+                      {
+                        referenceName: dataset_ingestion.name
+                        type: 'DatasetReference'
+                        parameters: {
+                          blobPath: {
+                            // Insert a "_<page number>" suffix before the ".parquet" extension so each page gets a unique file in the same ingestion folder.
+                            value: '@concat(substring(pipeline().parameters.ingestionPath, 0, sub(length(pipeline().parameters.ingestionPath), 8)), \'_\', variables(\'pageNumber\'), \'.parquet\')'
```
RolandKrummenacher, line 310: **Minor: the hardcoded `8` couples this module to the caller's file naming.** It works today — `IngestionQueries/app.bicep` builds `ingestionPath` as `... queryType + '.parquet'`. But a caller that passes a folder or a differently suffixed path gets the last 8 characters silently removed, with no error at the point of the mistake.

For `.../20260916-120000_abc1234__Microsoft-AdvisorCost` the pipeline would write `.../20260916-120000_abc1234__Microsoft-_1.parquet`, and it would surface much later as a missing or mis-routed ingestion.

`replace(pipeline().parameters.ingestionPath, '.parquet', concat('_', variables('pageNumber'), '.parquet'))` removes the coupling, or the caller could pass the base path without the extension.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/AzureResourceGraph/app.bicep
flanakin: 🤖 [AI][Claude] ✅ **Implemented**

Added `options.$top` = 500 (used by both the fetch and the write). I chose 500 over 250 to keep call count and 429 exposure low: about 2 MB per page at 4 KB/row, so it tolerates rows up to about 8 KB. Row size is not measured, so this is a hedge; it is a one-number change if real payloads are smaller.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/AzureResourceGraph/app.bicep
flanakin: 🤖 [AI][Claude] ✅ **Implemented**

Appended `| order by x_RecommendationId asc` to the composed query (all recommendation queries emit that column). The double fetch remains, but both calls now sort identically.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/AzureResourceGraph/app.bicep
flanakin: 🤖 [AI][Claude] ✅ **Implemented**

`hasMorePages` now depends only on the `$skipToken`. `hasResults` still gates the write.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/AzureResourceGraph/app.bicep
flanakin: 🤖 [AI][Claude] ✅ **Implemented**

Raised the `Until` timeout to 6h and set `Write ARG Page` retry to 2 (60s interval) so a single 429 doesn't fail the run.

src/templates/finops-hub/modules/Microsoft.FinOpsHubs/AzureResourceGraph/app.bicep
flanakin: 🤖 [AI][Claude] ✅ **Implemented**

Switched to `replace(ingestionPath, '.parquet', ...)`.

PR #2331, opened 2026-09-16
[Open Data] Sort instance size flexibility output ordinally so regeneration is host-independent

RolandKrummenacher:
## Summary

Follow-up to #2308. The ISF generator sorted with a bare `Sort-Object`, whose comparer follows the current culture, so the same records regenerated on a differently configured host come out in a different order.

`InstanceSizeFlexibility.csv` is **unchanged** — ordinal ordering is the order already committed on `dev`. This fixes only the script that produces it.

## How it surfaced

Dispatching **Update Instance Size Flexibility** after #2308 merged ([run 35117329813](https://github.com/microsoft/finops-toolkit/actions/runs/35117329813)) succeeded in 4m23s and reproduced the published dataset exactly — same 1,363 rows, same 213 groups, same ratios, verified as a set with line endings normalized. It still pushed a **247-insertion / 247-deletion** diff, every line of it pure reordering.

Measured on a Windows host at `en-CH`:

| pair | culture-aware (what the script did) | ordinal |
| --- | --- | --- |
| `DCadsv5-series DedicatedHost` vs `Dasv4 Series` | `Dasv4` first | `DCadsv5` first |
| `BS Series High Memory` vs `Basv2 Series` | `Basv2` first | `BS Series` first |

Culture-aware collation interleaves case; ordinal puts the uppercase letter first. The file committed on `dev` is in ordinal order while the script emitted culture-aware order, so **every scheduled run would have pushed this same diff** whether or not any ISF data changed — burying real changes in review, and flipping back and forth as the generator is run by hand on Windows per the README and on `ubuntu-latest` by the workflow.

## The fix

#2308 pinned `InvariantCulture` on both parse sites and the write path but left the sort key culture-dependent. This applies the same treatment there, with an explicit `[string]::CompareOrdinal` comparer.

One behavioral note: `Sort-Object` silently tolerated a null record set, which `List[object]` does not. An empty sweep is guarded explicitly — two existing tests (`Queries VirtualMachines, RedisCache, and DedicatedHost but not BlockBlob by default` and `Does not enumerate regions when -Location is supplied`) caught that on the first revision.

## Test plan

- [x] `Invoke-Pester ./src/powershell/Tests/Unit/* ./src/powershell/Tests/Lint/*` — **6,023 passed, 0 failed** (4 pre-existing skips), exactly +4 over the baseline of 6,019
- [x] Four new tests: group ordering, SKU ordering within a group, ordering held under `de-DE`, and an assertion that the published file is in ordinal order — the last catches a file regenerated by hand on a differently configured host
- [x] The comparer reproduces the committed 1,363-row order exactly from shuffled input, verified against the real file rather than mocks
- [x] `InstanceSizeFlexibility.csv` is byte-identical to `dev`
- [x] Empty sweep yields zero rows instead of throwing

No changelog entry: the ISF dataset is still under **Unreleased → Added** as a new dataset, and row ordering is not a documented property of it. Nothing user-visible changes.


src/scripts/Update-InstanceSizeFlexibility.ps1
```diff
@@ -368,10 +368,22 @@ if (-not $Raw)
     $allRecords = Get-NormalizedRecords -Records $allRecords
 }
 
+# Sorted ordinally rather than with Sort-Object, whose comparer follows the current culture.
+# Culture-aware collation orders "DCadsv5-series DedicatedHost" after "Dasv4 Series" while an
+# ordinal comparison puts it before, so the same records regenerated on a differently configured
+# host produce a few hundred lines of pure reordering that bury any real change in review.
+$sortedRecords = [Collections.Generic.List[object]]::new()
+if ($allRecords) { $sortedRecords.AddRange([object[]]@($allRecords)) }  # an empty sweep leaves $allRecords null
+$sortedRecords.Sort([Comparison[object]] {
```
flanakin, line 382: 🤖 [AI][Claude] 💡 **Suggestion**

`List.Sort` is an unstable sort, whereas `Sort-Object` is stable. In the normalized path `ArmSkuName` is unique so this can't matter, but with `-Raw` (no `Get-NormalizedRecords`) two records sharing the same group and SKU could come out in a different relative order between runs. That would reintroduce nondeterministic diffs. Either add a final tiebreaker (for example, compare `Ratio` ordinally) or add a comment that ties are impossible or irrelevant in this path.

src/powershell/Tests/Unit/Update-InstanceSizeFlexibility.Tests.ps1
```diff
@@ -388,6 +444,21 @@ Describe 'Update-InstanceSizeFlexibility' {
             @($csv | Where-Object { [double]$_.Ratio -le 0 }).ArmSkuName | Should -BeNullOrEmpty
         }
 
+        It 'Is written in ordinal sort order' {
```
flanakin, line 447: 🤖 [AI][Claude] 💡 **Suggestion**

This comparer is a copy of the one in `Update-InstanceSizeFlexibility.ps1`. If the script's sort key changes later (for example, a third key), this test will keep asserting the old order, or fail for the wrong reason. A short comment saying the two must stay in sync would make that visible.

flanakin (commented):
🤖 [AI][Claude] **PR Review**

**Summary:** The fix is correct and focused. The 28 tests in the ISF suite pass locally, and reverting the script to `Sort-Object` makes 3 of the new tests fail, so they guard the change. No blockers or should-fix items.

### 💡 Suggestions (2)

1. `List.Sort` is unstable, unlike `Sort-Object`. Ties on group + SKU could reorder between runs with `-Raw`.
2. The comparer is duplicated in the script and the published-file test. Note that they must stay in sync.

src/scripts/Update-InstanceSizeFlexibility.ps1
RolandKrummenacher: Checked this before adding a tiebreaker, and a tie can't reach the sort in either path.

`$seen` is keyed by `"$group|$ArmSkuName"`, so the collection holds one record per group + SKU pair by construction, and a duplicate `ArmSkuName` across groups throws a few lines above. `-Raw` only skips `Get-NormalizedRecords`; it doesn't touch that dedupe. So the two keys are already a total order and a `Ratio` comparison would be unreachable code.

Went with your second option and recorded the reasoning at the comparer, so the next person doesn't have to re-derive it before touching the sort keys.

Fixed in 37f96abe.

src/powershell/Tests/Unit/Update-InstanceSizeFlexibility.Tests.ps1
RolandKrummenacher: Added the sync note on both sides: the test now says the comparer mirrors `Update-InstanceSizeFlexibility.ps1` and that a new sort key in the script leaves this assertion on the old order, and the script points back at this test.

Fixed in 37f96abe.

